PR #1, opened 2023-02-21
feat: add initial implementation of Kubo RPC API 

nathanielc:
This commit adds the basic structure of the repo and an incomplete
implementation of the Kubo RPC API.
Currently only the `/dag` and `/swarm` endpoints are implemented.


ceramic-kubo-rpc/src/dag.rs
```diff
@@ -0,0 +1,280 @@
...
+        "RemPath": "",
+    });
+    let mut data = Vec::new();
+    resolved.encode(DagJsonCodec, &mut data).unwrap();
+    Ok(HttpResponse::Ok()
+        .content_type(ContentType::json())
+        .body(data))
+}
+
+#[cfg(test)]
```
nathanielc, line 205: Note, these tests do not pass. I seem to have found a bug in actix-web  or am not using their testing API correctly. I have pinged on their discord hoping to get some help.

nathanielc: I have since added tests for all implemented endpoints.

Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[workspace]
+members = ["ceramic-kubo-rpc", "ceramic-one"]
+
+[workspace.dependencies]
+anyhow = "1"
```
dbcfd, line 5: If we use full versions here, we can use dependency bot.

ceramic-kubo-rpc/src/error.rs
```diff
@@ -0,0 +1,21 @@
+//! Provides simple error type for communicating specific failures.
+use thiserror::Error;
+
+/// Error type for specific failures.
+/// These generally come from here
+/// https://grpc.github.io/grpc/core/md_doc_statuscodes.html
+/// However we start with a minimal set and can add more as needed.
+#[derive(Debug, Error)]
+pub enum Error {
```
dbcfd, line 9: Do we need these specific error types? I think we can do something like https://www.reddit.com/r/rust/comments/ozc0m8/an_actixanyhow_compatible_error_helper_i_found/?utm_source=share&utm_medium=web2x&context=3 if we need to convert to actix easily.

ceramic-kubo-rpc/src/dag.rs
```diff
@@ -0,0 +1,85 @@
...
+    pb::DagPbCodec,
+    prelude::{Codec, Decode, Encode},
+    Ipld,
+};
+
+use crate::{error::Error, IpfsDep};
+
+/// Get a DAG node from IPFS.
+#[tracing::instrument(skip(client, output_codec))]
+pub async fn get<T, C>(client: T, cid: Cid, output_codec: C) -> Result<Vec<u8>, Error>
```
dbcfd, line 18: Kind of seems like all of these should be hidden behind a trait, so we're not passing client around.

ceramic-kubo-rpc/src/http.rs
```diff
@@ -0,0 +1,611 @@
...
+                    .service(dag_scope::<T>())
+                    .service(swarm_scope::<T>()),
+            )
+    })
+    .bind(addrs)?
+    .run()
+    .await
+}
+
+fn dag_scope<T>() -> Scope
```
dbcfd, line 54: I'd like to see these places in their own files, with unit tests. Application level can do bind and run, while we should only expose Scopes and possibly App.

dbcfd (changes requested):

Cargo.toml
nathanielc: Hmm can dependency bot use the Cargo.lock file? That would be ideal so we do not get unintentionally stuck on older versions.

ceramic-kubo-rpc/src/error.rs
nathanielc: We don't _need_ these yet but as a general design I prefer libraries to use an error enum instead of anyhow so that consumers of the lib can handle specific errors as needed in their application. So while this error enum is very light right now the intent is to let it grow as the needs of the lib grow.

ceramic-kubo-rpc/src/dag.rs
nathanielc: Can you expand on this a bit? The `client` is a trait already and these functions provide the crate level API we want to expose. What code structure are you expecting?

ceramic-kubo-rpc/src/http.rs
nathanielc: Yeah an `http::dag` and `http::swarm` modules would make sense I can make that refactor

Cargo.toml
dbcfd: I usually don't include the Cargo.lock file. You can package it with the release if you want repeatable builds, although having the fully specified versions is usually enough.

nathanielc: If we specify each version precisely in Cargo.toml then we loose the ability to use `cargo update` to update to minor/patch versions. The workflow I am aiming for is that we lock our dependencies by default (so that we have reproducible builds) but then we can easily update to new non breaking versions of dependencies explicitly. 

So I would prefer we only specify major versions or minor version in the case of 0. releases. Then periodically we explicitly run `cargo update` to ensure we are getting the most recent versions of our dependencies. 

From [docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html):

> An update is allowed if the new version number does not modify the left-most non-zero digit in the major, minor, patch grouping. 

Thoughts?

nathanielc: Also from a read of this https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-the-dependency-graph#supported-package-ecosystems  it looks like dependabot can read the Cargo.lock file. I can verify once we get this initial PR merged.

ceramic-kubo-rpc/src/http/dag.rs
```diff
@@ -0,0 +1,315 @@
...
+{
+    let ipfs_path = IpfsPath::from_str(query.arg.as_str()).map_err(|e| Error::Invalid(e.into()))?;
+    match query.output_codec.as_str() {
+        DAG_JSON => Ok(HttpResponse::Ok()
+            .content_type(ContentType::json())
+            .body(dag::get(data.api.clone(), &ipfs_path, DagJsonCodec).await?)),
+        DAG_CBOR => Ok(HttpResponse::Ok()
+            .content_type(ContentType::octet_stream())
+            .body(dag::get(data.api.clone(), &ipfs_path, DagCborCodec).await?)),
+        _ => Err(Error::Invalid(anyhow!(
```
smrz2001, line 56: Now that you have support for dag-jose, should there be support for it here too?

nathanielc: Yeah lets do these as subsequent PRs

ceramic-kubo-rpc/src/http/dag.rs
```diff
@@ -0,0 +1,315 @@
...
+                            Error::Internal(
+                                Into::<anyhow::Error>::into(e).context("reading multipart chunk"),
+                            )
+                        })?
+                        .to_vec(),
+                )
+            }
+
+            let cid = match (query.input_codec.as_str(), query.store_codec.as_str()) {
+                (DAG_JSON, DAG_CBOR) => {
```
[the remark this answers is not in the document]
nathanielc: Yeah lets do these as subsequent PRs

Makefile
```diff
@@ -0,0 +1,24 @@
+# Makefile provides an API for CI related tasks
+# Using the makefile is not required however CI
+# uses the specific targets within the file.
+# Therefore may be useful in ensuring a change
+# is ready to pass CI checks.
+
+.PHONY: all
+all: build check-fmt check-clippy test
+
+.PHONY: build
```
AaronGoldman, line 10: may be nice to have a `run`
```
.PHONY: run
run:
	cargo run --all-features --locked --bin ceramic-one -- daemon -b 127.0.0.1:5001      
```
not a blocker but consider.

ceramic-kubo-rpc/src/http/dag.rs
```diff
@@ -0,0 +1,315 @@
...
+}
+// used to provide default to query structs
+fn dag_json() -> String {
+    DAG_JSON.to_string()
+}
+
+#[derive(Debug, Deserialize)]
+struct GetQuery {
+    arg: String,
+    #[serde(rename = "output-codec", default = "dag_json")]
```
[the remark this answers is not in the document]
nathanielc: No, but its confusing, the `default` refers to a function name that can be called to provide the default value. The `dag_json` function returns a string `dag-json`.

smrz2001 (approved):
🥳

dbcfd (approved):

PR #3, opened 2023-03-14
feat(event): Ability to create and sign events for use with js-ceramic

dbcfd:


dbcfd:
> What does "document" mean in this context?
> 
>     * Looks like you've mainly implemented "commits" so far, can we call them "events"?
>       See our [spec](https://developers.ceramic.network/protocol/streams/event-log/), we now call them Init, Data, and Time Event
> 
>     * Documents as they exist in js-ceramic we wanted to not have part of this code base, instead we should mainly focus on streams and events

Yep, this work so far focuses on commits, and then the format for stream id type document (3). I think what you're suggesting is
* Rename the package to "ceramic-events"
* Rename Commit to Event. Right now js-ceramic doesn't distinguish between an init event and a data event, but rather a deterministic commit (init event w/o data), signed commit (init event w/ data), and signed commit + stream_id (data event). It seems like if we make this implementation match the spec, it won't work with js-ceramic.

oed:
Yes, I think that's roughly what I said.

> It seems like if we make this implementation match the spec, it won't work with js-ceramic.

Not sure I follow this. js-ceramic should definitely be following the spec. cc @AaronGoldman

dbcfd:
@oed Reworked PR to match more closely with terms used in spec. As implemented, this can be used with js-ceramic for creating, updating, and reading MIDs.

event/src/event.rs
```diff
@@ -0,0 +1,100 @@
...
+    #[tokio::test]
+    async fn should_roundtrip_init_event() {
+        let did = DidDocument::new("did:key:blah");
+        let model =
+            StreamId::from_str("kjzl6kcym7w8y6of44g27v981fuutovbrnlw2ifbf8n26j2t4g5mmm6zc43nx1u")
+                .unwrap();
+        let args = EventArgs::new_with_parent(&did, &model);
+        let evt = args.init().await.unwrap();
+        let _: serde_json::Value = serde_ipld_dagcbor::from_slice(evt.encoded.as_ref()).unwrap();
+    }
```
nathanielc, line 65: Should this test have some assertions?

core/src/jws.rs
```diff
@@ -0,0 +1,84 @@
...
+use ssi::jwk::Algorithm;
+
+#[derive(Serialize, Deserialize)]
+pub struct JwsSignature {
+    pub protected: Option<Base64String>,
+    pub signature: Base64UrlString,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct Jws {
```
nathanielc, line 13: I'd like to understand what needs to change in the `dag-jose` crate in order to be able to use its types here?

I understand this `Jws` struct also add the signing/verification logic however I do not see the `dag-jose` crate used anywhere in the code base. As such seems like we should fix that crate to make it amenable to this code as its purpose was to be a building block for Ceramic events.

Cargo.toml
```diff
@@ -1,23 +1,37 @@
 [workspace]
-members = ["ceramic-kubo-rpc", "ceramic-one"]
+members = [
+    "core",
+    "event",
```
nathanielc, line 4: I like these shorter names without the `ceramic-` stutter I'll likely rename all the directories in the repo to match this pattern.

Cargo.toml
```diff
@@ -1,23 +1,37 @@
...
+]
 
 [workspace.dependencies]
 anyhow = "1"
 async-stream = "0.3"
 async-trait = "0.1"
 ceramic-one = { path = "./ceramic-one" }
 dag-jose = "0.1"
+cid = { version = "0.10.1", features = ["serde-codec"] }
+env_logger = "0.10"
```
nathanielc, line 16: I would not expect to need to use `env_logger` the code base uses `tracing` for its logging needs and the tracing ecosystem has an env filter based subscriber. Did you find this was necessary?

core/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+did-pkh = "0.2.1"
+hex = "0.4.3"
+multibase.workspace = true
+once_cell.workspace = true
+regex = "1.7.1"
+serde.workspace = true
+serde_json.workspace = true
+serde_ipld_dagcbor.workspace = true
+ssi.workspace = true
+unsigned-varint = "0.7"
```
nathanielc, line 25: should we move the rest of these deps to the workspace? Some of them at least are shared (hex, base64 ...)

core/src/lib.rs
```diff
@@ -0,0 +1,189 @@
+mod jwk;
```
nathanielc, line 1: Can you add  to this crate so we are consistent across the repo?
```
#![deny(warnings)]
#![deny(missing_docs)]
```

core/src/jwk.rs
```diff
@@ -0,0 +1,180 @@
...
+    async fn should_convert_did_key_generated_without_vm() {
+        let jwk = JWK::generate_ed25519().unwrap();
+        let did = did_method_key::DIDKey
+            .generate(&ssi::did::Source::Key(&jwk))
+            .unwrap();
+        let did = DidDocument::new(&did);
+        let other_jwk = Jwk::new(&did).await.unwrap();
+        if let (Params::OKP(did), Params::OKP(orig)) = (&other_jwk.params, jwk.params) {
+            assert_eq!(did.public_key, orig.public_key)
+        }
```
nathanielc, line 137: Should we have an else `panic` branch for the failure case when the let expressions do not match?

core/src/jwk.rs
```diff
@@ -0,0 +1,180 @@
...
+            ..Default::default()
+        };
+        let did = ssi::did::DocumentBuilder::default()
+            .id(did)
+            .verification_method(vec![VerificationMethod::Map(pko)])
+            .build()
+            .unwrap();
+        let other_jwk = Jwk::new(&did).await.unwrap();
+        if let (Params::OKP(did), Params::OKP(orig)) = (&other_jwk.params, jwk.params) {
+            assert_eq!(did.public_key, orig.public_key)
```
nathanielc, line 157: Same question here.

core/src/jwk.rs
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    #[tokio::test]
+    async fn should_convert_did_key_with_vm() {
+        let _ = env_logger::try_init();
+        let did = DidDocument::new("did:key:zQ3shokFTS3brHcDQrn82RUDfCZESWL1ZdCEJwekUDPQiYBme#zQ3shokFTS3brHcDQrn82RUDfCZESWL1ZdCEJwekUDPQiYBme");
+        let jwk = Jwk::new(&did).await.unwrap();
+        log::debug!("JWK={:?}", jwk);
+        if let Params::OKP(op) = &jwk.params {
+            log::debug!("OP={}", String::from_utf8_lossy(op.public_key.0.as_ref()))
```
nathanielc, line 168: Can you change this to not use `log` but instead the `tracing` lib?

Cargo.toml
```diff
@@ -1,23 +1,37 @@
...
 futures-util = "0.3"
 iroh-api = { git = "https://github.com/nathanielc/beetle", branch = "expose-api-client" }
 iroh-embed = { git = "https://github.com/nathanielc/beetle", branch = "expose-api-client" }
 iroh-metrics = { git = "https://github.com/nathanielc/beetle", branch = "expose-api-client" }
 iroh-rpc-client = { git = "https://github.com/nathanielc/beetle", branch = "expose-api-client" }
 libipld = "0.15"                                                                                 # use same version as Iroh
 libp2p = { version = "0.50", default-features = false }                                          # use same version as Iroh
+log = "0.4"
```
nathanielc, line 24: Similar question for `log` re `env_logger`. I'd like to exclusively use the `tracing` analogs for this code base. Thoughts?

event/src/args.rs
```diff
@@ -0,0 +1,170 @@
...
+            sep: SEP,
+        }
+    }
+
+    pub fn parent(&self) -> &StreamId {
+        self.parent
+    }
+
+    pub async fn init(&self) -> Result<DeterministicEvent> {
+        let cmt = UnsignedEvent::<()>::init(self)?;
```
nathanielc, line 50: nit, but the name `cmt` was confusing at first. Then I realized it mean `commit`. Could you rename the uses of `cmt` to something more clear, maybe `evt` or `event`?

event/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+mod args;
```
nathanielc, line 1: Same here, can you add  to this crate so we are consistent across the repo?
```
#![deny(warnings)]
#![deny(missing_docs)]
```

nathanielc (changes requested):
This looks great, I have a few comments/suggestions below. 

My biggest question is how we can update the `dag-jose` crate to be of use here.

event/src/event.rs
dbcfd: We could change the unwrap to assert not throw, but essentially doing assertion by unwrap.

core/src/jws.rs
dbcfd: Yeah, I was really surprised once I got everything implemented, that the events we pass to ceramic don't actually use dag-jose. They just have the data so that ipfs can encode them as dag-jose.

Seems like there's a mismatch there.

Cargo.toml
dbcfd: Yeah, I like the crate names with the full `ceramic-` indicator, but since we're in a ceramic project, tend to drop the prefix.

Cargo.toml
dbcfd: It's useful for testing, but I don't think there's a concept of `workspace.dev-dependencies`.

core/Cargo.toml
dbcfd: Yeah, will look and see which I can move to workspace level.

event/src/event.rs
nathanielc: Makes sense now

Cargo.toml
```diff
@@ -1,23 +1,37 @@
...
+    "ceramic-kubo-rpc",
+    "ceramic-one"
+]
 
 [workspace.dependencies]
 anyhow = "1"
 async-stream = "0.3"
 async-trait = "0.1"
 ceramic-one = { path = "./ceramic-one" }
 dag-jose = "0.1"
```
nathanielc, line 14: OK, as I understand it we don't need any of the ser/de logic that dag-jose provides. In that case can we remove this dep?

core/src/lib.rs
```diff
@@ -65,6 +70,7 @@ impl_multi_base!(MultiBase58BtcString, multibase::Base::Base58Btc);
 impl_multi_base!(MultiBase64String, multibase::Base::Base64);
 impl_multi_base!(MultiBase64UrlString, multibase::Base::Base64Url);
 
+/// Newtype to encapsulate o value that is DagCbor encoded
```
nathanielc, line 73: nit, `o` -> `a`.

core/src/stream_id.rs
```diff
@@ -6,23 +6,31 @@ use std::io::Write;
 use serde::{Deserialize, Deserializer, Serialize};
 use unsigned_varint::{decode, encode};
 
+/// A stream id, which is a cid with a type
 #[derive(Debug, Eq, PartialEq)]
 pub struct StreamId {
+    /// The type of the stream
     pub r#type: u64,
```
nathanielc, line 13: Should we make this an `enum` that maps back to its u64 value?

Cargo.toml
dbcfd: Looks like kubo-rpc needs it still

Cargo.toml
nathanielc: I'd expect this is still not used. The tracing crate we use already honors RUST_LOG without this dep.

nathanielc (approved):
LGTM, feel free to merge when its green.

nathanielc:
ok I updated this PR to pass CI. There are two major changes:

* rebased against https://github.com/3box/rust-ceramic/pull/32  since that PR contains a large refactor
* fixed failing tests

Once https://github.com/3box/rust-ceramic/pull/32 merged we can merge this PR

event/src/deterministic_event.rs
```diff
@@ -0,0 +1,25 @@
+use crate::args::UnsignedEvent;
+use anyhow::Result;
+use ceramic_core::DagCborEncoded;
+use serde::Serialize;
+use std::fmt::Formatter;
+
+/// A deterministic event, where it will always hash the same way
+pub struct DeterministicEvent {
```
oed, line 8: Maybe we should call this "DeterministicInitEvent"? DataEvents are never "deterministic"

event/src/event.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+impl Event {
+    /// Create a new event from an unsigned event, signer, and jwk
+    pub async fn new<'a, T: Serialize>(
+        unsigned: &'a UnsignedEvent<'a, T>,
+        signer: &'a DidDocument,
+        jwk: &'a Jwk,
+    ) -> Result<Self> {
+        // encode our commit with dag cbor, hashing that to create cid
```
oed, line 28:
```suggestion
        // encode our event with dag cbor, hashing that to create cid
```

event/src/event.rs
```diff
@@ -0,0 +1,138 @@
...
+use crate::DidDocument;
+use anyhow::Result;
+use ceramic_core::{Cid, DagCborEncoded, Jwk, Jws};
+use multihash::{Code, MultihashDigest};
+use serde::Serialize;
+
+// https://github.com/multiformats/multicodec/blob/master/table.csv
+const DAG_CBOR_CODEC: u64 = 0x71;
+
+/// A ceramic event, which is a signed commit
```
oed, line 11:
```suggestion
/// A ceramic event
```

event/src/event.rs
```diff
@@ -0,0 +1,138 @@
...
+pub struct Event {
+    /// Cid of the data for the event
+    pub cid: Cid,
+    /// The data for the event to be encoded in the block
+    pub linked_block: DagCborEncoded,
+    /// JWS signature of the event
+    pub jws: Jws,
+}
```
oed, line 19: Probably don't need to change this now, but events can also contain one `cacaoBlock` (currently), and possible multiple CACAO blocks in the future.

I wonder if a better structure for an event is just a CAR file with all the relevant IPLD blocks?

If not clear already: both `jws` and `linked_block` should be IPLD objects.

core/src/stream_id.rs
```diff
@@ -0,0 +1,229 @@
...
+
+/// A stream id, which is a cid with a type
+#[derive(Debug, Eq, PartialEq)]
+pub struct StreamId {
+    /// The type of the stream
+    pub r#type: StreamIdType,
+    /// Cid of the stream
+    pub cid: Cid,
+    /// Optional commit for this stream
+    pub commit: Option<Cid>,
```
oed, line 40: Do we need to introduce "commitIds" yet (or at all)? 

There is a painpoint around how commitIds are specified currently because they kind of break the multiformat nature of streamIds.

Me and @AaronGoldman talked about introducing a new way to reference a specific event inside of a streamId. This approach would allow streamIds to always be a real multiformat:
https://github.com/AaronGoldman/CIP/blob/stream-sets-synchronization/CIPs/CIP-124/cip-124.md#eventid-sort-order

dbcfd: I mainly have it there to make sure we can read cid's with commit id's specified. If we don't think we'll have stream id's with commits, I'd gladly remove it.

PR #8, opened 2023-03-14
feat: Create the ahash hasher

AaronGoldman:
First pass at the associative hash function for hashing ceramic sets


ceramic-ahash/src/main.rs
```diff
@@ -0,0 +1,161 @@
...
+            hex::encode(u32::to_le_bytes(self.0[4])),
+            hex::encode(u32::to_le_bytes(self.0[5])),
+            hex::encode(u32::to_le_bytes(self.0[6])),
+            hex::encode(u32::to_le_bytes(self.0[7])),
+        ]
+        .join("")
+        .to_uppercase()
+    }
+
+    pub fn identity() -> Ahash {
```
nathanielc, line 70: Rust has a `Default` trait which is similar in concept. Here, perhaps implement the `Default` trait to make this behavior more generally useful?  You should be able to derive it `#[derive(Default)]`

ceramic-ahash/src/main.rs
```diff
@@ -0,0 +1,161 @@
...
+                include_str!("tests/wordle_words5.txt"),
+                "11013813008811902644706A2874A0E155BE2ACC17FA6FA953C9406450DF870F",
+            ],
+        ] {
+            assert_eq!(Ahash::digest_many(words.split(" ")).to_hex(), expected);
+        }
+    }
+}
+
+fn main() {
```
nathanielc, line 157: As mentioned above move this function into its own `main.rs` file while leaving everything else here as `lib.rs`.

nathanielc (changes requested):
Looks really good, the API boundary seems straightforward. A few notes/questions below.

ceramic-ahash/src/ahash.rs
```diff
@@ -0,0 +1,210 @@
...
+{
+    let b: &serde_bytes::Bytes = Deserialize::deserialize(deserializer)?;
+    Ok(AHash::from_bytes(
+        b.to_vec()
+            .as_slice()
+            .try_into()
+            .map_err(|_| D::Error::invalid_length(b.len(), &"hash must be len 32"))?,
+    )
+    .0)
+    // todo return error on errors
```
nathanielc, line 47: I think this todo is done.

ceramic-ahash/src/ahash.rs
```diff
@@ -0,0 +1,210 @@
...
+        self.0[0] = 0;
+        self.0[1] = 0;
+        self.0[2] = 0;
+        self.0[3] = 0;
+        self.0[4] = 0;
+        self.0[5] = 0;
+        self.0[6] = 0;
+        self.0[7] = 0;
+    }
+    fn push(&mut self, key: (&String, &AHash)) {
```
nathanielc, line 65: What is the `key.0` (the string) for? Does not look to be used in this implementation. Do other implementations need the key string?

ceramic-ahash/src/recon/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+    }
+}
+
+pub trait Hash: std::ops::Add<Output = Self> + Clone + Default + PartialEq {
+    fn identity() -> Self {
+        Self::default()
+    }
+    fn is_zero(&self) -> bool;
+    fn clear(&mut self);
+    fn push(&mut self, key: (&String, &AHash));
```
nathanielc, line 165: Similar question to before, what is the string for? Can we remove it or use a struct to give it self documenting semantics?

ceramic-ahash/src/recon/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+            r.insert(&key);
+        }
+        r
+    }
+
+    pub fn insert(&mut self, key: &str) {
+        self.keys.insert(key.to_string(), AHash::digest(key));
+    }
+
+    pub fn hash_range<H: Hash>(&self, left_fencepost: &str, right_fencepost: &str) -> H {
```
nathanielc, line 37: does this function need to be public? I don't see any public uses and it feels like an internal detail of the Recon implementation.

nathanielc (changes requested):
Overall looks great. Some questions below

ceramic-ahash/src/main.rs
```diff
@@ -0,0 +1,37 @@
...
+    let response = cat.process_message(&received);
+    println!("<- {}", response);
+    println!("After {:?}", cat);
+
+    println!("serde_json {}", serde_json::to_string(&received).unwrap()); // Message as json
+    println!(
+        "serde_ipld_dagcbor_hex {}",
+        hex::encode(serde_ipld_dagcbor::ser::to_vec(&received).unwrap())
+    ); // Message as dag cbor hex
+}
```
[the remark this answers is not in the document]
nathanielc: Or even simpler just move it to an example https://doc.rust-lang.org/cargo/reference/cargo-targets.html?highlight=examples#examples

.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+    "rust-analyzer.linkedProjects": [
+        "./ceramic-ahash/Cargo.toml",
+        "./ceramic-ahash/Cargo.toml"
+    ]
+}
```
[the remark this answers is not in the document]
AaronGoldman: removing: was local hints to vscode

ceramic-ahash/src/main.rs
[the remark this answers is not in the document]
AaronGoldman: everything in main is reflected in a test maybe I should just remove main

ceramic-ahash/src/ahash.rs
AaronGoldman: MemoryAHash.push((&String, &AHash)) needs to remember the string and not just the hash

recon/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+version.workspace = true
+edition.workspace = true
+authors.workspace = true
+license.workspace = true
+repository.workspace = true
+
+[dependencies]
+hex = "0.4.3"
+multihash = "0.18.0"
+rusqlite = "0.29.0"
```
nathanielc, line 13: can you put this dep behind a feature?

nathanielc (approved):

PR #13, opened 2023-03-21
feat: add block endpoints

nathanielc:
Adds endpoints for direct storage and retrieval of blocks.

### DIFF
--- a/ceramic-kubo-rpc/src/block.rs
+++ b/ceramic-kubo-rpc/src/block.rs
@@ -1,0 +1,41 @@
+//! Implements the dag endpoints.
+
+use iroh_api::Cid;
+use libipld::{
+    multihash::{Code, MultihashDigest},
+    prelude::Codec,
+};
+
+use crate::{error::Error, IpfsDep};
+
+/// Get a block from IPFS.
+#[tracing::instrument(skip(client))]
+pub async fn get<T>(client: T, cid: Cid) -> Result<Vec<u8>, Error>
+where
+    T: IpfsDep,
+{
+    let bytes = client.block_get(cid).await?;
+    Ok(bytes.to_vec())
+}
+
+/// Store a block into IFPS.
+#[tracing::instrument(skip_all)]
+pub async fn put<T, C>(client: T, codec: C, data: Vec<u8>) -> Result<Cid, Error>
+where
+    T: IpfsDep,
+    C: Codec,
+{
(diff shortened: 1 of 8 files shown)


ceramic-kubo-rpc/src/block.rs
```diff
@@ -0,0 +1,41 @@
...
+/// Store a block into IFPS.
+#[tracing::instrument(skip_all)]
+pub async fn put<T, C>(client: T, codec: C, data: Vec<u8>) -> Result<Cid, Error>
+where
+    T: IpfsDep,
+    C: Codec,
+{
+    let hash = Code::Sha2_256.digest(&data);
+    let cid = Cid::new_v1(codec.into(), hash);
+    client.put(cid, data.into(), vec![]).await?;
```
smrz2001, line 30: Wouldn't this return the CID so you don't have to hash the data above?

ceramic-kubo-rpc/src/http/block.rs
```diff
@@ -0,0 +1,291 @@
...
+    data: web::Data<AppState<T>>,
+    query: web::Query<PutQuery>,
+    mut payload: Multipart,
+) -> Result<HttpResponse, Error>
+where
+    T: IpfsDep,
+{
+    if query.pin {
+        return Err(Error::Invalid(anyhow!(
+            "recursive pinning is not supported."
```
smrz2001, line 83: Does `pin` being `true` here always mean recursively?

nathanielc: Yes, according to their docs https://docs.ipfs.tech/reference/kubo/rpc/#api-v0-block-put We might need to implement recursive pin but I am hoping not too. So much of our data points back to the beginning of time that its dangerous to recursively pin so for now I have explicitly denied any recursive pin behavior.

ceramic-kubo-rpc/src/block.rs
nathanielc: The beetle API needs a CID in order to put the data in the first place.

zachferland (approved):

ceramic-kubo-rpc/src/block.rs
smrz2001: Oh ok, huh.

smrz2001 (approved):
🚢

PR #14, opened 2023-03-23
feat: adds id endpoint and adds dag-jose resolving support

nathanielc:
Prior to this change dag/get and dag/resolve requests could not follow links into DAG-JOSE encoded nodes. This is now supported by implementing a simple resolver.

Additionally the `id` endpoint has been added and various bugs fixes have been made.

### DIFF
--- a/ceramic-kubo-rpc/src/http/dag.rs
+++ b/ceramic-kubo-rpc/src/http/dag.rs
@@ -3,6 +3,7 @@ use std::{io::Cursor, str::FromStr};
 use actix_multipart::Multipart;
 use actix_web::{http::header::ContentType, web, HttpResponse, Scope};
 use anyhow::anyhow;
+use dag_jose::DagJoseCodec;
 use futures_util::StreamExt;
 use iroh_api::IpfsPath;
 use libipld::{cbor::DagCborCodec, ipld, json::DagJsonCodec, prelude::Encode};
@@ -11,7 +12,7 @@ use serde::Deserialize;
 use crate::{
     dag,
     error::Error,
-    http::{AppState, DAG_CBOR, DAG_JSON},
+    http::{AppState, DAG_CBOR, DAG_JOSE, DAG_JSON},
     IpfsDep,
 };
 pub fn scope<T>() -> Scope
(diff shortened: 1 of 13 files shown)


ceramic-kubo-rpc/src/http/dag.rs
```diff
@@ -107,6 +110,24 @@ where
                     )
                     .await?
                 }
+                (DAG_CBOR, DAG_CBOR) => {
```
smrz2001, line 113: Is there a way to capture the combinations more generically?

nathanielc: The`dag::put`function takes generic type parameters for the input and output codecs. This means that we need to enumerate which ones we support so the Rust compiler can compile each combination of codecs that we use. Therefore this match expression makes it very clear which combinations we support. 

We could change the design of the dag::put function to not take generic type parameters and instead take a dynamic value that is known to implements the codec trait. This would allow us to pass any combination of codecs at runtime. This comes at the cost of requiring dynamic dispatch to call the implementations of the codec functions. This code is not in a hot path so we would likely not notice the cost of dynamic dispatch in this case. However its simple enough to enumerate the finite set of codecs we support so for now the type parameter design made the most sense.

ceramic-kubo-rpc/src/lib.rs
```diff
@@ -160,3 +203,76 @@ impl IpfsDep for Api {
...
+            if let Ipld::Link(c) = current.data {
+                current = self.load_cid(c).await?;
+            }
+            // Parse part as an integer and if that fails parse as a string into an index.
+            let index: libipld::ipld::IpldIndex = if let Ok(i) = part.parse::<usize>() {
+                i.into()
+            } else {
+                part.clone().into()
+            };
+            current.data = current.data.take(index).map_err(|_| {
```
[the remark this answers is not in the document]
nathanielc: Yes, the lines above distinguish between numeric indices and named parts. The code assumes that a numeric part is an index and anything else is a map key. That is the assumption beetle was already making so I assumed its safe to make.

PR #24, opened 2023-04-04
chore: add dockerfile for running ceramic-one

nathanielc:
This change adds a Dockerfile that builds an image to run the ceramic-one binary.

Additionally CI workflows are added to publish a new version of the image on pushes to main.


smrz2001 (approved):
🚀

PR #25, opened 2023-04-04
chore: remove stutter in directory names

nathanielc:
This PR moves code around such that top level directory names do not repeat the `ceramic` prefix.


AaronGoldman (approved):
looks strait forward to me

smrz2001 (commented):
🚀

PR #31, opened 2023-04-25
feat: adds eye command

nathanielc:
The `eye` command to the `ceramic-one` CLI starts a normal daemon and then subscribes to a gossipsub ceramic topic locally pinning any tips it learns about.

The `eye` behavior along with the existing metrics and traces has proven to be a useful tool to understand how ceramic-one is behaving.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -65,7 +65,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "465a6172cf69b960917811022d8f29bc0b7fa1398bc4f78b3c466673db1213b6"
 dependencies = [
  "quote",
- "syn 1.0.109",
+ "syn",
 ]
 
 [[package]]
@@ -211,7 +211,7 @@ dependencies = [
  "actix-router",
  "proc-macro2",
  "quote",
- "syn 1.0.109",
+ "syn",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


ceramic-one/src/main.rs
```diff
@@ -19,72 +31,231 @@ struct Cli {
...
+                let msg: Message = serde_json::from_slice(message.data.as_slice())
+                    .expect("should be json message");
+                info!(?msg);
+                match &msg {
+                    Message::Update {
+                        stream: _,
+                        tip,
+                        model: _,
+                    } => {
+                        if let Ok(ipfs_path) = IpfsPath::from_str(tip) {
```
oed, line 209: The `tip` should always be a CID? Why are we using IpfsPath? (I'm not even sure what this function does)

ceramic-one/src/main.rs
```diff
@@ -19,72 +31,231 @@ struct Cli {
...
+                            // Spawn task to get the data for a stream tip when we see one
+                            let client = client.clone();
+                            let metrics = metrics.clone();
+                            task::spawn(async move {
+                                let result = dag::get(client, &ipfs_path, DagJsonCodec).await;
+                                metrics.record(&result);
```
oed, line 215: What's the purpose of only getting this object? Also I could sometimes be a DagJWS encoded object, so not sure always using DagJSON would work?

If we want to log and have metrics around the entire event we need to check if it's a DagJWS object and if so also fetch it's payload and maybe CACAO if used. 
I suspect this might be out of scope of what you are trying to do here though?

ceramic-one/src/main.rs
nathanielc: Correct the tip is always a CID, however the API to fetch a block expects a path so we create a path that is just the CID.

ceramic-one/src/main.rs
nathanielc: `dag::get` gets the block from IPFS and reencodes it in whatever codec you want, in this case we are asking for the block back in DAG-JSON format. If the block itself is encoded as DAG-JOSE then we correctly decode it and reencode it. 

However this code doesn't really do anything with the result except count if we were successful or not in retrieving the block.

PR #32, opened 2023-04-28
feat: add /metrics endpoint

nathanielc:
This change exposes a Prometheus /metrics endpoint to enable metric collection via scraping.

Additionally a refactor to the how Beetle is consumed was performed. It was becoming very time consuming to trace Beetle code paths through many unneeded abstractions (i.e. unixfs). As such the refactor means that we now consume a minimal set of dependent crates from Beetle.

This has the advantage that is now very clear how these systems interact. For example publishing DHT records or when/how IPLD blocks are added to the local store.

Specifically we now depend on:

* iroh-metrics
* iroh-p2p
* iroh-rpc-client
* iroh-rpc-types
* iroh-store

Previously we were dependent on:

* iroh-api
* iroh-bitswap
* iroh-car
* iroh-embed
* iroh-gateway
* iroh-metrics
* iroh-one
* iroh-p2p
* iroh-resolver
* iroh-rpc-client
* iroh-rpc-types
* iroh-store
* iroh-unixfs
* iroh-util

This change comes with some small duplicated code to setup the p2p and store services. However this is a small amount of code with a clear purpose.

### DIFF
--- a/kubo-rpc/src/http/pubsub.rs
+++ b/kubo-rpc/src/http/pubsub.rs
@@ -161,7 +161,6 @@ mod tests {
     use std::{collections::HashMap, str::FromStr};
 
     use crate::{
-        error::Error,
         http::tests::{assert_body_json, assert_body_json_nl, build_server},
         IpfsDep,
     };
@@ -236,6 +235,9 @@ mod tests {
 
     #[actix_web::test]
     async fn test_subscribe() {
+        use crate::{
+            Bytes, Cid, Error, GossipsubEvent, IpfsPath, Ipld, Multiaddr, PeerId, PeerInfo,
+        };
         // Can't use unimock because it expects the stream to be Sync.
         // So we create a one off implementation of the trait and only implement the method we
         // need.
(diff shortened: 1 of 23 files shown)


mxinden:
Let me know in case you have any questions around `prometheus-client` crate. Great to see more metrics out there!

kubo-rpc/src/pin.rs
```diff
@@ -24,12 +21,8 @@ pub async fn remove<T>(_client: T, ipfs_path: &IpfsPath) -> Result<Cid, Error>
 where
     T: IpfsDep,
 {
-    if let Some(cid) = ipfs_path.cid() {
-        // Beetle does not have any garbage collection for its store so everything is pinned.
-        // Therefore we do not need to track which blocks are pinned.
-        // Do nothing
-        Ok(*cid)
-    } else {
-        Err(Error::Invalid(anyhow!("IPFS path does not have a CID")))
```
smrz2001, line 33: This error isn't possible anymore?

kubo-rpc/src/http/pubsub.rs
```diff
@@ -250,60 +246,80 @@ mod tests {
         #[async_trait]
         impl IpfsDep for TestDeps {
             /// Get information about the local peer.
-            async fn lookup_local(&self) -> Result<PeerInfo, Error> {
+            async fn lookup_local(&self) -> Result<crate::PeerInfo, Error> {
```
dbcfd, line 249: Does it make sense to make a use for these, so the changes are less?

`use crate::{Bytes, Cid, Error, GossipsubEvent, MultiAddr, PeerId, PeerInfo};`

kubo-rpc/src/http/swarm.rs
```diff
@@ -64,35 +63,92 @@ where
...
+
+    let (addrs, peer_ids) =
+        itertools::process_results(iter, |iter| iter.unzip::<_, _, Vec<_>, Vec<_>>())?;
+
+    // Check we found exactly one unique peer id.
+    let peer_ids: HashSet<PeerId> = peer_ids
+        .into_iter()
+        .flat_map(|p| if let Some(p) = p { vec![p] } else { Vec::new() })
+        .collect();
+    let peer_id = match peer_ids.len() {
```
dbcfd, line 140: `split_first` instead?

dbcfd (changes requested):

kubo-rpc/src/http/swarm.rs
nathanielc: What is `split_first`?

kubo-rpc/src/http/pubsub.rs
nathanielc: Yeah I'll do that.

kubo-rpc/src/pin.rs
nathanielc: Correct, previously the code was using the `IpfsPath` type from Iroh that was either an path rooted in a CID or a path rooted in an IPNS name. The new code now uses an `IpfsPath` implemented within the crate that only allows the IPFS (cid) use case.

one/src/ipfs.rs
```diff
@@ -0,0 +1,166 @@
...
+        } else {
+            info!("Creating store at {}", config.path.display());
+            Store::create(config)
+                .await
+                .context("failed to create new store")?
+        };
+
+        let rpc_addr = addr.clone();
+        let task =
+            tokio::spawn(async move { iroh_store::rpc::new(rpc_addr, store).await.unwrap() });
```
dbcfd, line 63: Not a big fan of the unwrap here, maybe split new into a new and run, so we spawn the run?

kubo-rpc/src/http/swarm.rs
nathanielc: I prefer the current implementation, as peer_ids is not a slice and it requires a few step to create a slice from the set that the current logic is the simpler option. And the match on the length makes it very clear which each case means.

dbcfd (approved):

PR #38, opened 2023-05-04
feat: ceramic http client

dbcfd:
Add a package which can talk Ceramic's http protocol. Additional include remote functionality to make http calls against a Ceramic server.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1231,6 +1231,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "ceramic-http-client"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "ceramic-event",
+ "json-patch",
+ "reqwest",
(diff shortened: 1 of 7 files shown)


oed:
I worry about including http client functionality for the js-ceramic server in this package. The goal of this project is to create a new http api that is purely focused on event streams. Are we sure that adding this won't cause confusion between these two things in the future?

oed:
> I was figuring that at some point rust ceramic would provide a similar api, at least for models and model instances. I can put this in its own repo if we decide that is not the case.

Models and model instances are not a concept we are planning to expose from the Ceramic API. These are concepts that only will exist on the ComposeDB layer.

nathanielc:
This change ended up as its own indpendent repo https://github.com/ceramicstudio/rust-ceramic-http-client/blob/main/README.md

PR #46, opened 2023-05-15
chore: only deny warnings during CI

nathanielc:
Additionally this change updates the Dockerfile to build using first class cache provided by Docker buildkit.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,27 +1,25 @@
-FROM public.ecr.aws/r5b3e0r5/3box/rust-builder:latest as chef
+FROM public.ecr.aws/r5b3e0r5/3box/rust-builder:latest as builder
 
 RUN mkdir -p /home/builder/rust-ceramic
 WORKDIR /home/builder/rust-ceramic
 
-# TODO add to parent builder image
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -1,27 +1,25 @@
-FROM public.ecr.aws/r5b3e0r5/3box/rust-builder:latest as chef
+FROM public.ecr.aws/r5b3e0r5/3box/rust-builder:latest as builder
 
 RUN mkdir -p /home/builder/rust-ceramic
 WORKDIR /home/builder/rust-ceramic
 
-# TODO add to parent builder image
-RUN cargo install cargo-chef
+# Use the same ids as the parent docker image by default
+ARG UID=1001
```
AaronGoldman, line 7: Does this make it fragile to a parent docker image with more than one user?

AaronGoldman (approved):

Dockerfile
AaronGoldman: I don't see why this is necessary. If the permissions on the folder are forgiving to need to be a specific user.

PR #53, opened 2023-06-06
feat: Serialize eventid

AaronGoldman:
Rename
  `Hash` to `AssociativeHash`
  `AHash` to `ShaSum256`

implement `std::convert::From;` `From<String>` for `ShaSum256`

implement `fn event_id()` to Serialize event data to an eventid.


recon/src/ahash.rs
```diff
@@ -1,17 +1,18 @@
...
+use std::convert::From;
 use std::fmt::{self, Debug, Display, Formatter};
 
-use crate::Hash;
+use crate::AssociativeHash;
 
 /// AHash an associative hash function for use in set reconciliation
 #[derive(Default, PartialEq, Clone, Copy, Deserialize)]
-pub struct AHash(#[serde(deserialize_with = "deserialize_from_bytes")] [u32; 8]);
+pub struct ShaSum256(#[serde(deserialize_with = "deserialize_from_bytes")] [u32; 8]);
```
oed, line 10: Would it make more sense to call it *SumSha256* or *Sha256Sum*? That way we don't break up "sha256"

recon/src/eventid.rs
```diff
@@ -0,0 +1,198 @@
...
+    prev_timestamp: u64,
+    event_height: u64,
+    event_cid: &Cid,
+) -> String {
+    let event_data = EventData {
+        concat_bytes: [
+            encode::u64(network_id.to_u64(), &mut [0_u8; 10]),
+            last16_bytes(separator.as_bytes()),
+            last16_bytes(&sha256_digest(controller)),
+            last16_bytes(&init.to_bytes()),
```
oed, line 88: We wanted this to be last 8 bytes (or event last 4 bytes)?

recon/src/eventid.rs
```diff
@@ -0,0 +1,198 @@
...
+/// event_id builds a Vec<u8> with the event id data.
+pub fn event_id(
+    network_id: Network,
+    separator: &str,
+    controller: &str,
+    init: &Cid,
+    prev_timestamp: u64,
+    event_height: u64,
+    event_cid: &Cid,
+) -> String {
```
oed, line 82: Why does this return a string? Don't we want to send raw bytes over the wire?

recon/src/eventid.rs
```diff
@@ -0,0 +1,175 @@
+use cbor::Encoder;
+use cid::Cid;
+use multihash::{Hasher, Sha2_256};
+/// EventId generates EventIDs from event data.
+///
+/// varint(0xce) + // streamid, 1 byte
+/// varint(0x30) + // dag-cbor, 1 byte
```
oed, line 7: Why 0x30? Also, I guess it should no longer say dag-cbor?

recon/src/eventid.rs
```diff
@@ -0,0 +1,175 @@
...
+/// varint(networkId), // 1 byte (5 for local network)
+/// last8Bytes(separator_value), // 16 bytes
```
oed, line 9:
```suggestion
/// last8Bytes(sha256(separator_value)), // 16 bytes
```

recon/src/eventid.rs
AaronGoldman: When we moved from `dag-cbor` to `multicodec` I changed the number to `0x30` the multicode for `multicodec` itself.

I have no preference between
`0x30` multicodec
`0x05` next available
`0x7c` for cip-124 (0x7c)

`0x00` - `0x04` are taken
`0x80` - u64 are more than one byte.
so any number `0x05` - `0x7f` is fine by me 🤷‍♂️

oed: What do you mean that we moved "to multicodec"? We use `dag-cbor` because that's how we encoded the payload. I don't understand what `multicodec` signifies here

AaronGoldman: we picked `0x05`

recon/src/ahash.rs
```diff
@@ -126,7 +128,35 @@ impl crate::recon::Hash for AHash {
...
+}
+
+impl From<&str> for Sha256Sum {
+    fn from(s: &str) -> Self {
+        Sha256Sum::digest(s)
+    }
+}
+
+impl Sha256Sum {
     fn sha256_digest(s: &str) -> [u8; 32] {
```
nathanielc, line 160: Why does this method not return a `Sha256Sum` type?

recon/src/eventid.rs
```diff
@@ -0,0 +1,177 @@
...
+
+/// event_id builds a Vec<u8> with the event id data.
+pub fn event_id(
+    network_id: Network,
+    separator: &str,
+    controller: &str,
+    init: &Cid,
+    event_height: u64,
+    event_cid: &Cid,
+) -> Vec<u8> {
```
nathanielc, line 65: Should we create an `EventId` type? And make this method the `EventId::new` method?

nathanielc (changes requested):

recon/src/eventid.rs
AaronGoldman: done

recon/src/libp2p.rs
```diff
@@ -77,13 +82,13 @@ impl Recon for Arc<Mutex<crate::recon::Recon>> {
     }
 }
 
-/// Config specifies the configurable properties of the Behaviour.
-#[derive(Clone)]
+/// Config specifies the configurable properties of the Behavior.
+#[derive(Clone, Debug)]
```
AaronGoldman, line 86: warn(missing_debug_implementations) detected a missing debug

recon/src/libp2p/handler.rs
```diff
@@ -22,6 +22,7 @@ use tracing::debug;
 
 use crate::libp2p::{protocol, Recon, PROTOCOL_NAME};
 
+#[derive(Debug)]
```
AaronGoldman, line 25: warn(missing_debug_implementations) detected a missing debug

recon/src/eventid.rs
```diff
@@ -0,0 +1,185 @@
...
+impl Network {
+    /// network.to_u64() to get the network as a u64
+    pub fn to_u64(&self) -> u64 {
+        match self {
+            // https://github.com/ceramicnetwork/CIPs/blob/main/tables/networkIds.csv
+            Network::Mainnet => 0x00,
+            Network::TestnetClay => 0x01,
+            Network::DevUnstable => 0x02,
+            Network::InMemory => 0xff,
+            Network::Local(id) => 0x01_0000_0000_u64 + u64::from(*id),
```
zachferland, line 52: I forgot to add local to js impl

zachferland (approved):
lgtm!

recon/src/lib.rs
```diff
@@ -1,12 +1,14 @@
 //! Recon is a network protocol for set reconciliation
-#![deny(missing_docs)]
+#![warn(missing_docs, missing_debug_implementations, clippy::all)]
```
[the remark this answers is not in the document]
AaronGoldman: Resolving so we can merge.
missing_debug_implementations was not being checked
 clippy::all was in the makefile but when I called build, I did not get the warnings.
 This way I can resolve the clippy warnings with context rather than as a pre push check

recon/src/libp2p.rs
```diff
@@ -2,12 +2,12 @@
 //!
 //! There are various types within this module and its children.
 //!
-//! Behaviour - Responsible for coordiating the many concurrent Recon syncs
+//! Behavior - Responsible for coordinating the many concurrent Recon syncs
```
nathanielc, line 5: nit, the libp2p crate uses the British spelling `Behaviour` while less common we should use it for consistency. Also this comment referes to a type name `Behaviour` so the spelling is important. I should have used the link syntax to make that clear. I'll do that in another PR.

nathanielc (approved):
Looks great, one small nit on the spelling changes otherwise its good to go

PR #60, opened 2023-07-13
feat: add openapi implementation of events api

nathanielc:
Creates simple API with methods to create and event and subscribe to events.

Refactors the EventId type a bit to make it easier to construct fencepost events.


nathanielc:
@AaronGoldman heads up on reviewing, all code within the `api-server` directory is generated code from OpenAPI.

api/src/server.rs
```diff
@@ -0,0 +1,388 @@
...
+            )
+        };
+        let start = start_builder.with_min_event_height().build_fencepost();
+        let stop = stop_builder.with_max_event_height().build_fencepost();
+
+        debug!(%start, %stop, "subscribe");
+        Ok(CeramicSubscribeSortValueGetResponse::Success(
+            self.recon
+                .range((Bound::Included(start), Bound::Excluded(stop)))
+                .into_iter()
```
oed, line 172: What order will this sort in?

nathanielc: Ascending lexical order of EventId

oed: So this api currently returns all known eventIds in the requested range?

nathanielc: yes, that is the intent. Should some event Ids be skipped?

oed: So how would this replace pubsub then? Don't we need the offset logic here?

oed: The way I imagined this working is:
1. js-ceramic subscribes to the range
2. rust-ceramic returns all known ranges
3. js-ceramic subscribes again to get new updates

We talked about 3 using long polling to be able to get new events as they come in. Am I missing something about how this as implemented currently would be used?

core/src/event_id.rs
```diff
@@ -1,100 +1,70 @@
...
+//!   0x12 sha2-256, 1byte
+//!   0x20 varint(hash length), 1 byte
+//!   hash bytes, 32 bytes
 #![warn(missing_docs, missing_debug_implementations, clippy::all)]
 
 use cbor::Encoder;
-use cid::Cid;
-use multihash::{Hasher, Sha2_256};
+use cid::{
+    multihash::{Hasher, Sha2_256},
```
AaronGoldman, line 20: Why is it better to import multihash from cid then directly?

one/src/main.rs
```diff
@@ -81,27 +95,50 @@ struct DaemonOpts {
...
-    network_id: String,
+    #[arg(long, default_value = "testnet-clay", env = "CERAMIC_ONE_NETWORK")]
+    network: Network,
+
+    /// Unique key used to find other Ceramic peers via the DHT
+    #[arg(long, env = "CERAMIC_ONE_LOCAL_NETWORK_ID")]
+    local_network_id: Option<u32>,
+
+    /// When true mdns will be used to discover peers.
+    #[arg(long, default_value_t = false, env = "CERAMIC_ONE_MDNS")]
```
AaronGoldman, line 106: not sure what the best default value is here

one/src/main.rs
```diff
@@ -81,27 +95,50 @@ struct DaemonOpts {
...
+    mdns: bool,
+}
+
+#[derive(ValueEnum, Debug, Clone)]
+enum Network {
```
AaronGoldman, line 111: why is this not the enum from core/src/network ?

one/src/main.rs
```diff
@@ -181,22 +223,25 @@ impl Daemon {
         debug!(?p2p_config, "using p2p config");
 
         // Construct a recon implementation.
-        let recon = Arc::new(std::sync::Mutex::new(recon::Recon::from_set([].into())));
+        let recon = Arc::new(Mutex::new(Recon::from_set([].into())));
```
AaronGoldman, line 226: This implies recon should have a default to call.

one/src/main.rs
```diff
@@ -210,6 +255,15 @@ impl Daemon {
         let srv_handle = srv.handle();
         tokio::spawn(srv);
 
+        let api_bind_address = self.api_bind_address.clone();
+        // Start Ceramic API
+        let network = self.network.clone();
+        tokio::spawn(ceramic_api::start(
```
AaronGoldman, line 261: Does this mean we need a thread safe mutex on recon?

Makefile
```diff
@@ -14,6 +14,17 @@ build:
 	# Build with all features
 	cargo build --locked --all-features
 
+# Generates api-server crate from ceramic.yaml OpenAPI spec
+.PHONY: api-server
+api-server: api/ceramic.yaml
```
AaronGoldman, line 19: should  `api-server` be two phonys? `generate-api-server` and `check-api-server` so the `all` can check that the generated code is yonger then the `api/ceramic.yaml`

AaronGoldman (commented):
I think this looks good but there is a lot here might we worth walking through live

core/src/event_id.rs
```diff
@@ -149,9 +118,211 @@ fn last4_bytes(buf: &[u8]) -> &[u8] {
...
+            },
+        }
+    }
+}
+impl Builder<WithInit> {
+    pub fn with_min_event_height(mut self) -> Builder<WithEventHeight> {
+        let mut event_height_cbor = Encoder::from_memory();
+        event_height_cbor.encode([0]).unwrap();
+        // event_height cbor unsigned int
+        self.state.bytes.extend(event_height_cbor.as_bytes());
```
AaronGoldman, line 270: `event_height_cbor.as_bytes()` here is a constant spesificly the null byte `b"\0"`

core/src/event_id.rs
```diff
@@ -149,9 +118,211 @@ fn last4_bytes(buf: &[u8]) -> &[u8] {
...
+            },
+        }
+    }
+    pub fn with_max_event_height(mut self) -> Builder<WithEventHeight> {
+        let mut event_height_cbor = Encoder::from_memory();
+        // TODO: Use a smarter value here instead of u64::MAX to keep the event height length
+        // smaller.
+        event_height_cbor.encode([u64::MAX]).unwrap();
+        // event_height cbor unsigned int
+        self.state.bytes.extend(event_height_cbor.as_bytes());
```
AaronGoldman, line 283: The first byte of an `event_height_cbor` will always be 27 or less.
We could use the single byte `b"\xFF"` as a byte that will be grater then a 
`1B FFFFFFFFFFFFFFFF \\ event_height_cbor.encode([u64::MAX])`

`b"\xFF"` is the break stop code in CBOR.

core/src/event_id.rs
```diff
@@ -1,100 +1,70 @@
...
-                last8_bytes(&sha256_digest(separator)),       // separator [u8; 8]
-                last8_bytes(&sha256_digest(controller)),      // controller [u8; 8]
-                last4_bytes(init.to_bytes().as_slice()),      // StreamID [u8; 4]
-                event_height_cbor.as_bytes(),                 // event_height cbor unsigned int
-                //varint(event_height, &mut [0_u8; 10]), // event_height varint
-                event_cid.to_bytes().as_slice(), // [u8]
-            ]
-            .concat(),
-        )
+        EventId::builder()
```
AaronGoldman, line 50: May be easier to just add my functon to parce out the CID here rather than creating a merge conflict 
```rust
use unsigned_varint::{decode::u64 as de_varint, encode::u64 as varint};

/// try to parse a CID out of a CIP-124 EventID
    pub fn cid(&self) -> Option<Cid> {
        let (streamid, remainder) = de_varint(&self.0).unwrap_or_default();
        if streamid != 0xce {
            return None; // not a streamid
        };

        let (typeid, remainder) = de_varint(remainder).unwrap_or_default();
        if typeid != 0x05 {
            return None; // not a CIP-124 EventID
        };

        let (_network_id, mut remainder) = de_varint(remainder).unwrap_or_default();

        // strip separator [u8; 8] controller [u8; 8] StreamID [u8; 4]
        remainder = &remainder[(8 + 8 + 4)..];

        // hight cbor unsigned integer
        if remainder[0] <= 23 {
            // 0 - 23
            remainder = &remainder[1..]
        } else if remainder[0] == 24 {
            // u8
            remainder = &remainder[2..]
        } else if remainder[0] == 25 {
            // u16
            remainder = &remainder[3..]
        } else if remainder[0] == 26 {
            // u32
            remainder = &remainder[5..]
        } else if remainder[0] == 27 {
            // u64
            remainder = &remainder[9..]
        } else {
            // not a cbor unsigned int
            return None;
        };
        match Cid::read_bytes(remainder) {
            Ok(v) => Some(v),
            Err(_) => None, // not a CID
        }
    }
}
```

core/src/event_id.rs
```diff
@@ -149,9 +118,211 @@ fn last4_bytes(buf: &[u8]) -> &[u8] {
...
+/// Builder with event CID set.
+#[derive(Debug)]
+pub struct WithEvent {
+    bytes: Vec<u8>,
+}
+impl BuilderState for WithEvent {}
+
+impl Builder<Init> {
+    pub fn with_network(self, network: &Network) -> Builder<WithNetwork> {
+        // TODO what is max event id size?
```
AaronGoldman, line 183: varint(0xce) + // streamid, `1` byte
varint(0x05) + // cip-124 EventID, `1` byte
varint(networkId), // `5` bytes for local network
last8Bytes(sha256(separator_value)), // `8` bytes
last8Bytes(sha256(stream_controller_DID)), // `8` bytes
last4Bytes(init_event_CID) // `4` bytes
cbor(eventHeight), // u64_max `9` bytes
eventCID ? mostly `36` bytes but could be inline CID 

`1 + 1 + 5 + 8 + 8 + 4 + 9 + ?`
`36 + ?`
assuming 36 bytes for 32 byte hash CID
`36 + 36`
`72`

Makefile
```diff
@@ -14,6 +14,17 @@ build:
...
+	# Generate crate
+	npx @openapitools/openapi-generator-cli generate -i api/ceramic.yaml -g rust-server  --additional-properties=packageName=ceramic-api-server -o api-server
+	# Add missing clippy allow directive to example code
+	echo "#![allow(suspicious_double_ref_op)]" | cat - ./api-server/examples/server/server.rs > ./api-server/examples/server/server.rs.tmp
```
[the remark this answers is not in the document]
nathanielc: Would a script to generate the server code be better? Then we call the script from make?

README.md
```diff
@@ -29,6 +29,10 @@ Use the provided `Makefile` for basic actions to ensure your changes are ready f
 Using the makefile is not necessary during your developement cycle, feel free to use the relvant cargo commands directly.
 However running `make` before publishing a PR will provide a good signal if you PR will pass CI.
 
+### Generating `ceramic-api-server`
+
+The `ceramic-api-server` crate is generated using OpenAPI.
+Install `@openapitools/openapi-generator-cli` and use the make target `api-server` to generate the code anytime `api/ceramic.yaml` is modified.
```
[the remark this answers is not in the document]
nathanielc: Yeah I like your idea of a `check` for the generated code. I'll add a script for that.

core/src/event_id.rs
nathanielc: This guarantees we are using a compatible version of multihash with the cid crate.

one/src/main.rs
nathanielc: Mdns is not supported very well outside of LANs in home use. Generally datacenter/cloud networks do not allow mdns, as such its only valuable to the dev on their LAN if they are running multiple nodes. Since this seems like a narrow use case I decided default to off is best to reduce surface area of issues.

one/src/main.rs
nathanielc: This enum requires that it implement the ValueEnum trait so it can be used as a CLI argument. This decouples CLI parsing from the actual network enum. Specifically the CLI parsing does not allow enum variants that contain values so the `core::Network::Local(u32)` variant is not allowed to implement the ValueEnum trait. So this provides a clean mapping between them.

one/src/main.rs
nathanielc: Yes creating a new empty Recon would be a good function to add.

one/src/main.rs
nathanielc: Yes that is why we are using `Arc<Mutex<Recon>>`

core/src/event_id.rs
nathanielc: This looks great, I'll just add it to my PR.

api-server/api/openapi.yaml
```diff
@@ -0,0 +1,80 @@
...
+    This is the Ceramic API for working with streams and events
+  license:
+    name: MIT
+    url: https://mit-license.org/
+  title: Ceramic API
+  version: 0.1.0
+servers:
+- url: /
+paths:
+  /ceramic/subscribe/{sort-value}:
```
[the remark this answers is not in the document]
nathanielc: yes, its unused right now but I'll add a placeholder for it

nathanielc:
@AaronGoldman @oed Thanks for the reviews. I have addressed all comments.

PR #61, opened 2023-07-18
feat: add interest synchronization and update Recon 

nathanielc:
With these changes ceramic-one can do the following:

1. Maintain a set of interest ranges for itself and other peers
2. Synchronize interest ranges among peers
3. Use Recon to sync keys that exist within intersecting interest ranges.
4. Update its local interest when a subscription is requested.

Some things that are still lacking:

1. Interests are not expired
2. Interests are never removed from the local node
3. Interests are NOT signed but still shared with peers transitively 

This PR is large enough, I suggest we create issues for the above and address them in subsequent smaller PRs.

### DIFF
--- a/api/src/server.rs
+++ b/api/src/server.rs
@@ -229,7 +229,7 @@ where
         let interest = Interest::builder()
             .with_sort_key(&sort_key)
             .with_peer_id(&self.peer_id)
-            .with_range(start.as_slice()..stop.as_slice())
+            .with_range((start.as_slice(), stop.as_slice()))
             .with_not_after(0)
             .build();
         self.interest
@@ -393,7 +393,7 @@ mod tests {
                 Interest::builder()
                     .with_sort_key("model")
                     .with_peer_id(&peer_id)
-                    .with_range(start.as_slice()..end.as_slice())
+                    .with_range((start.as_slice(), end.as_slice()))
                     .with_not_after(0)
                     .build(),
             ))
@@ -456,7 +456,7 @@ mod tests {
                 Interest::builder()
                     .with_sort_key("model")
                     .with_peer_id(&peer_id)
-                    .with_range(start.as_slice()..end.as_slice())
+                    .with_range((start.as_slice(), end.as_slice()))
                     .with_not_after(0)
                     .build(),
             ))
@@ -522,7 +522,7 @@ mod tests {
                 Interest::builder()
                     .with_sort_key("model")
                     .with_peer_id(&peer_id)
-                    .with_range(start.as_slice()..end.as_slice())
+                    .with_range((start.as_slice(), end.as_slice()))
                     .with_not_after(0)
                     .build(),
             ))
@@ -588,7 +588,7 @@ mod tests {
                 Interest::builder()
                     .with_sort_key("model")
                     .with_peer_id(&peer_id)
-                    .with_range(start.as_slice()..end.as_slice())
+                    .with_range((start.as_slice(), end.as_slice()))
                     .with_not_after(0)
                     .build(),
             ))
--- a/api/src/server.rs
+++ b/api/src/server.rs
@@ -102,20 +102,20 @@ where
 }
 
 #[derive(Clone)]
-pub struct Server<C, IR, MR> {
+pub struct Server<C, I, M> {
     peer_id: PeerId,
     network: Network,
-    interest: IR,
-    model: MR,
+    interest: I,
+    model: M,
     marker: PhantomData<C>,
 }
 
-impl<C, IR, MR> Server<C, IR, MR>
+impl<C, I, M> Server<C, I, M>
 where
-    IR: Recon<Key = Interest>,
-    MR: Recon<Key = EventId>,
+    I: Recon<Key = Interest>,
+    M: Recon<Key = EventId>,
 {
-    pub fn new(peer_id: PeerId, network: Network, interest: IR, model: MR) -> Self {
+    pub fn new(peer_id: PeerId, network: Network, interest: I, model: M) -> Self {
         Server {
             peer_id,
             network,
@@ -132,11 +132,11 @@ use std::error::Error;
 use swagger::ApiError;
 
 #[async_trait]
-impl<C, IR, MR> Api<C> for Server<C, IR, MR>
+impl<C, I, M> Api<C> for Server<C, I, M>
 where
     C: Send + Sync,
-    IR: Recon<Key = Interest> + Sync,
-    MR: Recon<Key = EventId> + Sync,
+    I: Recon<Key = Interest> + Sync,
+    M: Recon<Key = EventId> + Sync,
 {
     async fn ceramic_events_post(
         &self,
--- a/api/src/server.rs
+++ b/api/src/server.rs
@@ -8,7 +8,7 @@ use async_trait::async_trait;
 use futures::{future, Stream, StreamExt, TryFutureExt, TryStreamExt};
 use hyper::server::conn::Http;
 use hyper::service::Service;
-use recon::{AssociativeHash, Key, Store};
+use recon::{AssociativeHash, InterestProvider, Key, Store};
 use std::{future::Future, ops::Range};
 use std::{marker::PhantomData, ops::RangeBounds};
 use std::{net::SocketAddr, ops::Bound};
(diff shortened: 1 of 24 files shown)


api/src/server.rs
```diff
@@ -209,9 +225,20 @@ where
         let start = start_builder.with_min_event_height().build_fencepost();
         let stop = stop_builder.with_max_event_height().build_fencepost();
 
+        // Update interest ranges to include this new subscription.
```
nathanielc, line 228: TODO: This is a naive way to populate interest ranges but it works for now.

nathanielc:
Note to reviews, there are 4 separate commits that each build on each other but mostly do not modify previous work. Reviewing each in separately may help digest the changes.

api/src/server.rs
```diff
@@ -95,17 +102,25 @@ where
...
+    interest: IR,
+    model: MR,
     marker: PhantomData<C>,
 }
 
-impl<C, R: Recon> Server<C, R> {
-    pub fn new(network: Network, recon: R) -> Self {
+impl<C, IR, MR> Server<C, IR, MR>
+where
+    IR: Recon<Key = Interest>,
```
dbcfd, line 115: Why `IR` and `MR`? Guessing it's for "InterestRecon" and "ModelRecon". Maybe just use `I` and `M` or `Interest` and `Model`?

core/src/interest.rs
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    /// Return the interest as a slice of bytes.
+    pub fn as_slice(&self) -> &[u8] {
+        self.0.as_slice()
+    }
+}
+
+fn decode_bytes(decoder: &mut Decoder<Cursor<Vec<u8>>>) -> Result<Vec<u8>> {
+    if let Some(item) = decoder.items().next() {
```
dbcfd, line 69: Maybe `let item = decoder.items.next().ok_or_else(|| anyhow!("expected top level cbor value, found nothing"))?`

core/src/interest.rs
```diff
@@ -0,0 +1,275 @@
...
+    encoder: Encoder<Vec<u8>>,
+}
+impl BuilderState for WithNotAfter {}
+
+impl Builder<Init> {
+    pub fn with_sort_key(self, sort_key: &str) -> Builder<WithSortKey> {
+        let mut hasher = Sha2_256::default();
+        hasher.update(sort_key.as_bytes());
+        // sha256 is 32 bytes safe to unwrap to [u8; 32]
+        let hash: [u8; 32] = hasher.finalize().try_into().unwrap();
```
dbcfd, line 148: So later on, this gets converted to a vec, just taking the last 8 items. Should just return the finalize, then take a part of the vec.

p2p/src/node.rs
```diff
@@ -129,15 +145,20 @@ impl<S: Storage, R: Recon> Drop for Node<S, R> {
...
+    S: Storage,
+    IR: Recon<Key = Interest, Hash = Sha256a>,
+    MR: Recon<Key = EventId, Hash = Sha256a>,
+{
     pub async fn new(
         config: Config,
         rpc_addr: P2pAddr,
         mut keychain: Keychain<S>,
-        recon: Option<R>,
+        recons: Option<(IR, MR)>,
```
dbcfd, line 161: This seems like it should at least be a type alias. I'm not sure if we'll need to add other recons, but at least that way it's a little easier to adjust.

recon/src/libp2p.rs
```diff
@@ -229,33 +249,34 @@ impl<R: Recon> NetworkBehaviour for Behaviour<R> {
     ) {
         match event {
             // The peer has started to synchronize with us.
-            FromHandler::Started => self.peers.entry(peer_id).and_modify(|info| {
-                if info.status != PeerStatus::Started {
-                    info.status = PeerStatus::Started;
-                    self.swarm_events_queue.push_front(Event::PeerEvent {
+            FromHandler::Started { stream_set } => self.peers.entry(peer_id).and_modify(|info| {
+                info.status = PeerStatus::Started { stream_set };
```
dbcfd, line 253: Originally there was a check here that we weren't already started, to prevent a double publish. Should that check be re-added? What is the expected behavior if the peer calls started twice, especially with different stream sets?

recon/src/recon.rs
```diff
@@ -21,114 +25,141 @@ use tracing::{instrument, trace};
...
+        let interests = self.interests()?;
+        let mut messages = Vec::with_capacity(interests.len());
+        for range in interests {
+            let mut response = Message::new(&range.start, &range.end);
+            if let Some((first, last)) = self.store.first_and_last(&range.start, &range.end) {
+                // Cannot use store.len, we need the count between first and last
+                let len = self.store.len();
+                match len {
+                    0 => {
+                        // this should be unreachable, but we explicitly add this case to be clear.
```
dbcfd, line 62: Although this should be unreachable, does it make sense to throw an error here in case dependent code breaks this assumption?

recon/src/recon.rs
```diff
@@ -21,114 +25,141 @@ use tracing::{instrument, trace};
...
     }
 
     /// Process an incoming message and respond with a message reply.
-    pub fn process_message(&mut self, received: &Message<K, H>) -> Result<Response<K, H>> {
+    #[instrument(skip_all, ret)]
+    pub fn process_messages(&mut self, received: &[Message<K, H>]) -> Result<Response<K, H>> {
+        // First we must find the intersection of interests.
+        // Then reply with a message per intersection.
+        let mut intersections: Vec<(RangeOpen<K>, BoundedMessage<K, H>)> = Vec::new();
+        for range in self.interests.interests()? {
```
dbcfd, line 93: Can we put a TODO here to explore more efficient alternatives to this implementation, with a ticket?

one/src/main.rs
```diff
@@ -226,25 +233,40 @@ impl Daemon {
...
         // Construct a recon implementation.
-        let recon = Arc::new(Mutex::new(ReconEvents::new(BTreeStore::default())));
+        let recon_interest = Arc::new(Mutex::new(Recon::new(
+            BTreeStore::default(),
+            FullInterests::default(),
+        )));
+        let recon_model = Arc::new(Mutex::new(Recon::new(
+            BTreeStore::default(),
+            // Use recon_interest as the InterestProvider for recon_model
+            recon_interest.clone(),
```
dbcfd, line 243: Given the multiple locks involved we should run this through https://lib.rs/crates/cargo-check-deadlock or the tool tokio uses (can't remember the name of it).

dbcfd (changes requested):

recon/src/recon.rs
nathanielc: In this case, the correct behavior is to do nothing anyway so that is why it doesn't through an error.

recon/src/recon.rs
nathanielc: Sure, I have some rough thoughts about using  a hierarchical tree something like (https://en.wikipedia.org/wiki/Bounding_volume_hierarchy) but in a single dimension. I'll add these to the TODO

one/src/main.rs
nathanielc: I tried using `cargo-check-deadlock` however it fails to compile in its latest release. While the idea is good, I don't think we need CI tooling for deadlocks yet. We only have two locks and they follow basic usage patterns. Its not possible for  deadlock currently as there is no cycle in the code between the locks. 

Additionally I expect we will be moving away from the locks in the medium terms as we look to make Recon more async.

core/src/interest.rs
```diff
@@ -0,0 +1,275 @@
+use anyhow::{anyhow, Result};
+use cbor::{CborBytes, Decoder, Encoder};
```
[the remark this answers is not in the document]
nathanielc: I have updated to using minicbor, and is a very positive change. Minicbor's focus on an allocation free API worked very well.

dbcfd (approved):

core/src/interest.rs
```diff
@@ -0,0 +1,275 @@
...
+use serde::{Deserialize, Serialize};
+use std::{fmt::Display, io::Cursor, str::FromStr};
+
+pub use libp2p_identity::PeerId;
+
+use crate::RangeOpen;
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize)]
+/// Interest declares an interest in a keyspace for a given peer.
+/// cbor([sort_key, peer_id, start_key, stop_key, not after])
```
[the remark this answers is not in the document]
nathanielc: @AaronGoldman and I chatted and agreed to leave it as is for now as it enables decoding portions of Interest without allocating for all fields.

PR #62, opened 2023-07-24
feat: dapp functionality

dbcfd:
Adds necessary functionality to support dApps with the http client. This functionality is

 * Ability to set a custom signer, for use with browser wallets
 * Ability to create JWS for non CID payloads

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1399,6 +1399,7 @@ name = "ceramic-core"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "async-trait",
  "base64 0.21.2",
  "cid 0.9.0",
  "did-method-key",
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -31,12 +32,12 @@ futures = "0.3"
...
-iroh-rpc-client = { git = "https://github.com/nathanielc/beetle", branch = "remove-p2p" }
-iroh-rpc-types = { git = "https://github.com/nathanielc/beetle", branch = "remove-p2p" }
-iroh-store = { git = "https://github.com/nathanielc/beetle", branch = "remove-p2p" }
-iroh-util = { git = "https://github.com/nathanielc/beetle", branch = "remove-p2p" }
+iroh-bitswap = { git = "https://github.com/nathanielc/beetle", branch = "custom-behaviour" }
+iroh-metrics = { git = "https://github.com/nathanielc/beetle", branch = "custom-behaviour" }
+iroh-rpc-client = { git = "https://github.com/nathanielc/beetle", branch = "custom-behaviour" }
+iroh-rpc-types = { git = "https://github.com/nathanielc/beetle", branch = "custom-behaviour" }
+iroh-store = { git = "https://github.com/nathanielc/beetle", branch = "custom-behaviour" }
+iroh-util = { git = "https://github.com/nathanielc/beetle", branch = "custom-behaviour" }
```
nathanielc, line 40: Why was this changed? The `custom-behaviour` branch is older than the `remove-p2p` branch. In fact we should be using `main`. I'll create a small PR to fix that and then we can rebase this.

nathanielc: Feel free to merge once https://github.com/3box/rust-ceramic/pull/65 merges and this is rebased.

dbcfd: Will do. `remove-p2p` branch was deleted, so wasn't able to build without switching.

PR #71, opened 2023-07-26
feat: recon durability

AaronGoldman:


core/src/event_id.rs
```diff
@@ -146,6 +146,12 @@ impl From<&Vec<u8>> for EventId {
     }
 }
 
+impl From<&str> for EventId {
```
dbcfd, line 149: This should be a TryFrom with checks that the event id as valid. Or added to EventId as `from_str_unchecked`.

recon/src/recon/durability.rs
```diff
@@ -0,0 +1,280 @@
...
+            row.get(3).unwrap(),
+            row.get(4).unwrap(),
+            row.get(5).unwrap(),
+            row.get(6).unwrap(),
+            row.get(7).unwrap(),
+        ];
+        H::from(bytes)
+    }
+
+    fn range(
```
dbcfd, line 171: So rather than changing the type of range, one way to do this is we return a SqliteRange type that then implements `Iterator<Item = &Self::Key>`.  The store trait then has a RangeType with those bounds.

```
trait Store {
    type Range;
    
    fn range(...) where Self::Range: for<'a> Iterator<Item = &'a Self::Key>;
}

impl Store for SqliteStore {
   type Range = SqliteRange
}```

Comment (dbcfd) [reply recon/src/recon/durability.rs:171]: Although I think with that, we'd probably want to make Range a trait, and then recon returns a Range, which then implements the functionality that Recon type has.

**Review by dbcfd — CHANGES_REQUESTED** [review]

File: core/src/event_id.rs [hunk 3]
```diff
@@ -146,6 +146,12 @@ impl From<&Vec<u8>> for EventId {
...
+    fn from(s: &str) -> Self {
+        EventId(s.as_bytes().to_vec())
```
nathanielc, line 151: A UTF-8 string is not a valid EventId, so unless we are decoding the string as some binary representation I don't think we should have a From<str> at all.

recon/src/sha256a.rs
```diff
@@ -160,6 +166,33 @@ impl From<&[u8; 32]> for Sha256a {
...
+
+impl From<&str> for Sha256a {
+    fn from(input: &str) -> Self {
+        Sha256a::digest(&Bytes::from(input))
+    }
+}
+
+impl Sha256a {
+    /// allocate a new hash with the state from the hex string
+    pub fn from_hex(hex_data: &String) -> Result<Sha256a, hex::FromHexError> {
```
nathanielc, line 189: I don't see any uses of this function.

nathanielc (changes requested):
Overall looks great, a few small points below.

core/src/event_id.rs
AaronGoldman: I was using this to make EventIDs for the test but now we that we
impl Key for Bytes
we just use Bytes for the test key and I can remove this.

recon/src/sha256a.rs
AaronGoldman: no longer used Ill remove it

recon/src/sha256a.rs
```diff
@@ -214,8 +209,15 @@ impl Key for Bytes {
...
+        // We need a value that sorts greater than any multiformat.
+        // Multiformats start with a multiformats.varint
+        // (https://github.com/multiformats/unsigned-varint)
+        // the max_value for a multiformat varint is 0x7fffffff_ffffffff (2^63-1)
+        // multiformats.varint.encode(0x7fffffff_ffffffff)
+        // b'\xff\xff\xff\xff\xff\xff\xff\xff\x7f'
+        // therefore all multiformat data will be less then
+        // b'\xff\xff\xff\xff\xff\xff\xff\xff\x80'
+        // [0xFF, 9]
         [0xFF; 1024].as_slice().into()
```
nathanielc, line 221: looks like a small typo here, the last line should be deleted and the last comment uncommented.

nathanielc (approved):

PR #73, opened 2023-07-28
refactor: remove mutex locking of Recon

nathanielc:
This change introduces a client/server model for communicating with Recon. The Server runs as a single async task and the clients send requests to the Server via async channels.

This change in design removes the need to use an Arc<Mutex<Recon>> as the client can be cheaply cloned and owned by the callers. As a result this change moves us towards a full async stack top to bottom so we are not mixing sync and async code.

Will wait till #71 merges and rebase this change before merging. PR #71 changes the iteration to be over owned values which will make this code better once it lands.


recon/src/recon/sqlitestore.rs
```diff
@@ -312,83 +281,87 @@ where
...
+        Ok(rows.get(0).map(|row| {
+            let bytes: Vec<u8> = row.get(0);
+            K::from(bytes)
+        }))
     }
 
     #[instrument(skip(self))]
-    fn first_and_last(
-        &self,
+    async fn first_and_last(
```
AaronGoldman, line 299: todo
```
SELECT first.key, last.key
FROM 
	(
		SELECT key
		FROM recon
		WHERE
            		sort_key = ? AND
            		key > ? AND key < ?
		ORDER BY cid ASC
		LIMIT 1
	) as first
JOIN
	(
		SELECT key
		FROM recon
		WHERE
            		sort_key = ? AND
            		key > ? AND key < ?
		ORDER BY cid DESC
		LIMIT 1
	) as last
```

AaronGoldman (approved):
This looks like a good refactor from rusqlite to sqlx.

recon/src/recon/sqlitestore.rs
```diff
@@ -215,59 +185,58 @@ where
...
+        );
+        let rows = query
+            .bind(&self.sort_key)
+            .bind(left_fencepost.as_bytes())
+            .bind(right_fencepost.as_bytes())
+            .bind(limit as i64)
+            .bind(offset as i64)
+            .fetch_all(&self.pool)
+            .await?;
         debug!(count = rows.len(), "rows");
```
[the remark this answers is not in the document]
nathanielc: No, rows is a vector of all results. We convert it into an iterator on the next line.

recon/src/recon/sqlitestore.rs
```diff
@@ -279,28 +248,28 @@ where
...
-            .map(|row| row.unwrap().clone());
-        let rows = rows.collect::<Vec<Self::Key>>();
-        rows.get(0).cloned()
+    ; ",
+        );
+        let rows = query
+            .bind(&self.sort_key)
+            .bind(left_fencepost.as_bytes())
+            .bind(right_fencepost.as_bytes())
+            .fetch_all(&self.pool)
```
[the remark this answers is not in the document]
nathanielc: `.fetch_one` expects exactly one row. Its possible that no rows are returned. `.fetch_one` will error if no rows are found were as `fetch_all` will just return an empty list that we can use `.get(0)` to return an Option.

recon/src/recon/sqlitestore.rs
nathanielc: I'll submit a follow up PR for this change.

api/src/server.rs
```diff
@@ -59,45 +59,41 @@ pub async fn create(
...
-    S: Store<Key = K, Hash = H> + Send,
-    I: InterestProvider<Key = K> + Send,
 {
     type Key = K;
 
-    fn insert(&self, key: &Self::Key) -> Result<()> {
-        self.lock()
-            .expect("should be able to acquire lock")
-            .insert(key)
+    async fn insert(&self, key: Self::Key) -> Result<()> {
```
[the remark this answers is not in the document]
nathanielc: Yep, I'll fix this to return the boolean value.

recon/src/libp2p/tests.rs
```diff
@@ -26,43 +26,40 @@ use libp2p::swarm::{keep_alive, Swarm, SwarmEvent};
...
-    let mut data = [0u8; 64];
+    let mut data = [0u8; 8];
     thread_rng().fill(&mut data);
     let hash = Code::Sha2_256.digest(&data);
-    Cid::new_v1(0x12, hash)
+    Cid::new_v1(0x00, hash)
 }
 
-fn build_swarm(name: &str, config: Config) -> SwarmTest {
+fn build_swarm(runtime: &Runtime, name: &str, config: Config) -> SwarmTest {
```
dbcfd, line 59: Why are we using runtime, and not just tokio::spawn?

recon/src/recon/tests.rs
```diff
@@ -1424,114 +1465,114 @@ fn message_cbor_deserialize_zero_hash() {
...
-        });
-        for row in r4? {
-            println!("{:?}", row.unwrap())
-        }
-
-        Ok(())
-    }
-    store().unwrap();
-}
+//#[test]
```
dbcfd, line 1468: Planning on bringing back, or was this moved and should be deleted?

recon/src/libp2p/tests.rs
nathanielc: Added a comment in the code.

dbcfd (approved):

PR #78, opened 2023-07-31
fix: regex char not allowed, change api names

zachferland:
Api names produced regex errors (dash char), changed names and generated api again

### DIFF
--- a/api-server/api/openapi.yaml
+++ b/api-server/api/openapi.yaml
@@ -10,13 +10,13 @@ info:
 servers:
 - url: /
 paths:
-  /ceramic/subscribe/{sort-key}/{sort-value}:
+  /ceramic/subscribe/{sortkey}/{sortvalue}:
     get:
       parameters:
       - description: name of the sort-key
         explode: false
         in: path
-        name: sort-key
+        name: sortkey
         required: true
         schema:
           type: string
@@ -24,7 +24,7 @@ paths:
       - description: value associated with the sort key
         explode: false
         in: path
-        name: sort-value
+        name: sortvalue
         required: true
         schema:
           type: string
--- a/api-server/api/openapi.yaml
+++ b/api-server/api/openapi.yaml
@@ -10,13 +10,13 @@ info:
 servers:
 - url: /
 paths:
-  /ceramic/subscribe/{sortkey}/{sortvalue}:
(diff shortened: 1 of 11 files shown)


api-server/src/server/mod.rs
```diff
@@ -21,27 +21,25 @@ pub use crate::context;
...
+use crate::{Api, CeramicEventsPostResponse, CeramicSubscribeSortkeySortvalueGetResponse};
 
 mod paths {
     use lazy_static::lazy_static;
 
     lazy_static! {
         pub static ref GLOBAL_REGEX_SET: regex::RegexSet = regex::RegexSet::new(vec![
             r"^/ceramic/events$",
-            r"^/ceramic/subscribe/(?P<sort-key>[^/?#]*)/(?P<sort-value>[^/?#]*)$"
+            r"^/ceramic/subscribe/(?P<sortkey>[^/?#]*)/(?P<sortvalue>[^/?#]*)$"
```
zachferland, line 32: "error: invalid capture group character" thrown before, removed dash

api/src/lib.rs
```diff
@@ -11,8 +11,8 @@ pub async fn start(
     peer_id: PeerId,
     network: Network,
     addr: String,
-    interest: impl Recon<Key = Interest>,
-    model: impl Recon<Key = EventId>,
+    interest: impl Recon<Key = Interest> + 'static,
```
zachferland, line 14: not sure what the compiler was complaining about here

api-server/src/server/mod.rs
nathanielc: Yeah I ran into that yesterday too, thanks for the fix.

nathanielc (changes requested):

api-server/api/openapi.yaml
```diff
@@ -10,21 +10,21 @@ info:
 servers:
 - url: /
 paths:
-  /ceramic/subscribe/{sort-key}/{sort-value}:
+  /ceramic/subscribe/{sortkey}/{sortvalue}:
```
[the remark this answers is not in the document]
zachferland: yeah see it makes auto names better, just updated in latest commit

nathanielc (approved):

AaronGoldman (approved):

PR #79, opened 2023-08-02
feat: separator from "sort_key|sort_key"

AaronGoldman:
The separator is not the Sha256 of not just the sort_key but the concatenation of
the  sort_key
Literal pipe "|"
and the sort_value.


oed:
Nice, do we have an issue to track updating the CIP as well?

core/src/event_id.rs
```diff
@@ -247,10 +248,10 @@ impl Builder<Init> {
 }
 impl Builder<WithNetwork> {
     // TODO sort_value should be bytes not str
-    pub fn with_sort_value(mut self, sort_value: &str) -> Builder<WithSortValue> {
-        self.state
-            .bytes
-            .extend(last8_bytes(&sha256_digest(sort_value)));
+    pub fn with_sort_value(mut self, sort_key: &str, sort_value: &str) -> Builder<WithSortValue> {
+        self.state.bytes.extend(last8_bytes(&sha256_digest(
+            &(sort_key.to_owned() + "|" + sort_value),
```
nathanielc, line 253: nit, it would require less allocations to instead pass the `sort_key`, `|` and the `sort_value` independently to the digest instead of doing the concatenation into a string and passing the whole string into the digest. Probably not a performance change that would matter :man_shrugging:

nathanielc (approved):
LGTM with one comment on possible improvement.

core/src/event_id.rs
AaronGoldman: sha256_digest is called else were using just one arg for the controller.

AaronGoldman:
> Nice, do we have an issue to track updating the CIP as well?

https://github.com/ceramicnetwork/CIPs/pull/144/

core/src/event_id.rs
dbcfd: We can actually save a couple allocations by using a format then

`&format!("{}|{}", sort_key, sort_value)`

PR #83, opened 2023-08-09
feat: add version endpoint to api server

dbcfd:


oed (commented):
Looks like we are using "node" in some places and "server" in others. Let's align.

api/build.rs
```diff
@@ -0,0 +1,9 @@
+fn main() {
+    let mut opts = built::Options::default();
+    opts.set_dependencies(true);
+
+    let src = project_root::get_project_root().unwrap();
```
AaronGoldman, line 5: What is saving this data for?

AaronGoldman:
LGTM

api/ceramic.yaml
```diff
@@ -79,8 +90,18 @@ components:
...
+    Version:
+      title: Version
+      description: Version of the ceramic server
+      type: object
+      required:
+        - version
+      properties:
+        version:
+          type: string
+          description: Version of the ceramic server
```
nathanielc, line 102: Are we making any guarantees as to the content of this version string? Is it semver? Something else. Can they be compared in any way?

nathanielc (approved):
LGTM with one question about the API spec.

api/ceramic.yaml
dbcfd: It is semver. Do we want to define a semver type, or just say the string is semver in description?

nathanielc: Semver in the description is good enough for me.

nathanielc (approved):

smrz2001 (approved):
🚀

PR #84, opened 2023-08-10
feat: sqlite blockstore

AaronGoldman:
work in progress

### DIFF
--- a/p2p/src/behaviour.rs
+++ b/p2p/src/behaviour.rs
@@ -24,11 +24,13 @@ use libp2p::{
 use libp2p::{core::identity::Keypair, kad::RecordKey};
 use libp2p_identity::PeerId;
 use recon::{libp2p::Recon, Sha256a};
+use sqlx::SqlitePool;
 use tracing::{info, warn};
 
 pub use self::event::Event;
 use self::peer_manager::PeerManager;
 use crate::config::Libp2pConfig;
+use crate::sqliteblockstore::SQLiteBlockStore;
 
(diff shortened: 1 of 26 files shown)


p2p/src/behaviour.rs
```diff
@@ -112,7 +114,13 @@ where
             } else {
                 BitswapConfig::default_client_mode()
             };
-            Some(Bitswap::new(peer_id, BitswapStore(rpc_client), bs_config).await)
```
nathanielc, line 115: Should we remove the `rpc_client` arg to this function and instead pass in the SQLiteBlockStore ?  This way the path to the sql db doesn't need to be known in multiple places. Just where we make the first connection pool?

p2p/src/config.rs
```diff
@@ -101,6 +102,7 @@ pub struct Libp2pConfig {
     pub notify_handler_buffer_size: usize,
     pub connection_event_buffer_size: usize,
     pub dial_concurrency_factor: u8,
+    pub sql_db_path: PathBuf,
```
nathanielc, line 105: See previous comment, I don't think we want this field here. Instead I would rather we just pass an instance of the blockstore down into the deps instead of passing a duplicate config path and creating a new connection.

nathanielc (commented):
Generally looks good.  I'd like to see the SQLiteBlockStore get created once instead of having multiple places we initialize it. But otherwise looks good.

nathanielc (approved):

PR #85, opened 2023-08-10
feat: add missing kubo endpoints for import and version

dbcfd:


kubo-rpc/src/lib.rs
```diff
@@ -245,6 +253,31 @@ impl IpfsDep for Arc<IpfsService> {
...
     }
+    async fn import<T: std::io::Read + std::io::Seek + Send>(
+        &self,
+        car: T,
+    ) -> Result<Vec<Cid>, Error> {
+        let mut reader =
+            blockless_car::reader::new_v1(car).map_err(|e| Error::Internal(e.into()))?;
+        let roots = reader.header().roots();
+        let mut stored = vec![];
+        for cid in roots.iter() {
```
oed, line 264: I don't believe all roots will be in the `roots`. Normally only one of the CIDs are there. Maybe try iterating through `reader.sections()`?

oed (changes requested):

kubo-rpc/src/lib.rs
AaronGoldman: Do we care about the roots at all?
I thought that the import api took all the blocks and stored them wether they were in the IPLD tree from the root or not.

metadata/src/lib.rs
```diff
@@ -0,0 +1,18 @@
+//! Metadata information and types for ceramic
+
+include!(concat!(env!("OUT_DIR"), "/built.rs"));
```
AaronGoldman, line 3: Why can't we just use a normal `import` here.
```rust
mod built;
pub use built::PKG_VERSION
```

Why don't we check in the build.rs output?
On the one hand it is generated data but the other code reads more easaly if I know what is in the file it is getting values from.

AaronGoldman (commented):
I think it is ok if you have tested the root save loop that it actually stores all the blocks.

kubo-rpc/src/lib.rs
oed: We don't have to care about which CID is root no. We should just import all of the blocks.

kubo-rpc/src/lib.rs
```diff
@@ -4,7 +4,7 @@
 //! <https://docs.ipfs.tech/reference/kubo/rpc/>
 //!
 //! The http server implementation is behind the `http` feature.
-#![deny(missing_docs)]
+//#![deny(missing_docs)]
```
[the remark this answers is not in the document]
dbcfd: Yeah, I cannot figure out what docs are missing on that. Spent a bit before I commented out the docs deny.

kubo-rpc/src/lib.rs
smrz2001: Iteration is now be over the blocks.

For roots, we're aiming for parity with the Go implementation as per the IPLD [CARv1 spec](https://ipld.io/specs/transport/car/carv1/#number-of-roots). We'll error out if there is no root and return the first root if there is more than one and not throw an error in that case.

kubo-rpc/src/http/version.rs
```diff
@@ -0,0 +1,79 @@
...
+{
+    web::scope("/version").route("", web::post().to(version::<T>))
+}
+
+#[tracing::instrument(skip(data))]
+async fn version<T>(data: web::Data<AppState<T>>) -> Result<HttpResponse, Error>
+where
+    T: IpfsDep,
+{
+    #[derive(Serialize)]
```
dbcfd, line 18: Personal preference on this, but I like it outside of the method.

kubo-rpc/src/lib.rs
oed: Great!

PR #93, opened 2023-08-21
feat: use debian image

dbcfd:


Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM public.ecr.aws/r5b3e0r5/3box/rust-builder:latest as builder
+#FROM public.ecr.aws/r5b3e0r5/3box/rust-builder:latest as builder
```
smrz2001, line 1: Would need to uncomment and switch this to debian tag, I suppose.

smrz2001 (approved):
🥳

PR #94, opened 2023-08-23
refactor: reworks kubo-rpc as an openapi server

nathanielc:
With this change we no longer use actix-web for the kubo-rpc server. Instead we use openapi and hyper. Additionally we run the both APIs under the same TCP port.

Also all uses of unimock have been replaced with mockall as it has proven to be much easier to use and assert behavior.


api/ceramic.yaml
```diff
@@ -11,8 +11,11 @@ info:
     name: MIT
     url: https://mit-license.org/
 
+servers:
+  - url: /ceramic
+
 paths:
-  '/ceramic/subscribe/{sort_key}/{sort_value}':
+  '/subscribe/{sort_key}/{sort_value}':
```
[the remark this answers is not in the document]
AaronGoldman: I would default to not having a version unless we need it.
It is just one more thing to deal with and we can use a new name if we want a new behavior

kubo-rpc-server/.gitignore
```diff
@@ -0,0 +1,2 @@
+target
```
AaronGoldman, line 1: why do we need this rather than just the top level

AaronGoldman (approved):
This is long but I think it looks good.
That patch may be fragile to changes in the generated code.
How do you think we would detect it if that broke?

nathanielc:
> How do you think we would detect it if that broke?

Applying the patch is a CI step so if the patch fails to apply CI will fail. Additionally we have locked down the version of the code generator we use so the generated code should not change unexpectedly. So I imagine a workflow where we update the code generator and the patch in the same PR and both need to be working before CI will let it merge.

kubo-rpc-server/.gitignore
nathanielc: We don't however the generated code doesn't know we are in a workspace so it adds this file. Thankfully is a noop.

dbcfd (approved):

smrz2001 (approved):
🚀

PR #95, opened 2023-08-25
refactor: update if let into match for readability

nathanielc:
We did this during Rust office hours as we learned about pattern matching.


smrz2001 (approved):
:shipit:

PR #99, opened 2023-08-30
fix: wrap returned peers in "Peers" key

nathanielc:
The change to openapi for kubo dropped the `Peers` key from the /swarm/peers endpoint. This fixes it.


smrz2001 (approved):
🥳

PR #100, opened 2023-09-01
feat: Add a manual workflow to publish to crates and bump versions

dbcfd:


nathanielc:
I have some thoughts on this some general comments/questions below:

1. We can't publish on crates.io until we have all deps also on crates.io. What should we do with beetle? My vote would be to bring our fork directly into this repo and not bother with publishing our of fork of beetle.
2. There is value in not letting humans pick the next version instead letting automation do so. This workflow is close to doing that but does require the `major|minor|patch` input. The git commit log has `fix/feat/etc`. I would like to see a system that chooses `minor vs patch` based on if the git log contains only fixes. See https://github.com/ceramicnetwork/rust-dag-jose/tree/main/scripts for an example. That example follows  a two step process. Use a workflow to create a release PR with auto picked version and auto generated release notes. If the PR looks good then on merge of the PR a release is built and published to crates.io. 
3. There is benefit to allowing the release to be done locally (i.e. if github is having issues etc). Can we move the logic from the workflows into scripts and create make  targets for them? Then the workflow logic only contains the bits needed to setup the env and call make. 

Thoughts?

ci-scripts/check_api_server.sh
```diff
@@ -8,7 +8,7 @@ cd $DIR/..
 $DIR/gen_api_server.sh > /dev/null
 
 # Check if anything changed, the README contains a generation date, ignore that file.
-changes=$(git status --porcelain api-server | grep -v README.md)
+changes=$(git status --porcelain api-server | grep -v README.md | grep -v Cargo.toml)
```
nathanielc, line 11: why was this changed? Does the Cargo.toml file have something dynamic in it?

nathanielc (approved):
LGTM, with one clarifying question below.

ci-scripts/check_api_server.sh
dbcfd: When I was trying this, release would touch those files, but I think I can adjust release to only touch the yaml files, and not mess with this file.

nathanielc: is that because the release wants to update the version in the Cargo.toml file? If so how does that change get committed into the repo?

dbcfd: Yeah, release wants to update all Cargo.toml files. I adjusted the release script to exclude the generated packages. Release also commits those changes before it does a tag.

PR #104, opened 2023-09-07
fix: remove cycle in debug output

nathanielc:
The bitswap Session type had a cycle (via an Arc<SessionManager>) back to itself causing a stackoverflow when formatted via Debug. This removes the cycle.


smrz2001 (approved):
:shipit:

PR #110, opened 2023-09-12
fix: usage of args should correctly parse and apply now

dbcfd:
Not sure if this is causing the hang, but it should correctly work now.


smrz2001 (approved):
🤞🏼

ci-scripts/package.sh
```diff
@@ -31,34 +27,38 @@ if [[ "$OSTYPE" == "darwin"* ]]; then
   TARGET=$ARCH"-apple-darwin"
 fi
 
-echo "Evaluating program arguments "$@
+echo "Evaluating program arguments '$@'"
```
smrz2001, line 30: Wouldn't the single quotes make `$@` a string literal and not expand the list of args?

dbcfd: Not when it's already in a double quote for echo. Or at least, works on my machine :D

PR #112, opened 2023-09-12
fix: change bitswap dial behavior

nathanielc:
Prior to this change bitswap assumed it was the only behaviour asking the swarm to dial peers, as such when a dial failure occurred it marked the peer as failed and would not send want-haves to the peer.

However this behavior did not work well with the behavior to auto discover ceramic peers and dial them. When ceramic peers were discovered it was possible that they were already connected in which case the `DialFailedPrecondition` error would be reported causing bitswap to mark the peer as failed even though we have a valid connection open to the peer.

This change modifies the bitswap behavior to only modify peer state when a connection closed event is received instead of dial failures. A peer is not marked as good until a connection is established so marking a peer as failed when the connection closes is a valid/consistent behavior.

A few small changes are also made in order to improve debugging.

### DIFF
--- a/beetle/iroh-bitswap/src/lib.rs
+++ b/beetle/iroh-bitswap/src/lib.rs
@@ -7,7 +7,7 @@ use std::fmt::Debug;
 use std::pin::Pin;
 use std::sync::{Arc, Mutex};
 use std::task::{Context, Poll};
-use std::time::{Duration, Instant};
+use std::time::Duration;
 
 use ahash::AHashMap;
 use anyhow::Result;
@@ -48,8 +48,6 @@ pub mod peer_task_queue;
 pub use self::block::{tests::*, Block};
 pub use self::protocol::ProtocolId;
 
-const DIAL_BACK_OFF: Duration = Duration::from_secs(10 * 60);
-
 type DialMap = AHashMap<
     PeerId,
     Vec<(
@@ -82,7 +80,6 @@ enum PeerState {
     Unresponsive,
     #[default]
     Disconnected,
-    DialFailure(Instant),
 }
 
 impl PeerState {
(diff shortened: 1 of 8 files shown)


beetle/iroh-bitswap/src/lib.rs
```diff
@@ -423,23 +432,17 @@ impl<S: Store> NetworkBehaviour for Bitswap<S> {
...
+                        DialError::Denied { cause: _ } => {
+                            // TODO check that the denied cause is because of a connection limit.
+                            //if let Ok(connection_limits::Exceeded { .. }) = cause.downcast() {
+                            self.pause_dialing = true;
+                            //}
                         }
-                    }
+                        _ => {}
+                    };
+                    trace!(%peer_id, error = %event.error,  "report dial failure");
```
AaronGoldman, line 444: Will this give us count of the different `event.error`s happening here?

beetle/iroh-bitswap/src/lib.rs
```diff
@@ -333,9 +330,7 @@ impl<S: Store> Bitswap<S> {
                     *entry.get_mut() = new_state;
                 }
                 match new_state {
-                    PeerState::DialFailure(_)
-                    | PeerState::Disconnected
-                    | PeerState::Unresponsive => {
+                    PeerState::Disconnected | PeerState::Unresponsive => {
```
dbcfd, line 333: Do we expect to ever see DialFailure here?

nathanielc: No the DialFailure variant has been removed. Its no longer a valid reason to change the peer state, so its no longer a peer state.

beetle/iroh-bitswap/src/lib.rs
nathanielc: No, this will just capture a span of the event error. We can add an explicit metric if needed.

beetle/iroh-bitswap/src/lib.rs
```diff
@@ -417,29 +426,23 @@ impl<S: Store> NetworkBehaviour for Bitswap<S> {
             libp2p::swarm::FromSwarm::ConnectionClosed(event) => {
                 self.pause_dialing = false;
                 if event.remaining_established == 0 {
-                    // Last connection, close it
```
smrz2001, line 420: I _think_ "last" actually was correct here since it's checking if `remaining_established=0`.

smrz2001 (approved):
LGTM, thanks for walking through the code!!

:shipit:

PR #113, opened 2023-09-13
chore: use sccache for better Rust caches

nathanielc:
With this change we now cache the registry via `Swatinem/rust-cache@v2` and we cache intermediate compilation steps via `mozilla-actions/sccache-action@v0.0.3`. Additionally we split the single `build` job into three parallel jobs, `check`, `test`, and `build`.  After a few tries I am seeing builds around 20m now.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -10,23 +10,82 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
+  SCCACHE_GHA_ENABLED: "true"
+  RUSTC_WRAPPER: "sccache"
 
 jobs:
-  build:
-
+  check:
     runs-on: ubuntu-latest
-
     steps:
     - uses: actions/checkout@v3
     - uses: dtolnay/rust-toolchain@stable
-    # Cache rust/cargo build steps
     - uses: Swatinem/rust-cache@v2
       with:
         # The prefix cache key, this can be changed to start a new cache manually.
         # default: "v0-rust"
         prefix-key: v0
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -49,7 +47,57 @@ jobs:
...
+        PROTOC_ZIP=protoc-$PROTOC_VERSION-$PROTOC_ARCH.zip
+        curl --retry 3 --retry-max-time 90 -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" -OL https://github.com/protocolbuffers/protobuf/releases/download/v$PROTOC_VERSION/$PROTOC_ZIP
+        sudo unzip -o $PROTOC_ZIP -d /usr/local bin/protoc
+        sudo unzip -o $PROTOC_ZIP -d /usr/local 'include/*'
+        rm -f $PROTOC_ZIP
+        echo "PROTOC=/usr/local/bin/protoc" >> $GITHUB_ENV
+        echo "PROTOC_INCLUDE=/usr/local/include" >> $GITHUB_ENV
     - name: Run tests
       run: make test
+  build:
```
dbcfd, line 76: If we're doing test first, do we need the build step?

nathanielc: They run in parallel, this ensures the binary itself can build independent of if the tests can pass. Technically the binaries are not built as part of the tests since we do not have CLI tests for the binary itself.

dbcfd (approved):

PR #135, opened 2023-10-11
fix: change pubsub/ls to return base64url encoded topics

nathanielc:
Prior to this change the pubsub/ls endpoint would return the raw topic names when the should be base64url encoded names.

Additionally the subscription would remain even after the subscription connection had closed. This is now fixed.

### DIFF
--- a/kubo-rpc/src/lib.rs
+++ b/kubo-rpc/src/lib.rs
@@ -10,7 +10,10 @@ use std::{
     fmt::{self, Display, Formatter},
     io::Cursor,
     path::PathBuf,
-    sync::atomic::{AtomicUsize, Ordering},
+    sync::{
+        atomic::{AtomicUsize, Ordering},
+        Mutex,
+    },
 };
 use std::{str::FromStr, sync::Arc};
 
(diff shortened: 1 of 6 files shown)


kubo-rpc/src/lib.rs
```diff
@@ -288,6 +295,14 @@ impl IpfsDep for Arc<IpfsService> {
         &self,
         topic: String,
     ) -> Result<BoxStream<'static, anyhow::Result<GossipsubEvent>>, Error> {
+        {
+            self.topics
+                .lock()
+                .expect("should be able to lock topics set")
```
smrz2001, line 301: Could this expect cause a panic? If so, is that ok?

smrz2001 (approved):
Apart from one question - LGTM 🚀

kubo-rpc/src/lib.rs
nathanielc: In theory it can cause a panic, but not in practice. From the docs:

> This function might panic when called if the lock is already held by the current thread.

We do not hold the lock and then call any code that would try to acquire the lock again. That is why we explicitly use `expect` in order to make it clear we do not expect the panic to ever occur in practice.

PR #138, opened 2023-10-16
feat: bootstrap peers

AaronGoldman:
bootstrap peers for Mainnet, TestnetClay, and DevUnstable from
https://github.com/ceramicnetwork/
js-ceramic/blob/develop/packages/ipfs-topology/src/ipfs-topology.ts

### DIFF
--- a/one/src/lib.rs
+++ b/one/src/lib.rs
@@ -60,6 +60,7 @@ struct DaemonOpts {
         env = "CERAMIC_ONE_BIND_ADDRESS"
     )]
     bind_address: String,
+
     /// Listen address of the p2p swarm.
     #[arg(
         long,
@@ -69,18 +70,11 @@ struct DaemonOpts {
         env = "CERAMIC_ONE_SWARM_ADDRESSES"
     )]
     swarm_addresses: Vec<String>,
-    /// Address of bootstrap peers.
-    /// There are no default address, use this arg or the API to connect to bootstrap peers as needed.
-    #[arg(
-        long,
-        use_value_delimiter = true,
-        value_delimiter = ',',
-        env = "CERAMIC_ONE_BOOTSTRAP_ADDRESSES"
-    )]
-    bootstrap_addresses: Vec<String>,
+
     /// Path to storage directory
     #[arg(short, long, env = "CERAMIC_ONE_STORE_DIR")]
     store_dir: Option<PathBuf>,
+
     /// Bind address of the metrics endpoint.
     #[arg(
         short,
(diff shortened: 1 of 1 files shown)


one/src/main.rs
```diff
@@ -220,6 +220,40 @@ impl Daemon {
         };
         debug!("using directory: {}", dir.display());
 
+        let mut bootstrap_addresses = opts.bootstrap_addresses.clone();
```
nathanielc, line 223: I believe we decided to remove the `bootstrap_addresses` from the CLI options. Can you make that change?

one/src/main.rs
```diff
@@ -220,6 +220,40 @@ impl Daemon {
...
+        // Add the default bootstrap peers for Mainnet, TestnetClay, and DevUnstable.
+        // from https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/ipfs-topology/src/ipfs-topology.ts
+        let bootstrap_addresses = match opts.network {
```
nathanielc, line 226: We could add this `match` expression as a method on the `Network` type itself. That way its clear the list of bootstrap address is a property of the network selection itself. Thoughts?

one/src/main.rs
```diff
@@ -287,11 +317,14 @@ impl Daemon {
             .build(sql_pool.clone())
             .await?;
 
+        let bind_address = opts.bind_address.clone();
+        let metrics_bind_address = opts.metrics_bind_address.clone();
```
nathanielc, line 321: What is this change for? Why do we clone the address?

nathanielc (changes requested):
LGTM with a few small changes

one/src/main.rs
AaronGoldman: removed

one/src/main.rs
AaronGoldman: Moved to Network enum looks cleaner.

one/src/main.rs
AaronGoldman: Removed

This was a change from a different branch by accident.
 I was changing the way we pass the opts to have all DaemonOpts stored on the Daemon struct.
`daemon.opts` so they could be used in `daemon.run()` not just `daemon.build()`.

one/src/main.rs
```diff
@@ -136,6 +133,34 @@ impl Network {
...
+                "/dns4/go-ipfs-ceramic-private-qa-external.3boxlabs.com/tcp/4011/ws/p2p/QmXcmXfLkkaGbQdj98cgGvHr5gkwJp4r79j9xbJajsoYHr".to_string(),  // cspell:disable-line
+                "/dns4/go-ipfs-ceramic-private-cas-qa-external.3boxlabs.com/tcp/4011/ws/p2p/QmRvJ4HX4N6H26NgtqjoJEUyaDyDRUhGESP1aoyCJE1X1b".to_string(),  // cspell:disable-line
+            ],
+            Network::Local => vec![],
+            Network::InMemory => vec![],
+        }
+        .iter()
+        .map(|addr| addr.parse())
+        .collect::<Result<Vec<Multiaddr>, multiaddr::Error>>()
+        .expect("hard coded bootstrap addresses parse failed")
```
nathanielc, line 162: nit, the idiomatic tense for this is `should`:

```
hard coded bootstrap addresses should parse
```

nathanielc (approved):
LGTM with one small nit

nathanielc (approved):
LGTM

PR #142, opened 2023-10-17
fix: append peer id to all listen addrs

nathanielc:
This change follows Kubo's implementation and is helpful for clients as then can simply iterate through the list of addrs when attempt to peer with them instead of having to construct the full mulitaddr themselves.


smrz2001 (approved):
🚀

PR #143, opened 2023-10-17
fix: add block get timeout

nathanielc:
There is an undocumented `timeout` query parameter to the `block/get` Kubo RPC endpoint that the js-ceramic code relies on. This change adds an implementation of that argument.

TODO:

- [x]  ~Decide if we should emulate the Go error string or allow the Rust specific error string to be reported.~ We are emulating the Go error string.


smrz2001 (approved):
⏰🚀

PR #145, opened 2023-10-19
chore: update libp2p to 0.52.4

nathanielc:
This change updates libp2p to 0.52.4. There are three commits in this PR that break up the change:

1. First remove any unused deps and old beetle iroh-store code.
2. Run cargo update for all deps.
3. Finally update libp2p itself and address any deprecation warnings.


beetle/iroh-bitswap/src/handler.rs
```diff
@@ -181,6 +175,8 @@ impl ConnectionHandler for BitswapHandler {
         self.keep_alive
     }
 
+    // TODO(nathanielc): Remove uses of KeepAlive::Until
```
stbrody, line 178: one thing that my last job did that I liked a lot was that any TODO comment had to reference a ticket number describing the work that would address the TODO.  So something like `// TODO(WS1-1234): remove this`.

Eventually we hooked up tooling so that whenever a ticket was closed, if there were TODOs in the codebase referencing that ticket number, a notification would be sent reminding the team that the TODO could now be addressed.

What would you think about enforcing a similar rule in the rust-ceramic codebase?

stbrody (approved):

beetle/iroh-bitswap/src/handler.rs
nathanielc: I like it, I have created an issue. Will update the PR with the issue link

PR #148, opened 2023-10-24
feat: add offline parameter to block/get

nathanielc:
Adds another undocumented parameter to block/get. Offline indicates the node should not ask the network for the block but instead only check its own store.

### DIFF
--- a/kubo-rpc/kubo-rpc.yaml
+++ b/kubo-rpc/kubo-rpc.yaml
@@ -199,6 +199,12 @@ paths:
           schema:
             type: string
           required: false
+        - name: offline
+          in: query
+          description: When true the peer will not consult other peers for the block, defaults to false.
+          schema:
+            type: bool
+          required: false
       responses:
         '200':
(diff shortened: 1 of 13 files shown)


kubo-rpc/kubo-rpc.yaml
```diff
@@ -199,6 +199,12 @@ paths:
           schema:
             type: string
           required: false
+        - name: offline
```
AaronGoldman, line 202: I would have called this `local` offline feels weird for a http api.

AaronGoldman: Are we cloning a kubo undocumented name?

nathanielc: Same, however we are forced to use the same name as Kubo. Good news is we will eventually deprecate and replace all of these API calls with equivalent and correct abstractions in the Ceramic API.

AaronGoldman (approved):

PR #159, opened 2023-10-30
feat: add explicit log format CLI option

nathanielc:
There are now three log format options.
* Single Line - Uses colors and everything is on a single line
* Multi Line - Uses colors and multiple lines to display more context
* JSON - Does not use colors and formats add structures as JSON on a single line.

This change also makes the default log level INFO.

Fixes WS1-1292

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9135,6 +9135,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "tracing-serde"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bc6b213177105856957181934e4920de57730fc69bf42c37ee5bb664d406d9e1"
+dependencies = [
+ "serde",
+ "tracing-core",
+]
+
+[[package]]
 name = "tracing-subscriber"
 version = "0.3.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


AaronGoldman:
Why not take the langwidge from the logger `compact`, `pretty`, `json`?

AaronGoldman:
Why not take the langwidge from the logger `compact`, `pretty`, `json`

nathanielc:
> Why not take the langwidge from the logger compact, pretty, json?

I felt those names didn't communicate how they differed in structure well enough. Its not clear from `pretty` that it will span multiple lines. And since spanning multiple lines can be very difficult to consume I wanted that to be clear in its name.

However I do not feel strongly and could use either names.

AaronGoldman:
Do we think the help line covers evrything we want to say? 
```
--log-format <LOG_FORMAT>
          Specify the format of log events [env: CERAMIC_ONE_LOG_FORMAT=] [default: multi-line] [possible values: multi-line, single-line, json]
```
I like that it has default and possible values.
I don't like that we don't mention that human forms have color.

nathanielc:
If you do `--help` you get:

```
      --log-format <LOG_FORMAT>
          Specify the format of log events

          [env: CERAMIC_ONE_LOG_FORMAT=]
          [default: multi-line]

          Possible values:
          - multi-line:  Format log events on multiple lines using ANSI colors
          - single-line: Format log events on a single line using ANSI colors
          - json:        Format log events newline delimited JSON objects. No ANSI colors are used
```

If you do `-h` you get:

```
      --log-format <LOG_FORMAT>
          Specify the format of log events [env: CERAMIC_ONE_LOG_FORMAT=] [default: multi-line] [possible values: multi-line, single-line, json]
```

So the info about the colors is in there but only in a more verbose help message.

Help itself says that `--help` will print more information:
```
  -h, --help
          Print help (see more with '--help')
```

AaronGoldman (approved):

PR #160, opened 2023-10-30
feat: default swarm port

AaronGoldman:
CERAMIC_ONE_SWARM_ADDRESSES default_values port 4001

### DIFF
--- a/one/src/lib.rs
+++ b/one/src/lib.rs
@@ -63,7 +63,7 @@ struct DaemonOpts {
     /// Listen address of the p2p swarm.
     #[arg(
         long,
-        default_values_t = vec!["/ip4/0.0.0.0/tcp/0".to_string(), "/ip4/0.0.0.0/udp/0/quic-v1".to_string()],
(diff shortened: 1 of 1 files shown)


one/src/main.rs
```diff
@@ -61,7 +61,7 @@ struct DaemonOpts {
     /// Listen address of the p2p swarm.
     #[arg(
         long,
-        default_values_t = vec!["/ip4/0.0.0.0/tcp/0".to_string(), "/ip4/0.0.0.0/udp/0/quic-v1".to_string()],
+        default_values_t = vec!["/ip4/0.0.0.0/tcp/4001".to_string(), "/ip4/0.0.0.0/udp/4001/quic-v1".to_string()],
```
oed, line 64: Can we run both tcp and quic on a single port?

nathanielc: Yes they are TCP vs UDP and so can use the same port number.

nathanielc (approved):
LGTM, we may need to watch out for any testing code that relied on the fact it would always pick an open port. But all the code I am aware of was explicit so it _should_ be fine.

one/src/main.rs
oed: Thanks, I was suspecting this but wasn't sure!

oed (approved):

PR #161, opened 2023-10-30
fix: remove Ceramic peer discovery

nathanielc:
We are seeing issue related to Kademlia and seem to be connected to Ceramic peer discovery. As this feature is not needed currently, its being removed to eliminate any compounding variables. Once its needed (when we remove gossipsub) it can be reintroduced and fully tested against our established baseline.


AaronGoldman (approved):
There is something very satisfying about removing a feature.

PR #163, opened 2023-10-31
feat: add kubo rpc api metrics

nathanielc:
This change also includes a refactor of the `iroh-metrics` crate into a `ceramic-metrics` crate to make it easier to reuse. However there are some pattern in `iroh-metrics` we do not want to continue. As such I marked those as deprecated so its clear we should use the newer traits etc.

The PR has separate commits for the various refactors to make reviewing easier. The move/rename of the crate happens in a single commit. The edits and new features are separate commits.


stbrody (approved):
seams reasonable as best as I can tell

PR #164, opened 2023-11-01
feat: add cli opts for connection limits

nathanielc:
I expect that we will need to adjust these limits experimentally. So for now I have exposed all limit config directly. If we find that there are values that work for nearly all situations we can remove the options and hard code that value.

### DIFF
--- a/one/src/lib.rs
+++ b/one/src/lib.rs
@@ -96,7 +96,7 @@ struct DaemonOpts {
     #[arg(long, default_value = "testnet-clay", env = "CERAMIC_ONE_NETWORK")]
     network: Network,
 
-    /// Unique key used to find other Ceramic peers via the DHT
+    /// Unique id when the network type is 'local'.
     #[arg(long, env = "CERAMIC_ONE_LOCAL_NETWORK_ID")]
     local_network_id: Option<u32>,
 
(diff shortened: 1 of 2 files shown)


one/src/lib.rs
```diff
@@ -245,28 +274,33 @@ impl Daemon {
...
+            bitswap_server: true,
+            bitswap_client: true,
+            kademlia: true,
+            autonat: true,
+            relay_server: true,
+            relay_client: true,
+            gossipsub: true,
+            max_conns_out: opts.max_conns_out,
+            max_conns_in: opts.max_conns_in,
+            max_conns_pending_out: opts.max_conns_pending_out,
```
stbrody, line 288: looks like all these opts other than `max_conns_in` and `max_conns_out` weren't being set by us before.  Do you know what they defaulted to?  Are we expecting setting this now with the defaults in this PR to change any behavior, or did we match the defaults here to the defaults within libp2p?

stbrody (approved):
lgtm.  One question about whether we expect this PR to cause any behavior change or it just exposes the ability to tweak these later.

one/src/lib.rs
nathanielc: I pulled all the existing defaults up to this layer of the code. So no behavior change is expected. FWIW you can find the previous defaults here https://github.com/3box/rust-ceramic/blob/main/p2p/src/config.rs#L91-L95

PR #165, opened 2023-11-02
feat: add ipfs metrics

nathanielc:
Now we can separately see calls counts and durations to Ipfs operations directly. Instead of only the toplevel Kubo RPC API requests.

### DIFF
--- a/kubo-rpc/src/http/metrics.rs
+++ b/kubo-rpc/src/http/metrics.rs
@@ -20,6 +20,12 @@ struct RequestLabels {
     path: &'static str,
 }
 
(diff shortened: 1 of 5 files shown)


AaronGoldman (approved):
IpfsMetricsMiddleware does not feel very ergonomic. I feel like I could forget to update it during a change but looks good to get the info we want.

nathanielc:
Yeah I want to make this middle metrics pattern generic. But that is going to take some thinking so for now I am just being very verbose about it. But no need to worry to forget about it, compiler will remind you :)

PR #168, opened 2023-11-03
feat: recon-over-http

smrz2001:


recon/src/recon.rs
```diff
@@ -243,13 +265,27 @@ pub trait Store: std::fmt::Debug {
...
+
     /// Return the hash of all keys in the range between left_fencepost and right_fencepost.
     /// Both range bounds are exclusive.
+    /// Returns Result<(Hash, count), Err>
     async fn hash_range(
         &mut self,
         left_fencepost: &Self::Key,
         right_fencepost: &Self::Key,
-    ) -> Result<Self::Hash>;
+    ) -> Result<(Self::Hash, u64)>;
```
nathanielc, line 288: Can we create a struct ala 
```
struct HashCount<H> {
    hash: H,
    count: u64,
}
```

This way its clear that the `u64` is a count.

recon/src/recon.rs
```diff
@@ -87,8 +86,12 @@ where
     }
 
     /// Process an incoming message and respond with a message reply.
+    /// Return Result<(response_message, synced_count, syncing_count), Error>
     #[instrument(skip_all, ret)]
-    pub async fn process_messages(&mut self, received: &[Message<K, H>]) -> Result<Response<K, H>> {
+    pub async fn process_messages(
+        &mut self,
+        received: &[Message<K, H>],
+    ) -> Result<(Response<K, H>, u64, u64)> {
```
nathanielc, line 94: Can we add synced and syncing  on the Response struct itself? Response does not go across the wire so we can add progress metadata like this to it.

nathanielc (approved):
Looks great, just two small changes to make the API a bit clearer. No need to re-review

core/src/event_id.rs
```diff
@@ -37,15 +37,16 @@ impl EventId {
...
-        init: &Cid,
-        event_height: u64,
-        event_cid: &Cid,
+        network: &Network, // e.g. Network::Mainnet
+        sort_key: &str,    // e.g. "model"
+        sort_value: &str,  // e.g. "kh4q0ozorrgaq2mezktnrmdwleo1d" // cspell:disable-line
+        controller: &str,  // e.g. "did:key:z6MkgSV3tAuw7gUWqKCUY7ae6uWNxqYgdwPhUJbJhF9EFXm9"
+        init: &Cid, // e.g. Cid::from_str("bagcqceraplay4erv6l32qrki522uhiz7rf46xccwniw7ypmvs3cvu2b3oulq") // cspell:disable-line
+        event_height: u64, // e.g. 1
+        event_cid: &Cid, // e.g. Cid::from_str("bafyreihu557meceujusxajkaro3epfe6nnzjgbjaxsapgtml7ox5ezb5qy") // cspell:disable-line
```
nathanielc, line 49: This is a nice example. I way to make this show up in the docs is to show an example call in the doc comment.

Something like this:

```rust
///  EventId.new builds a Vec<u8> with the event id data.
/// ## Example
/// ```
/// let event = Event::new(
///     Network::Mainnet,
///     "model",
///     "kh4q0ozorrgaq2mezktnrmdwleo1d",
///     "did:key:z6MkgSV3tAuw7gUWqKCUY7ae6uWNxqYgdwPhUJbJhF9EFXm9",
///     Cid::from_str("bagcqceraplay4erv6l32qrki522uhiz7rf46xccwniw7ypmvs3cvu2b3oulq"),
///     1,
///     Cid::from_str("bafyreihu557meceujusxajkaro3epfe6nnzjgbjaxsapgtml7ox5ezb5qy"),
/// );
///
/// ```
 pub fn new() -> EventId {
 // ...
}
```


You might need to add some `use` statements to make is actually compile and run. But then we can keep the example in sync with the code this way.

recon/src/recon.rs
```diff
@@ -684,20 +729,22 @@ where
         right_fencepost: &K,
         received_hash: &H,
         local_store: &mut S,
-    ) -> Result<bool>
+    ) -> Result<(bool, u64)>
```
nathanielc, line 732: Can you update the doc comment to state what the u64 is for?

recon/src/recon/tests.rs
```diff
@@ -544,7 +545,7 @@ async fn response_is_synchronized() {
     let response = a.process_messages(&response.messages).await.unwrap();
     assert!(!response.is_synchronized);
     let response = x.process_messages(&response.messages).await.unwrap();
-    assert!(!response.is_synchronized);
+    assert!(response.is_synchronized);
```
nathanielc, line 548: Why did this test invert?

nathanielc: Can you update the test comments if this is expected?

nathanielc (commented):
Added some comments, just some feedback on idiomatic Rust nothing significant. Feel free to merge once addressed

recon/src/recon/tests.rs
smrz2001: The `is_synchronized` logic changed so that it can detect synchronization one round earlier. Now, we don't have to wait for a round where nothing is exchanged so detect that we're done. We're synchronized as soon as the ranges match.

PR #172, opened 2023-11-06
feat: cd

smrz2001:
This PR includes automatic deployments to Dev and QA on merges to main. It also includes a new workflow for manually deploying a specified image tag to any of our infrastructure envs.


.github/workflows/publish.yml
```diff
@@ -1,23 +1,52 @@
...
+        password: ${{ env.AWS_SECRET_ACCESS_KEY }}
       env:
         AWS_REGION: us-east-1
-    - name: Build and Publish
-      run: make publish-docker
+    -
+      name: Build and Publish
+      run: |
+        SHA_TAG=$(echo ${{ github.SHA }} | head -c 12)
+        RELEASE_TAG=$(cargo metadata --format-version=1 --no-deps | jq '.packages[0].version' | tr -d '"')
```
nathanielc, line 34: This is not quite right. We need to only set the RELEASE_TAG if this is a release commit. Otherwise we will keep moving the release tag to the latest commit after the release was cut. 

Maybe a simple solution is to add an equivalent `Build and Publish` step to the `publish-release.yml` workflow that already exists? This way we only add a RELEASE_TAG when we actually have a release?

nathanielc (changes requested):
LGTM with one change to how RELEASE_TAG works.

.github/workflows/publish.yml
nathanielc: FWIW  this is how the publish-release.yml knows its a release commit https://github.com/ceramicnetwork/rust-ceramic/blob/main/.github/workflows/publish-release.yml#L17

smrz2001: Ah makes sense, ty!

smrz2001: I like the idea of having a similar step in the release workflow.

smrz2001: Actually, let me see if I can reuse the workflow.

smrz2001: Ok, nvm, just seemed cleaner/easier to add the commit check you indicated to the current workflow.

nathanielc (approved):

PR #175, opened 2023-11-08
feat: continuously publish provider records

nathanielc:
This adds a mechanism to continuously publish provider records. The behavior is that publishes are spread out over the entire publication interval with retries for any failed publishes. The effect is that each block in the store will have its hash published as a provider record at least once per interval.

## Implementation Details

Some things of note for how this was implemented.

The Publisher type implements a Stream trait that continuously emits batches of Keys that need to be published. The node then tells kademlia to publish them and informs the Publisher of the results. When a batch has completed i.e. the publisher knows the pass/fail result of each key, it starts a new batch. 

Between each batch the publisher may sleep/delay if its ahead of its projected deadline (the end of the interval). Structured logs report whether the publisher is ahead or behind.

Metrics have been added to monitor the success/fail rate of publish events.

The kademlia protocol itself has been configured to not do any of its own republishing and to store all provider records in memory. This way  it can answer queries from other peers. This means that the data is duplicated on disk and in memory which is not ideal but seems to be the best we can do for now.

## Performance

On my local system I was able to publish records at a very high rate, 500K in a few hours. This seems to be because we iterate the records in order of hash and so we talk to adjacent sets of peers for large numbers of records. And then we slowly move through the space. The publisher has some settings we can tweak to adjust for production workloads as needed.


p2p/src/node.rs
```diff
@@ -250,6 +264,17 @@ where
                         }
                     }
                 }
+                provide_records = self.publisher.next() => {
+                    if let Some(kad) = self.swarm.behaviour_mut().kad.as_mut() {
+                        if let Some(provide_records) =  provide_records {
```
smrz2001, line 269:
```suggestion
                        if let Some(provide_records) = provide_records {
```
For some reason, my eyes are very good at catching extra/missing whitespaces 😂

AaronGoldman (approved):

one/src/lib.rs
```diff
@@ -78,7 +78,7 @@ struct DaemonOpts {
     #[arg(
         short,
         long,
-        default_value = "127.0.0.1:9090",
+        default_value = "127.0.0.1:9464",
```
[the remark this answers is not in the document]
nathanielc: 9464 is the standard port for hosting metrics, 9090 is the standard port for hosting prometheus itself. I got them mixed up initially. This changes to the more standard port.

p2p/src/node.rs
nathanielc: Hmm I would expect `cargo fmt` to catch this. I'll take a look.

nathanielc: cargo fmt can't always format code inside of a macro and this code block lives inside a `select!` macro invocation. This is my I generally like to put as little logic inside the macro as possible and just call out to functions in the bodies of the branches. 

Fixed

PR #176, opened 2023-11-08
feat: add control over autonat

nathanielc:
Prior to this change autonat would always be enabled. Now it can be disabled in cases where NAT is not being used.

Additionally previously an unbounded number of probes would be running concurrently for all observed addresses. When an dynamic outbound NAT is used this results in many probes. Now we limit to one active probe at a time and also track failed external address so we do not repeatedly attempt to probe them.

### DIFF
--- a/p2p/src/node.rs
+++ b/p2p/src/node.rs
@@ -84,6 +84,8 @@ where
     listen_addrs: Vec<Multiaddr>,
 
     trust_observed_addrs: bool,
+    failed_external_addresses: HashSet<Multiaddr>,
+    active_address_probe: Option<Multiaddr>,
 }
 
 impl<I, M> fmt::Debug for Node<I, M>
@@ -194,6 +196,8 @@ where
             providers: Providers::new(4),
             listen_addrs,
             trust_observed_addrs: libp2p_config.trust_observed_addrs,
+            failed_external_addresses: Default::default(),
+            active_address_probe: Default::default(),
         })
     }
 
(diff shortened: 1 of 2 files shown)


p2p/src/node.rs
```diff
@@ -640,14 +645,16 @@ where
...
-                            // See OutboundProbeEvent case for
-
-                            debug!(
-                                address=%info.observed_addr,
-                                %peer_id,
-                                "probing observed address from peer for external connectivity",
-                            );
-                            autonat.probe_address(info.observed_addr.clone());
+                            // Only probe one address at a time.
+                            if self.active_address_probe.is_none() {
```
stbrody, line 649: if there's already a probe running, then it looks like we drop this addr on the floor.  Should we insert it into a queue to probe later?

nathanielc: I thought about it but I don't think its needed. Everytime we connect to a peer it will report the observed address and trigger this code. So this logic is run very frequently. Additionally its typical to have only a single external address. In which case its likely that the in-progress address probe is the correct one. In cases where there are lots of different observed addresses its likely that NAT hasn't been setup correctly for the peer and so doesn't have an external address.  

Maybe I could add this as a comment on in the code?

p2p/src/node.rs
```diff
@@ -640,14 +645,26 @@ where
...
-                                %peer_id,
-                                "probing observed address from peer for external connectivity",
-                            );
-                            autonat.probe_address(info.observed_addr.clone());
+                            // Only probe one address at a time.
+                            //
+                            // This logic is run very frequently because any new peer connection
+                            // for a new observed address triggers this path. Its typical to have
+                            // only a few external addresses, in which cases its likely that the
+                            // in-progress address probe is one that will succeeded.
```
stbrody, line 653:
```suggestion
                            // in-progress address probe is one that will succeed.
```

stbrody (approved):
lgtm

smrz2001 (approved):
:shipit:

PR #181, opened 2023-11-10
fix: rename iroh to ceramic-one in agent

nathanielc:


AaronGoldman (approved):

p2p/src/behaviour.rs
```diff
@@ -37,7 +37,7 @@ mod event;
 mod peer_manager;
 
 pub const PROTOCOL_VERSION: &str = "ipfs/0.1.0";
-pub const AGENT_VERSION: &str = concat!("iroh/", env!("CARGO_PKG_VERSION"));
+pub const AGENT_VERSION: &str = concat!("ceramic-one/", env!("CARGO_PKG_VERSION"));
```
stbrody, line 40: did we decide to move to `ceramicd` instead of `ceramic-one` as our naming convention?

nathanielc: Tenatively, but until we do so I didn't want to create competing names and ceramic-one is better than iroh.

PR #183, opened 2023-11-14
fix: check limits first before other behaviours

nathanielc:
I have learned that behaviours are called in order and therefore its best practice to place the behaviours that deny connections first in the list. See https://github.com/libp2p/rust-libp2p/pull/4777#discussion_r1391833734


smrz2001 (approved):
🚀

PR #185, opened 2023-11-14
fix: fix JSON log format errors

nathanielc:
Prior to this change JSON fields from spans would produce an error in the logs. Apparently there are two kinds of formats, one for events and one for fields. We need to configure both and they need to align in order for JSON fields to work properly. This change adds a wrapping type to help with this.


smrz2001 (approved):
🚀

PR #186, opened 2023-11-15
fix: block_put_post handle multipart bytes

fdkevin0:
Code in Kubo rpc client use `read_to_string` handle file bytes, cause error like: `Unable to build body: stream did not contain valid UTF-8`. 
https://github.com/ceramicnetwork/rust-ceramic/blob/55dd5d868ef635da819c67601731da1205fce6f9/kubo-rpc-server/src/client/mod.rs#L582
This issue also reappear in official test case: https://github.com/ceramicnetwork/rust-ceramic/blob/55dd5d868ef635da819c67601731da1205fce6f9/kubo-rpc/src/http.rs#L814-L857.
Using `read_to_end` could handle raw bytes and work well with my environment (ceramic-one as Kubo server).

### DIFF
--- a/kubo-rpc-server/README.md
+++ b/kubo-rpc-server/README.md
@@ -15,7 +15,7 @@ To see how to make this your own, look here:
 [README]((https://openapi-generator.tech))
 
 - API version: 0.9.0
-- Build date: 2023-11-13T20:39:18.296854378Z[Etc/UTC]
+- Build date: 2023-11-15T16:13:27.108848Z[Etc/UTC]
 
 
 
--- a/kubo-rpc-server/api/openapi.yaml
+++ b/kubo-rpc-server/api/openapi.yaml
@@ -210,7 +210,7 @@ paths:
         content:
           multipart/form-data:
             schema:
-              $ref: '#/components/schemas/_dag_put_post_request'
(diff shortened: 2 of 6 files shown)


nathanielc:
@fdkevin0 Thanks for the PR! This is indeed a bug that we need to fix. However the code you have changed is part of the generated code for the server client logic. In order to fix this bug we will need to change the api definition  here https://github.com/ceramicnetwork/rust-ceramic/blob/main/kubo-rpc/kubo-rpc.yaml#L228 

You can use the `make gen-kubo-rpc-server` to regenerate the client with the code that can handle bytes vs utf-8 strings.

nathanielc:
> I think it would be better to add a note saying Code generated - DO NOT EDIT at the top of the code file.

I agree, I wish the openapi generator would add that to every file it generates.


Something looks odd here. It seems while the client has changed to treating the file as raw bytes the server code has changed to treating the file as utf-8 bytes. Seems like there might be a bug in the openapi-generator itself.

I'll take a look and see if I can figure out what is up. Thanks again for finding this bug.

>  The current code runs correctly in my project.

Curious what your project is? Mind sharing any details? We currently are actively developing these APIs and they may change. Any input you have on them and your use case would be valuable to the team.

PR #187, opened 2023-11-15
feat: add publisher batch metrics

nathanielc:
These metrics were already in the structured logs. Turns out they are quite useful so I added them as actual metrics.


smrz2001 (approved):
🥳

PR #188, opened 2023-11-15
fix: run publisher in its own task

nathanielc:
This change makes the Publisher logic run in its own task. This is to relieve CPU pressure from the swarm task.

### DIFF
--- a/p2p/src/publisher.rs
+++ b/p2p/src/publisher.rs
@@ -9,10 +9,14 @@ use anyhow::Result;
 use ceramic_metrics::Recorder;
 use futures_timer::Delay;
 use futures_util::{future::BoxFuture, Future, Stream};
-use libp2p::kad::{record::Key, AddProviderError, AddProviderOk};
+use libp2p::kad::{record::Key, AddProviderError, AddProviderOk, AddProviderResult};
 use multihash::Multihash;
-use tokio::sync::oneshot;
-use tracing::{debug, warn};
+use tokio::sync::{
+    mpsc::{channel, error::TrySendError, Receiver, Sender},
+    oneshot,
+};
+use tokio_stream::StreamExt;
+use tracing::{debug, error, warn};
 
 use crate::{
     metrics::{self, Metrics},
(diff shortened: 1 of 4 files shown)


nathanielc:
Locally this works well, however I'd like to test in dev/QA with better load to determine how well it works. We can revert if its not sufficient.

p2p/src/publisher.rs
```diff
@@ -35,10 +39,77 @@ const OPTIMISM: f64 = 0.5;
...
+
+        // We should rarely be behind by more than a single batch.
+        // If we are backpressure is good as we cannot use the new batch.
+        let (batches_tx, batches_rx) = channel(1);
+
+        // Do real work of the publisher on its own task.
+        let mut stream = PublisherStream::new(results_rx, interval, block_store, metrics.clone());
+        let task_metrics = metrics.clone();
+        tokio::spawn(async move {
+            while let Some(batch) = stream.next().await {
```
smrz2001, line 60: So:
- The `Publisher` drives `PublisherStream` in a separate task.
- `PublisherStream` builds and returns batches into a channel that gets polled by the swam task.
- The swarm publishes a batch and returns results via another channel that gets consumed first before a new batch is returned.

Does this sound like an accurate description?

smrz2001: And it's the SQL lookups/batching logic that's been moved out from the swarm task to its own task, via `PublisherStream`.

smrz2001 (approved):
🚀

p2p/src/publisher.rs
nathanielc: >PublisherStream builds and returns batches into a channel that gets polled by the swarm task.

This part is technically incorrect. Both the Publisher and PublisherStream types implement the Stream trait. The swarm is unware that PublisherStream exists or that the channel exists. So the swarm task polls the Publisher for the next batch. The only work the Publisher does is to read a channel if something is available. 

> The swarm publishes a batch and returns results via another channel that gets consumed first before a new batch is returned.

Yes except the swarm is unware of the results channel as well. It simply calls the handle method on the Publisher which puts the result on a channel that the PublisherStream reads. 

So generally correct except that the Publisher abstraction means that the swarm doesn't care that channels and a separate task are being used under the hood.

Perhaps I should rename PublisherStream to PublisherWorker? That way its not misleading that they both implement Stream?

smrz2001: Ah gotcha, makes sense, ty! I got confused by the two layers of streams.

Yeah, I think calling it PublisherWorker would help.

PR #191, opened 2023-11-17
fix: trickle publisher keys to swarm

nathanielc:
Instead of sending all 1000 keys in a batch in one iteration of the swarm loop send one. This way the swarm can continue to process other concurrent work. Tests show this is still fast enough to send out all publish queries within a few seconds.


smrz2001 (approved):
:shipit:

PR #194, opened 2023-11-20
feat: add peering support

smrz2001:
This PR adds support for dialing bootstrap peers with exponential backoff in case of connection or dial failures.

Also, somehow, support for injecting bootstrap addresses got removed from the code, so I added it back.


smrz2001:
Looks like libp2p has deprecated some stuff and Clippy is complaining:
```
error: use of deprecated associated type `libp2p::libp2p_swarm::ConnectionHandler::Error`: Will be removed together with `ConnectionHandlerEvent::Close`. See <https://github.com/libp2p/rust-libp2p/issues/3591> for details.
   --> recon/src/libp2p/handler.rs:205:13
    |
205 |             Self::Error,
    |             ^^^^^^^^^^^
    |
    = note: `-D deprecated` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(deprecated)]`

    Checking libp2p-quic v0.9.3
error: use of deprecated tuple variant `libp2p::libp2p_swarm::ConnectionHandlerEvent::Close`: To close a connection, use `ToSwarm::CloseConnection` or `Swarm::close_connection`. See <https://github.com/libp2p/rust-libp2p/issues/3591> for more details.
   --> beetle/iroh-bitswap/src/handler.rs:334:55
    |
334 |                         yield ConnectionHandlerEvent::Close(err);
    |                                                       ^^^^^
    |
    = note: `-D deprecated` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(deprecated)]`

    Checking ceramic-event v0.9.0 (/home/runner/work/rust-ceramic/rust-ceramic/event)
    Checking iroh-rpc-client v0.9.0 (/home/runner/work/rust-ceramic/rust-ceramic/beetle/iroh-rpc-client)
    ```

File: metrics/src/p2p.rs [hunk 1]
```diff
@@ -65,37 +77,44 @@ impl Metrics {
...
         M: MetricType + std::fmt::Display,
     {
+        // Counters should never record negative values
         if m.name() == P2PMetrics::BadPeer.name() {
-            self.bad_peers.inc_by(value);
+            self.bad_peers.inc_by(value.unsigned_abs());
+        } else if m.name() == P2PMetrics::BadPeerRemoved.name() {
+            self.bad_peers_removed.inc_by(value.unsigned_abs());
+        } else if m.name() == P2PMetrics::BootstrapPeersConnected.name() {
+            // This is a gauge and can record negative values
```
smrz2001, line 90: Was this a reasonable way to handle gauges in the existing framework?

one/src/lib.rs
```diff
@@ -70,6 +70,16 @@ struct DaemonOpts {
     )]
     swarm_addresses: Vec<String>,
 
+    /// Address of bootstrap peers.
+    /// There are no default addresses, use this arg or the API to connect to bootstrap peers as needed.
+    #[arg(
+        long,
+        use_value_delimiter = true,
+        value_delimiter = ',',
+        env = "CERAMIC_ONE_BOOTSTRAP_ADDRESSES"
```
smrz2001, line 79: From Nathaniel: Change to `CERAMIC_ONE_EXTRA_BOOTSTRAP_ADDRESSES` to concatenate additional bootstrap nodes to the official list. Makes it clear and also provides flexibility.

p2p/src/behaviour/peer_manager.rs
```diff
@@ -170,3 +207,133 @@ impl NetworkBehaviour for PeerManager {
...
+    fn new(bootstrap_peers: &[Multiaddr]) -> Self {
+        Self {
+            bootstrap_peers: bootstrap_peers
+                .iter()
+                .filter_map(|multiaddr| {
+                    let mut addr = multiaddr.to_owned();
+                    if let Some(Protocol::P2p(peer_id)) = addr.pop() {
+                        Some((peer_id, BootstrapPeer::new(multiaddr.to_owned())))
+                    } else {
+                        warn!("Could not parse bootstrap addr {}", multiaddr);
```
smrz2001, line 241: @nathanielc: Return a `Result` that gets propagated back to `main`.

p2p/src/behaviour/peer_manager.rs
```diff
@@ -170,3 +207,133 @@ impl NetworkBehaviour for PeerManager {
...
+    fn new(multiaddr: Multiaddr) -> Self {
+        let mut dial_backoff = ExponentialBackoff {
+            initial_interval: BOOTSTRAP_MIN_DIAL_SECS,
+            randomization_factor: BOOTSTRAP_DIAL_JITTER,
+            multiplier: BOOTSTRAP_DIAL_BACKOFF,
+            max_interval: BOOTSTRAP_MAX_DIAL_SECS,
+            max_elapsed_time: None,
+            ..ExponentialBackoff::default()
+        };
+        let dial_future = Some(Box::pin(time::sleep(dial_backoff.next_backoff().unwrap())));
```
smrz2001, line 308: @nathanielc: Don't wait on startup, just return `Ready`.

one/src/lib.rs
```diff
@@ -70,6 +70,15 @@ struct DaemonOpts {
     )]
     swarm_addresses: Vec<String>,
 
+    /// Extra bootstrap peer addresses to be used in addition to the official bootstrap addresses.
```
nathanielc, line 73: Should we add a comment that these address will be a best effort to always remain connected?

one/src/lib.rs
```diff
@@ -371,7 +380,16 @@ impl Daemon {
             max_conns_pending_in: opts.max_conns_pending_in,
             max_conns_per_peer: opts.max_conns_per_peer,
             idle_connection_timeout: Duration::from_millis(opts.idle_conns_timeout_ms),
-            bootstrap_peers: opts.network.bootstrap_addresses(),
+            // Add extra bootstrap addresses to the list of official bootstrap addresses, so that our bootstrap nodes
+            // are always included.
+            bootstrap_peers: [
+                opts.network.bootstrap_addresses(),
+                opts.extra_bootstrap_addresses
```
nathanielc, line 387: I believe there is an `extend` or maybe `append` method on a vector that can take in an iterator. Would simplify this code a bit.

p2p/src/behaviour/peer_manager.rs
```diff
@@ -170,3 +203,143 @@ impl NetworkBehaviour for PeerManager {
...
+    dial_backoff: ExponentialBackoff,
+    dial_future: Option<BoxFuture<'static, ()>>,
+}
+
+impl Debug for BootstrapPeer {
+    fn fmt(&self, f: &mut Formatter) -> Result<(), fmt::Error> {
+        f.debug_struct("BootstrapPeer")
+            .field("multiaddr", &self.multiaddr)
+            .field("dial_backoff", &self.dial_backoff)
+            .finish()
```
nathanielc, line 231: Here you can do something like 

```

            .field("dial_future", self.dial_future.is_some())
```

to indicate if any future is set.

p2p/src/behaviour/peer_manager.rs
```diff
@@ -170,3 +203,143 @@ impl NetworkBehaviour for PeerManager {
...
+                .iter()
+                .map(|multiaddr| {
+                    let mut addr = multiaddr.to_owned();
+                    if let Some(Protocol::P2p(peer_id)) = addr.pop() {
+                        Ok((peer_id, BootstrapPeer::new(multiaddr.to_owned())))
+                    } else {
+                        Err(anyhow!("Could not parse bootstrap addr {}", multiaddr))
+                    }
+                })
+                .collect();
```
nathanielc, line 248: nit, it is common to do `.collect::< Result<AHashMap<PeerId, BootstrapPeer>, anyhow::Error>>()?;` This way the `?` operator is as close to the Result as possible.

p2p/src/behaviour/peer_manager.rs
```diff
@@ -170,3 +203,143 @@ impl NetworkBehaviour for PeerManager {
...
+    }
+}
+
+impl BootstrapPeerManager {
+    fn new(bootstrap_peers: &[Multiaddr], metrics: Metrics) -> Result<Self> {
+        let bootstrap_peers: Result<AHashMap<PeerId, BootstrapPeer>, anyhow::Error> =
+            bootstrap_peers
+                .iter()
+                .map(|multiaddr| {
+                    let mut addr = multiaddr.to_owned();
```
nathanielc, line 241: `multiaddr` has an `.iter()` method that you can use to walk through the protocols without having to copy it.

Something like https://github.com/ceramicnetwork/rust-ceramic/blob/main/kubo-rpc/src/http.rs#L498-L501

p2p/src/behaviour/peer_manager.rs
```diff
@@ -170,3 +203,143 @@ impl NetworkBehaviour for PeerManager {
...
+            bootstrap_peers: bootstrap_peers?,
+            metrics,
+        })
+    }
+
+    fn handle_connection_established(&mut self, peer_id: &PeerId) {
+        if let Some(peer) = self.bootstrap_peers.get_mut(peer_id) {
+            info!(
+                "Connection established, stop dialing bootstrap peer {}",
+                peer.multiaddr
```
nathanielc, line 259: We can use structured logging here. Instead of using format directive i.e. `{}` we can place the fields of the log event before the message:

```
    info!(
                multiaddr= %peer.multiaddr,
                "connection established, stop dialing bootstrap peer",
            );
```

This way the `multiaddr` can be filtered on precisely in the logs. Generally any format directives in a log should be fields instead.

p2p/src/behaviour/peer_manager.rs
```diff
@@ -170,3 +203,143 @@ impl NetworkBehaviour for PeerManager {
...
+                "Connection closed, redial bootstrap peer {}",
+                peer.multiaddr
+            );
+            peer.start_redial();
+            self.metrics.record(&metrics::PeeringEvent::Disconnected);
+        }
+    }
+
+    fn handle_dial_failure(&mut self, peer_id: &PeerId) {
+        if let Some(peer) = self.bootstrap_peers.get_mut(peer_id) {
```
nathanielc, line 278: Do we need a metric for this case?

p2p/src/node.rs
```diff
@@ -196,6 +192,17 @@ where
             metrics,
         );
 
+        let rpc_task = tokio::task::spawn(async move {
```
nathanielc, line 195: Can you add a comment about why moving this later fixing something?

nathanielc (commented):
Looks great, gave it a complete review now. Just a few small recommendations.

p2p/src/behaviour/peer_manager.rs
smrz2001: Yeah, I'll add one.

nathanielc (approved):
:shipit:

one/src/lib.rs
```diff
@@ -70,6 +70,16 @@ struct DaemonOpts {
     )]
     swarm_addresses: Vec<String>,
 
+    /// Extra bootstrap peer addresses to be used in addition to the official bootstrap addresses.
+    /// A best-effort attempt will be made to maintain a connection to these addresses.
+    #[arg(
+        long,
+        use_value_delimiter = true,
+        value_delimiter = ',',
+        env = "CERAMIC_ONE_EXTRA_BOOTSTRAP_ADDRESSES"
```
stbrody, line 79: should we call these "bootstrap" addresses or something like "peering" addresses?  We're starting to have more people ask about running ComposeDB with a high availability setup.  For one user to run multiple nodes in an HA cluster, they'll need to make sure all of those nodes are well connected, so I imagine we'll be starting to tell people to use this option soon, so want to be aligned on the best terminology here.

I feel like the hard coded ones are "bootstrap" addresses because they're the main way you discover the rest of the network, but if you're manually setting other addresses to connect to, that's more like a "peering" address (which I think is also the terminology that kubo uses?)

nathanielc: Rust-ceramic doesn't distinguish between bootstrap vs peering. I don't have a preference for which term we use, however we should only use one.

smrz2001: I'm also cool either way, if you have a preference for "bootstrap" or "peering", @stbrody.

stbrody: Okay, if we feel like we want to keep the same term for both the hard-coded and user-provided peers then I don't feel strongly which one we use.  Fine to keep things as-is

PR #195, opened 2023-11-20
fix: use AIMD for publisher batch size

nathanielc:
AIMD (Additive Increase Multiplicative Decrease), i.e. what TCP windowing uses to dynamically adjust the batch size of the publisher.

The idea is to find a batch size that has a high success rate and therefore increase the throughput of the publisher as its not spending time doing work that is likely to fail.


smrz2001 (approved):
🥳

PR #197, opened 2023-11-21
Feat/event tool

AaronGoldman:
event tool

Event tool is a tool for scanning the block store and finding the Ceramic events it contains.

Scan for Time proof metadata blocks to extract the StreamIDs.
Scan for JOSE signed envelopes for stream init CIDs.

CborValue an enum type to hold and work with arbitrary decoded values from Cbor.


AaronGoldman:
This was done in the validator work

PR #198, opened 2023-11-22
fix: do not use bootstrap list with kademlia

nathanielc:
Prior to the peer_manager connecting to bootstrap peers we relied on
kademlia to connect to them. Now not only is it not necessary but the
peer_manager applies an appropriate backoff for redials where kademlia
does not. There should be no change in external behavior except that we
no longer spam dials to disconnected bootstrap peers.


smrz2001 (approved):
🚀

PR #199, opened 2023-11-22
fix: do not use bootstrap list with kademlia

nathanielc:
Prior to the peer_manager connecting to bootstrap peers we relied on kademlia to connect to them. Now not only is it not necessary but the peer_manager applies an appropriate backoff for redials where kademlia does not. There should be no change in external behavior except that we no longer spam dials to disconnected bootstrap peers.

Now the peer_manager (renamed to ceramic_peer_manager) can report of a peer id is a ceramic peer. If we connect to a ceramic peer we add it to the kademlia routing table regardless of its reported protocols. This is because we discovered that peers sometimes do not report they support the kademlia protocol.

We (@smrz2001  and I ) decided that renaming bootstrap peers to ceramic peers was best because it communicates the role the peers play as opposed to what we are doing with the peers. We may in the future change how we bootstrap with ceramic peers, however we will always need to know about some ceramic peers to bootstrap the network.

### DIFF
--- a/p2p/src/behaviour/ceramic_peer_manager.rs
+++ b/p2p/src/behaviour/ceramic_peer_manager.rs
@@ -63,7 +63,7 @@ impl CeramicPeerManager {
     pub fn new(ceramic_peers: &[Multiaddr], metrics: Metrics) -> Result<Self> {
         let ceramic_peers = ceramic_peers
             .iter()
-            // Extract peer id from mutliaddr
+            // Extract peer id from multiaddr
             .map(|multiaddr| {
                 if let Some(peer) = multiaddr.iter().find_map(|proto| match proto {
                     Protocol::P2p(peer_id) => {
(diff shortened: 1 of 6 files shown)


p2p/src/behaviour/ceramic_peer_manager.rs
```diff
@@ -47,29 +51,40 @@ impl Info {
...
 #[derive(Debug)]
 pub enum PeerManagerEvent {}
 
-impl PeerManager {
-    pub fn new(bootstrap_peers: &[Multiaddr], metrics: Metrics) -> Result<Self> {
+impl CeramicPeerManager {
+    pub fn new(ceramic_peers: &[Multiaddr], metrics: Metrics) -> Result<Self> {
+        let ceramic_peers = ceramic_peers
+            .iter()
+            // Extract peer id from mutliaddr
```
smrz2001, line 66:
```suggestion
            // Extract peer id from multiaddr
```

smrz2001 (approved):
:shipit:

PR #200, opened 2023-11-22
fix: simplify the publisher

nathanielc:
Prior to this change the publisher attempted to estimate and amortize the DHT publish load over the entire interval. Now we use a simpler AIMD method and then publish batches without any delay. Additionally no estimate metrics are computed instead two simple metrics of how many remaining records need to be published and the amount of time before the deadline. Now operationally you can estimate the lag, buts the code itself is simpler.

Additionally logic was added to not start publishing until kademlia has been bootstrapped.

Finally the dynamic batch size has been capped to min and max sizes to ensure we always make at least some progress and we have a upper bound (for the internal channel) on how large a batch will get.

### DIFF
--- a/p2p/src/metrics.rs
+++ b/p2p/src/metrics.rs
@@ -18,7 +18,8 @@ pub struct Metrics {
     publisher_batch_max_retry_count: Gauge,
 
     publisher_batches_finished: Counter,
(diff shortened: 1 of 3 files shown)


smrz2001 (approved):
🚀

PR #204, opened 2023-11-30
feat: add context to EventID

AaronGoldman:


oed (changes requested):
What exactly is context and blockNumber? I'd like to see a spec before we merge this.

AaronGoldman:
canceled we are doing flat recon instead

PR #209, opened 2023-12-22
feat: remove all usage of gossipsub

nathanielc:
Additionally the `eye` command was removed as it only listened to gossipsub messages. In the future we can ad it back based on Recon behavior.

I also discovered there was some left over `gateway` code in the iroh rpc types and client crates. It has been removed too.


AaronGoldman (approved):

smrz2001 (approved):
:shipit: 🚀 

This PR makes me so happy 🥳

PR #211, opened 2024-01-08
feat: Stop publishing CIDs to DHT on write

dav1do:
Linear ticket WS1-1400.

Removed CID publishing to the DHT. We still have the RPCs to [publish](https://github.com/search?q=repo%3Aceramicnetwork%2Frust-ceramic+start_providing&type=code) to the DHT (in p2p/rpc.rs and the corresponding beetle client), but we no longer do it on write, and the node no longer has a publisher.


nathanielc (changes requested):
We should also remove the `publisher` and its related logic. Otherwise looks good.

The `publisher` lives in `p2p/src/publisher.rs`. That whole module can be removed which will result in some changes to `node.rs` as well.

dav1do:
> We should also remove the `publisher` and its related logic. Otherwise looks good.
> 
> The `publisher` lives in `p2p/src/publisher.rs`. That whole module can be removed which will result in some changes to `node.rs` as well.

Thanks, added (or deleted?). I wasn't sure which of these we wanted to keep for future compatibility, and which we wanted to nuke. Makes sense that the RPC stays, everything else goes.

nathanielc (approved):

PR #216, opened 2024-01-16
fix: update deploy workflows for k8s

smrz2001:
This PR allows manual deployments to any of our 4 k8s envs. The CD workflow on pushes to main has been minimally updated so we can test CD and will need to be expanded once we have the additional branch and pre/release workflow updates.

### DIFF
--- a/.github/workflows/deploy-image.yml
+++ b/.github/workflows/deploy-image.yml
@@ -1,30 +1,22 @@
-name: Deploy Release or Commit
+name: Deploy to Durable Infra
 
 on:
   workflow_dispatch:
     inputs:
       environment:
         type: choice
-        description: Infrastructure environment to deploy to
+        description: Durable environment to deploy to
         required: true
-        default: dev
         options:
           - dev
           - qa
           - tnet
           - prod
-      type:
-        type: choice
-        description: Release or Commit
-        required: true
-        default: release
-        options:
-          - release
-          - commit
       tag:
         type: string
         description: Tag to deploy
         required: true
+        default: latest
 
 env:
   AWS_REGION: ${{ secrets.AWS_REGION }}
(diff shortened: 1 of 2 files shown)


samika98:
@smrz2001 Changes look good :)
Maybe impacts the work being done here, if we are planning to keep deploy-image.yml  (https://github.com/ceramicnetwork/rust-ceramic/compare/main...feature/ws1-1414-create-branches-for-rust-ceramic-cd).

I was planning not to use the deploy-image.yml bcs it is confusing and was planning on removing that entirely and creating a separate pre-release workflow.

samika98 (approved):
Changes look good! Might not be able to build on this cause my current approach differs : https://github.com/ceramicnetwork/rust-ceramic/compare/main...feature/ws1-1414-create-branches-for-rust-ceramic-cd. Please provide feedback on the related work being done in the PR :)

smrz2001:
> @smrz2001 Changes look good :)
> Maybe impacts the work being done here, if we are planning to keep deploy-image.yml  (https://github.com/ceramicnetwork/rust-ceramic/compare/main...feature/ws1-1414-create-branches-for-rust-ceramic-cd). Please suggest changes in this PR if needed
> 
> I was planning not to use the deploy-image.yml bcs it is confusing and was planning on removing that entirely and creating a separate pre-release workflow. 

I think there is some value in having a manual workflow as well as an automated one. That way we don't _have_ to make a commit or pre/release to deploy to an environment. That's helpful when we want to debug in an environment without making changes to the formal branches. We can even deploy images from a test branch, if needed.

So we can consider keeping this workflow but including the changes you're making to the structure of the automated workflow. Could also consider adding manual execution to the automated workflow so that there is only one workflow. Will add some complexity to that workflow but not too much, I think.

3benbox (approved):
LGTM

PR #218, opened 2024-01-17
feat: Add sqlite read/write pool split

dav1do:
Linear ticket WS1-1431.

For now, put a basic struct in ceramic_core that manages access to the pools. The sqlx details are not hidden and still relied upon by crates interacting with the database, as the `sqlx::SqlitePool` is part of the public interface.


core/src/sql.rs
```diff
@@ -0,0 +1,47 @@
...
+
+impl SqlitePool {
+    /// Connect to the sqlite database at the given path. Creates the database if it does not exist.
+    /// Uses WAL journal mode.
+    pub async fn connect(path: impl AsRef<Path>) -> anyhow::Result<Self> {
+        let db_path = format!("sqlite:{}", path.as_ref().display());
+        let conn_opts = SqliteConnectOptions::from_str(&db_path)?
+            .journal_mode(SqliteJournalMode::Wal)
+            .create_if_missing(true)
+            // .synchronous(sqlx::sqlite::SqliteSynchronous::Normal) // normally enough in WAL mode?
```
dav1do, line 21: Opinions on this setting? I also assume optimize on close is okay for both the RO and W pools.

core/src/sql.rs
```diff
@@ -0,0 +1,47 @@
...
+            .optimize_on_close(true, None);
+
+        let ro_opts = conn_opts.clone().read_only(true);
+
+        let writer = SqlitePoolOptions::new()
+            .max_connections(1)
+            .connect_with(conn_opts)
+            .await?;
+        let reader = SqlitePoolOptions::new()
+            .max_connections(32) //TODO
```
dav1do, line 31: How configurable do we want the pool to be? we could accept ConnectionOptions (or something similar if we want a lot of control), but it seems like RO connections might be the only value we change for now.

nathanielc: I am fine without much configuration until we discover we need it.

nathanielc (approved):
LGTM

PR #222, opened 2024-01-19
chore: use latest stable openapi-generator-cli

nathanielc:
Turns out that the openapi-generator-cli npm crate is a whole major version behind (6 vs 7) of the openapi-generator jar. This change removes the usage of the npm package as it was only a wrapper for downloading the jar and running it. Now we have a simple bash script that uses that latest stable release of openapi-generator.

This means we now have the fix for the extra .clone() calls and can enable -D warnings for test/example code not just the binary build.


smrz2001 (approved):
:shipit:

PR #223, opened 2024-01-19
fix: rename workflows

smrz2001:
Also includes some syntax tweaks to the CD workflow.


.github/workflows/cd-to-infra.yml
```diff
@@ -1,11 +1,10 @@
-name: Publish and Deploy Image
+name: Continuous Deployment to Infra
 
 on:
   push:
     branches: [ "main" ]
   release:
-    types: [published, edited]
-  workflow_dispatch:
+    types: [published, edited, prereleased, released]
```
samika98, line 7: Can we add 'created' here as well?

samika98:
Thanks Mohsin, for doing this! Added a small comment to include 'created' in the list as well :)

samika98 (approved):
Thanks!

PR #241, opened 2024-01-23
feat: updated release workflow

smrz2001:
Based on @samika98's [PR](https://github.com/ceramicnetwork/rust-ceramic/pull/219) with some minor changes. More comments than actual code changes 😄 

We've retained the release PR process to bump the TOML version. This ensures that a release is not only a manual action but one that _requires approval_ from at least one other team member.

A workflow that could create a release directly would also have been fine but requiring approval seems like a better approach in principle.

The process for creating (pre)releases and their relationship to deployments is as follows:
- Create release PR bumping TOML version (e.g. v0.10.0).
- Merge of PR creates a GH pre-release with tag v0.10.0.
	- This generates the `published` trigger that causes a deployment to `tnet`.
- Repeat as many times as needed to get a stable pre-release.
- Once stable, someone toggles the `latest` flag from the GitHub UI to promote the pre-release to a release.
	- This generates the `released` trigger that causes a deployment to `prod`.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -11,10 +11,10 @@ on:
       - 'Cargo.toml'
 
 jobs:
-  # Build and packages all the things
+  # Build and package all the things
   build-binaries:
     if: |
-      contains(github.event.head_commit.message, 'chore: Release')
+      contains(github.event.head_commit.message, 'chore: version v')
     strategy:
       matrix:
         # For these target platforms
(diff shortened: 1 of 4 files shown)


.github/workflows/publish-release.yml
```diff
@@ -82,8 +82,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_TOKEN }}
-      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      GH_TOKEN: ${{ secrets.GH_TOKEN_PAT }}
```
dbcfd, line 85: Why did this switch to PAT? What did github_token not provide?

smrz2001: Using the regular token won't trigger the next workflow (the `release` triggers, in our case) because GitHub doesn't want workflow recursion. They do allow you to use a PAT for this however. 

From their [docs](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow):
```
If you do want to trigger a workflow from within a workflow run, you can use a GitHub App installation access token or a personal access token instead of GITHUB_TOKEN to trigger events that require a token.
```

smrz2001: We followed the advice from [this](https://github.com/orgs/community/discussions/25281) discussion thread that is very similar to our use case.

3benbox (approved):
LGTM

PR #243, opened 2024-01-24
feat: Modify recon storage tables, trait and sqlite config to improve throughput

dav1do:
Linear WS1-1432.

Before making any changes, I was seeing about 400/100 rps keys added/synced. After these changes, we're able to push the incoming over 1k, and a steady workload targeting 300 completed with ~310/240 keys exchanged/sec. 

**Changes**: 

- I moved the recon values into their own table: `recon_value`. This doesn't have any real benefit up front, but as the table grows it will help. Eventually, a single table will pay during reads as scans will load a lot more data. The value is about 10x bigger and will explode row sizes: a key is ~80 bytes and the value is ~1KB. The initial change moved writes up and syncs down: 450/65.
  - In a POC, I left things in a single table and hit around 800/100 rps just changing the events_post insert to use a single statement. I expect we could push things higher than we are with that format, but eventually I would expect it to slow down while our multiple table approach should stay relatively consistent. 
  - All that to say, I think this is the right choice even though it reduces pure throughput potential initially.
- I refactored the recon traits (particularly storage) to handle inserting a key and value simultaneously. This allows us more control at the database layer about how we handle these writes (in a transaction, batched, etc). Re-using a single transaction for all the writes for a request improved throughput substantially (36%) to 600/85.
- I modified some sqlite config settings. Changing to `pragma synchronous = NORMAL` had a huge impact after the other changes. When I tried it originally, it made zero noticeable difference. After the batched write/transaction usage, it sped up another 80% to 1120/150.

**Follow ups**:
- Something seems wrong with the `hash_range` query (or rather, the indexes/table layout), after a certain point it begins it increase in duration linearly (seems like it's potentially a table scan as we add more rows).
- It's possible using `BEGIN IMMEDIATE` for transactions would be faster (not in sqlx currently, so would likely require a wrapper to make sure drop is called/things are rolled back, or updating to a single statement).
- Other settings like `mmap_size` or `temp_store = memory` might be worth changing, and possibly other pragma options. 
- We could also INSERT multiple rows in one statement. We currently insert a single key/value or at most 2 keys and values on all paths, but 2-4 round trips is probably quite a bit slower than one, even if it is a local file.
- I think our next bottlenecks may be in the recon client/server protocol loops. I need to prove that, but it appears we prioritize incoming writes and starve syncs (potentially a result of the `select!(biased)` usage). It's not necessarily a bad thing, but we may want to look at it.

-----
**Steady simulation results**:

I ran a simulation with the following spec and it almost hit 300 rps on both sides. One side was 280314 over 15m for 311 rps. The sync target was 217945 new keys for a rate of 242 rps. It took about 5 minutes for the recv to catch up, but it ended at 380313 keys (one short) 🤔 This seems like it may be a result of the hanging hash_range queries.

```yaml
apiVersion: "keramik.3box.io/v1alpha1"
kind: Simulation
metadata:
  name: event-ids
  # Must be the same namespace as the network to test
  namespace: keramik-small
spec:
  scenario: recon-event-sync
  devMode: true
  throttleRequests: 350
  successRequestTarget: 280
  users: 1
  runTime: 15
```

The SQL queries were quite consistent throughout and much more performant. Before these changes, the best I was seeing was 1.5m on inserts and upwards of 2.5m depending on the query setup. Here things stay about 0.7m on the high end, while most are much faster.

image

Difference between keys on both sides (logarithmic)
image

Simulation sync rate
image

-------
**Stress test results**:

When adding keys much faster to one node, we see the `middle` and `hash_range` queries on the write side taking the longest. This is 5x what we were seeing above, but they do level out pretty quickly. The second spike resulted in a substantial dip in sync rate. Unfortunately, the result was 1008588 keys on one side and only 883593 on the other after 30 minutes and nothing was happening and the `hash_range` duration kept increasing (over 1s queries in the logs). 

image

-------
**Botched metrics**:

Unfortunately, all of the values below are inflated due to counting value updates (existing key) as a new key. It is interesting compared to the corrected results, as it demonstrates how syncing the values for keys we already have is what is consuming most of the remaining distance between the new keys on the writer and our side. One side was 280395 req in 15 minutes for 311 rps and the other was 282165 (counting keys and new_keys || values), which shows the key + value follow up are going faster than the incoming requests, but just getting keys is not.

image

image

### DIFF
--- a/recon/src/recon.rs
+++ b/recon/src/recon.rs
@@ -98,12 +98,31 @@ where
     /// Reports any new keys and what the range indicates about how the local and remote node are
     /// synchronized.
     pub async fn process_range(&mut self, range: Range<K, H>) -> Result<(SyncState<K, H>, Vec<K>)> {
+        let mut should_add = Vec::with_capacity(2);
         let mut new_keys = Vec::with_capacity(2);
-        if !range.first.is_fencepost() && self.insert(&range.first).await? {
-            new_keys.push(range.first.clone());
+
+        if !range.first.is_fencepost() {
+            should_add.push(range.first.clone());
         }
-        if !range.last.is_fencepost() && self.insert(&range.last).await? {
-            new_keys.push(range.last.clone());
+
+        if !range.last.is_fencepost() {
+            should_add.push(range.last.clone());
+        }
+
+        if !should_add.is_empty() {
+            let new = self
+                .insert_many(should_add.iter().map(|key| ReconItem::new_key(key)))
+                .await?;
+            debug_assert_eq!(
+                new.len(),
+                should_add.len(),
+                "new and should_add must be same length"
+            );
+            for (idx, key) in should_add.into_iter().enumerate() {
+                if new[idx] {
+                    new_keys.push(key);
+                }
+            }
         }
 
         let calculated_hash = self.store.hash_range(&range.first, &range.last).await?;
@@ -224,31 +243,40 @@ where
         self.store.value_for_key(&key).await
     }
 
-    /// Associate a value with a recon key
-    pub async fn store_value_for_key(&mut self, key: K, value: Vec<u8>) -> Result<()> {
-        if self.store.store_value_for_key(&key, &value).await? {
-            self.metrics.record(&ValueInsertEvent);
+    /// Insert many keys into the key space. Includes an optional value.
+    /// Returns a boolean (true) indicating if the key was new.
+    pub async fn insert(&mut self, item: ReconItem<'_, K>) -> Result<bool> {
+        let new_val = item.value.is_some();
+        let new = self.store.insert(item).await?;
+
+        if new {
+            self.metrics.record(&KeyInsertEvent { cnt: 1 });
         }
-        Ok(())
+        if new_val {
+            self.metrics.record(&ValueInsertEvent { cnt: 1 });
+        }
+
+        Ok(new)
     }
 
-    /// Insert a new key into the key space.
-    /// Returns true if the key did not previously exist.
-    pub async fn insert(&mut self, key: &K) -> Result<bool> {
-        let new_key = self.store.insert(key).await?;
-        if new_key {
-            self.metrics.record(&KeyInsertEvent);
-        }
-        Ok(new_key)
-    }
-
-    /// Insert many keys into the key space.
-    pub async fn insert_many<'a, IT>(&mut self, keys: IT) -> Result<bool>
+    /// Insert many keys into the key space. Includes an optional value for each key.
+    /// Returns an array with a boolean for each key indicating if the key was new.
+    /// The order is the same as the order of the keys. True means new, false means not new.
+    pub async fn insert_many<'a, IT>(&mut self, items: IT) -> Result<Vec<bool>>
     where
-        IT: Iterator<Item = &'a K> + Send,
+        IT: ExactSizeIterator<Item = ReconItem<'a, K>> + Send + Sync,
     {
-        let new_key = self.store.insert_many(keys).await?;
-        Ok(new_key)
+        let result = self.store.insert_many(items).await?;
+        let key_cnt = result.keys.iter().filter(|k| **k).count();
+
+        self.metrics.record(&KeyInsertEvent {
+            cnt: key_cnt as u64,
+        });
+        self.metrics.record(&ValueInsertEvent {
+            cnt: result.value_count as u64,
+        });
+
+        Ok(result.keys)
     }
 
     /// Reports total number of keys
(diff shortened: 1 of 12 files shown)


recon/src/recon/sqlitestore.rs
```diff
@@ -127,6 +165,50 @@ where
...
+{
+    type Key = K;
+    type Hash = H;
+
+    /// Returns true if the key was new. The value is always updated if included
+    async fn insert(&mut self, item: ReconItem<'_, Self::Key>) -> Result<bool> {
+        let mut tx = self.pool.writer().begin().await?;
+        let (new_key, new_val) = self.insert_item_int(&item, &mut tx).await?;
+        tx.commit().await?;
+        Ok(new_key || new_val)
```
dav1do, line 184: well boo. this might be why things look so weird in the numbers

nathanielc:
> Something seems wrong with the hash_range query (or rather, the indexes/table layout), after a certain point it begins it increase in duration linearly (seems like it's potentially a table scan as we add more rows).

It is a table scan. In the synchronized state two nodes will be doing a hash_range query for an entire model range (i.e. a single interest). If they both get the same hash then they are synchronized and they stop the synchronization process. This means we often do hash_range queries over large ranges which require that we do the sum for that entire range. I would expect this query to grow as datasize grows. (We have some medium term) plans for how to address this but for now hopefully we can kept it fast enough.

> I think our next bottlenecks may be in the recon client/server protocol loops. I need to prove that, but it appears we prioritize incoming writes and starve syncs (potentially a result of the select!(biased) usage). It's not necessarily a bad thing, but we may want to look at it.

FWIW the biased select usage is intentional to avoid potential deadlocks in the protocol. Happy to explore its usage more in depth.

> When adding keys much faster to one node, we see the middle and hash_range queries on the write side taking the longest

I discovered an easy optimization we can make to the middle query last time I was in the code just didn't have to the time to add it. In fact we may not even need middle anymore. The indent is that we use middle to split the key space. It does this by counting the rows. However hash_range also counts the rows now and we always run hash_range right before we we run middle. So we can just reuse that count to split instead of calling middle. I'll create an issue for this so we can track it.

Still reviewing the code... Thanks for the in depth analysis

recon/src/recon.rs
```diff
@@ -224,31 +243,40 @@ where
...
     }
 
-    /// Associate a value with a recon key
-    pub async fn store_value_for_key(&mut self, key: K, value: Vec<u8>) -> Result<()> {
-        if self.store.store_value_for_key(&key, &value).await? {
-            self.metrics.record(&ValueInsertEvent);
-        }
-        Ok(())
-    }
+    /// Insert many keys into the key space. Includes an optional value.
```
nathanielc, line 246: nit, `Insert many` -> `Insert`

nathanielc (approved):
Change looks great

dav1do:
> > Something seems wrong with the hash_range query (or rather, the indexes/table layout), after a certain point it begins it increase in duration linearly (seems like it's potentially a table scan as we add more rows).
> 
> It is a table scan. In the synchronized state two nodes will be doing a hash_range query for an entire model range (i.e. a single interest). If they both get the same hash then they are synchronized and they stop the synchronization process. This means we often do hash_range queries over large ranges which require that we do the sum for that entire range. I would expect this query to grow as datasize grows. (We have some medium term) plans for how to address this but for now hopefully we can kept it fast enough.

Yep, makes sense. I looked more closely and figured out what it's doing, and it clicked seeing the min/max fenceposts being used. Since it should be associative and we don't support deletion, we can probably create a new table that's updated by a trigger that keeps the hash (total, over some ranges, etc). Think this is worth a discussion.

> 
> > I think our next bottlenecks may be in the recon client/server protocol loops. I need to prove that, but it appears we prioritize incoming writes and starve syncs (potentially a result of the select!(biased) usage). It's not necessarily a bad thing, but we may want to look at it.
> 
> FWIW the biased select usage is intentional to avoid potential deadlocks in the protocol. Happy to explore its usage more in depth.

Yeah, that makes sense. I realized that was probably the reason. What I was seeing and want to figure out, is why it can exchange 1000+ keys/sec, but it's staying below the write speed when it's steady. I was assuming it might be the loop and message exchange frequency, but it could also be key discovery and when there are huge differences it shrinks faster.

> 
> > When adding keys much faster to one node, we see the middle and hash_range queries on the write side taking the longest
> 
> I discovered an easy optimization we can make to the middle query last time I was in the code just didn't have to the time to add it. In fact we may not even need middle anymore. The indent is that we use middle to split the key space. It does this by counting the rows. However hash_range also counts the rows now and we always run hash_range right before we we run middle. So we can just reuse that count to split instead of calling middle. I'll create an issue for this so we can track it.

Awesome, we can discuss this with the hash_range optimizations and tackle in a follow up.

> 
> Still reviewing the code... Thanks for the in depth analysis

recon/src/recon.rs
dav1do: fixed

PR #246, opened 2024-01-28
feat: interest registration endpoint

smrz2001:


nathanielc (commented):
LGTM once you add some tests. Should be a simple copy past edit of the existing tests.

smrz2001:
> LGTM once you add some tests. Should be a simple copy past edit of the existing tests.

Thanks! And yes, haha, was planning to add unit tests before marking the PR ready 😁 (after I get the e2e tests passing)

api/src/server.rs
```diff
@@ -254,6 +269,24 @@ where
...
+    #[instrument(skip(self, _context), ret(level = Level::DEBUG), err(level = Level::ERROR))]
+    async fn interests_sort_key_sort_value_post(
+        &self,
+        sort_key: String,
+        sort_value: String,
+        controller: Option<String>,
+        stream_id: Option<String>,
+        _context: &C,
+    ) -> Result<InterestsSortKeySortValuePostResponse, ApiError> {
+        debug!(
```
nathanielc, line 282: nit, this debug is redundant as the params are already part of the tracing context and will get logged at debug when the function returns successfully or logged at error in the case of an error.

nathanielc (approved):
:shipit:

api/src/server.rs
smrz2001: Ah nice, ok, taking it out.

dav1do (approved):
LGTM, one comment but I think it can be handled separately.

api/src/server.rs
```diff
@@ -104,6 +104,70 @@ where
...
+                        .with_controller(&controller)
+                        .with_init(&stream_id.cid),
+                )
+            }
+            (Some(controller), None) => (
+                start_builder.with_controller(&controller).with_min_init(),
+                stop_builder.with_controller(&controller).with_max_init(),
+            ),
+            (None, Some(_)) => {
+                return Err(ApiError(
```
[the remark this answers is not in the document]
smrz2001: Yeah, I like that, it's more aligned with how the status codes should be used.

PR #252, opened 2024-01-31
chore: adding ci for cargo machete

nathanielc:
This adds a Makefile target to use machete to check for any unused dependencies.

Its possible that machete has a false positive. For those cases simply add the dependencies to the ignored list for machete in the Cargo.toml

In order to make machete happy we needed to clean up the generated Cargo.toml files for the api-server and kubo-rpc-server crates. Since these updates to the toml file are now more involved we use a augtool to safely edit these files.

Builders now must install augeas/augtool.


.github/workflows/rust-build-and-test.yml
```diff
@@ -17,9 +17,9 @@ env:
 jobs:
   check:
     runs-on: ubuntu-latest
+    container: 'public.ecr.aws/r5b3e0r5/3box/rust-builder:latest'
```
nathanielc, line 20: This changes our actions to using our own builder image to simplify installing dependencies. I seem to remember we had that initially and had to change away from it. Does anyone remember why? It will be much simpler if we do not have to maintain two separate build images.

smrz2001: I don't remember the reason we changed away either. Could it have been related to GH/cargo caching?

smrz2001 (approved):
:shipit:

.github/workflows/rust-build-and-test.yml
nathanielc: I checked and GH/cargo caching is still working so I guess we will go forward and if it becomes a problem again we can fix it.

PR #253, opened 2024-01-31
[WIP] feat: add checks for wasi incompatible functionality

dbcfd:
Although we could use a fork, would prefer https://github.com/spruceid/ssi/pull/544 merged in.


dbcfd:
> Was the change to ssi the only change needed to get the checks passing? If so lets wait till it merges before we merge this one.
> 
> We are not in a rush to support WASI

SSI was bringing in libraries that don't work on WASI. For rust-ceramic with the sqlx change, we just need to add a check to prevent issues in the future.

PR #254, opened 2024-01-31
fix: refactor recon storage and remove sqlx dependency from recon and core crates

dav1do:
Removed sqlx (and tokio) dependencies from core (helps with WASM). Technically a breaking change, but we're the only consumers right now and it's alpha.

I modified the store implementations to be interest/model specific as they have different requirements. Interests don't have values, so those requests no-op (they error if you specify a value other than []). I also removed the sort_key column from the tables as its purpose was interest/model differentiation, and renamed the tables accordingly and adjusted the queries.

I had to move some tests around. Interests is currently panicking during formatting (I'm confused by it as the default debug format branch doesn't even have an error condition), so those tests assert without using `expect!` but do still verify things.

Partially addresses [WS1-1462](https://linear.app/3boxlabs/issue/WS1-1462/reduce-ceramic-core-dependencies)


store/src/sqlite/model.rs
```diff
@@ -0,0 +1,1185 @@
...
+            ahash_7 INTEGER,
+            value_retrieved BOOL, -- indicates if we have the value
+            PRIMARY KEY(key)
+        )";
+        const CREATE_VALUE_RETRIEVED_INDEX: &str =
+            "CREATE INDEX IF NOT EXISTS idx_key_value_retrieved
+            ON model_key (key, value_retrieved)";
+
+        const CREATE_MODEL_BLOCK_TABLE: &str = "CREATE TABLE IF NOT EXISTS model_block (
+            sort_key TEXT, -- the field in the event header to sort by e.g. model
```
dav1do, line 70: missed removing this column!

store/src/sqlite/model.rs
```diff
@@ -0,0 +1,1184 @@
...
+            "CREATE INDEX IF NOT EXISTS idx_key_value_retrieved
+            ON model_key (key, value_retrieved)";
+
+        const CREATE_MODEL_BLOCK_TABLE: &str = "CREATE TABLE IF NOT EXISTS model_block (
+            cid BLOB, -- the cid of the Block as bytes no 0x00 prefix
+            key BLOB, -- network_id sort_value controller StreamID height event_cid
+            idx INTEGER, -- the index of the block in the CAR file
+            root BOOL, -- when true the block is a root in the CAR file
+            bytes BLOB, -- the Block
+            PRIMARY KEY(cid, key)
```
nathanielc, line 75: Does the order of the columns in the key matter? We want all the cids for key contiguous for a key if possible. Would `PRIMARY KEY(key, cid) ` do that?

store/src/sqlite/model.rs
```diff
@@ -0,0 +1,1184 @@
...
+            LIMIT
+                ?
+            OFFSET
+                ?
+        ) key
+        JOIN
+            model_block
+        ON
+            key.key = model_block.key
+        ;",
```
nathanielc, line 409: I expect this query is relying on the model_block order index. If not we should likely find a better index.

nathanielc (approved):
Looks great, a few questions below but nothing blocking. 

I found the bug in the Interests Debug impl. However since you have a workaround I say you go ahead with this PR and I'll work on the fix for the Debug bug out of band.

store/src/sqlite/model.rs
dav1do: yea, this one was interesting! already told you on discord, but when I changed the PK to `(key, cid)` it changed the order of these results and resulted in a different CAR file. added an explicit order by `ORDER BY model_block.key, model_block.idx` and we're good. 

it should be using indexes, we have two on model_block `(key, cid)` and `(cid)`, and the key should be sufficient here (can do some EXPLAIN but I'll need to build another db with data in it). model_key has `(key, value_retrieved)` and `(key)` so I think we have good index coverage on all the predicates.

store/src/sqlite/model.rs
dav1do: agree that this makes sense. I changed it and it was interesting (as described below). It could definitely change the disk layout and it should keep all the blocks for a key closer together which is good.

PR #255, opened 2024-02-01
feat: Add feed endpoint to propagate event data to js-ceramic

dav1do:
Implements linear WS1-1454.

## Summary

The API `Recon` trait was removed in favor of two traits (`AccessModelStore` and `AccessInterestStore`), as the API has different behavior for these types and it makes it easier to add a new CRUD endpoints without updating anything recon specific or having to do it for interests and models when it only applies to one. The API no longer depends on the recon client, and the new traits are both implemented by the storage system. This allows the API access to data, without going through the recon protocol loop. This is safe as long as the database writes are synchronized, which they are. Both the `recon::Store` and new API traits access data through the same struct, which uses SQL transactions (and only supports a single writer connection).

The main feature is a new endpoint: `GET /feed/events?resumeAt=$token` which returns a json body of `{events: [Event]}` i.e. `{ events: [{id: "", data: ""]}}`. For now, `data` is always empty **for this endpoint** and `GET /events/$eventId` has to be called to retrieve that data (note that endpoint, which is brand new, was modified in this PR to return the same id/data event body, rather than the older eventId/eventData version).

## More details 

As the new traits share methods with the `recon::Store` trait, it required a lot of disambiguation in tests. I also changed the API traits to return info about whether the key was new in order to capture it in metrics. It is not actually needed by the API, but it makes metrics a lot simpler and API is perfectly happy to ignore the additional information.

I moved the storage related metrics out of the recon crate and into the store crate. This required keeping the same recon prefix, as it needs to pass existing tests. If using the BTreeStore, you need to explicitly register it wrapped by the new metrics struct, which is impossible in the recon crate (as store depends on recon), but perfectly feasible in a binary like ceramic-one.

On my local machine, this seems to speed up the benchmarks (seeing like 338/265 instead of 338/215), which I assume is because we can handle recon syncs via the protocol loop, and incoming writes directly from the API endpoint. We will still be blocked blocked waiting on write DB access for new keys and values, but we have multiple event loops reading now. I was worried I might have changed how we count this (or double counted), but looking at the sqlite DB row counts, I don't see evidence of that.

## Follow up

I want to move more things around (for example, the api/server.rs tests and the new traits), but that can wait. This is on the critical path and those are just quality of life internal changes.


api/src/lib.rs
```diff
@@ -1,7 +1,11 @@
+mod dto;
```
nathanielc, line 1: What is the dto acronym?

api/src/recon.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    async fn keys_since_highwater_mark(
+        &self,
+        highwater: HighwaterMark,
+        limit: i64,
+    ) -> anyhow::Result<(String, Vec<Self::Key>)>;
+}
+
+#[derive(Clone, Debug)]
+pub struct HighwaterMark(i64);
```
nathanielc, line 59: Why are there two types? HighwaterMark and ResumeToken? I see we can convert between them but what is the semantic difference?

api/src/recon.rs
```diff
@@ -0,0 +1,170 @@
...
+        recon::Client::value_for_key(&self.recon, key).await
+    }
+
+    async fn keys_since_highwater_mark(
+        &self,
+        highwater: HighwaterMark,
+        limit: usize,
+    ) -> Result<(ResumeToken, Vec<Self::Key>)> {
+        let (hw, res) = self
+            .model
```
nathanielc, line 150: Does the `model` type have the `insert` `range_with_values` and `value_for_key` methods? Could we implement this ReconModel trait using only `M`? Then we can stop using the Recon::Client.

store/src/sqlite/model.rs
```diff
@@ -47,6 +47,11 @@ where
         Ok(store)
     }
 
+    /// Access the underlying sqlite database pool.
+    pub fn pool(&self) -> &SqlitePool {
```
nathanielc, line 51: What is this for? I didn't find the usages.

store/src/sqlite/model.rs
```diff
@@ -253,6 +258,51 @@ where
...
+        limit: i64,
+    ) -> Result<(i64, Vec<EventId>)> {
+        let query = sqlx::query(
+            "WITH entries AS (
+                    SELECT key, MAX(rowid) as max_rowid
+                FROM model_block
+                    WHERE rowid >= ? -- we return rowid+1 so we must match it next search
+                GROUP BY key
+                ORDER BY rowid
+                limit ?
```
nathanielc, line 279: nit, `limit ` -> `LIMIT`

store/src/sqlite/model.rs
```diff
@@ -1181,4 +1231,63 @@ mod test {
...
+        let (hw, res) = store.new_keys_since_value_rowid(1, 2).await.unwrap();
+        assert_eq!(2, res.len());
+        assert_eq!(11, hw); // see comment in prep_highwater_tests
+        assert_eq!([key_a, key_b], res.as_slice());
+
+        let (hw, res) = store.new_keys_since_value_rowid(hw, 1).await.unwrap();
+        assert_eq!(1, res.len());
+        assert_eq!(22, hw); // see comment in prep_highwater_tests
+        assert_eq!([key_c], res.as_slice());
+    }
```
nathanielc, line 1292: These tests are great, made it easy to follow the logic. Thanks

nathanielc (changes requested):
Overall looks great. I think we are close to dropping the Recon::Client dependency from the ceramic_api Server type. I think we should do that as part of this change. 

A few other small comments below

api-server/api/openapi.yaml
```diff
@@ -135,6 +135,40 @@ paths:
         "204":
           description: success
       summary: Register interest for a sort key
+  /events/feed:
```
stbrody, line 138: so, the firehose endpoint in js-ceramic is at `/api/v0/feed/aggregation/documents`.  If we want to use a consistent convention then this one should probably be `/feed/events` instead of `events/feed`?  I'm assuming the `/api/v0` is already handled somehow elsewhere?

stbrody (commented):
my main concerns are thinking about how best to future-proof this.

Like right now this feed just returns eventIds without the payload.  When we start adding the full event payload CAR file, do you think it will be okay to just add that as a new field to the output in the same endpoint, or would we want a new endpoint?

Also, do we have any idea how hard it will be to convert this feed to use SSE?  Does openAPI support SSE?  When we make that transition, will we be able to re-use the same endpoint or will we want a new URL?  If we'll need a new url, should we change the naming convention of the endpoint we're using now to prepare for that in some way?

store/src/sqlite/model.rs
dav1do: Good catch, will delete. I added it thinking I was going to implement the new Access traits in api, and would need access to the pool.. and realized it was a really bad idea to expose sqlx stuff there and forgot to delete it.

api/src/recon.rs
dav1do: We discussed this Friday and it seems like it, so I'll be working on this today

api/src/recon.rs
dav1do: Good question.. I think highwater mark can be deleted. I originally wanted to make the resume token/cursor something that might apply to multiple routes and be the public serializable type, and then use the highwater mark internally to wrap the storage system's type.. but I didn't implement it well and having both types in the one interface is a mistake.

api/src/lib.rs
dav1do: It's data transfer object.. a coworker at my last job used it a lot (c# background). I was going to call it `types` even though it's a bad name... I basically wanted to signal that these are public types that are meant to be sent over the wire, but it didn't even really turn out like that since ResumeToken isn't in a public API interface (just built from a string body). 

So anyway, I'm going to move some things around today and this will likely end up in its own file or in a new module called types (or something better?).

api-server/api/openapi.yaml
stbrody: > I stuck to the "spec" in the doc, but I originally added a new type to wrap the body e.g. something like {eventids: [string] } where we could add new fields

What if instead, the objects that came back in responses from the feed were like:
```
{
  events: [ {id: "<eventId1>" }, { id: "<eventId2>"}, ... ]
  resumeToken: "XXXXX",
}```

That way if in the future we add an arg to the endpoint to instruct it to start including event payload data too, we could add a `data` field alongside the `id` field in the inner `event` objects?

File: store/src/lib.rs [hunk 8]
```diff
@@ -3,7 +3,9 @@
 #![warn(missing_docs)]
 
 mod sqlite;
+mod store_metrics;
```
nathanielc, line 6: Should we just call this `metrics` since we are in the `store` crate already?

store/src/sqlite/interest.rs
```diff
@@ -119,6 +116,65 @@ where
...
+        //debug!(count = rows.len(), "rows");
+        Ok(Box::new(rows.into_iter().map(|row| {
+            let bytes: Vec<u8> = row.get(0);
+            Interest::from(bytes)
+        })))
+    }
+}
+
+#[async_trait::async_trait]
+impl<H> AccessInterestStore for InterestStore<H>
```
nathanielc, line 159: I think it would help to use a qualified path for this trait. Something like `ceramic_api::AccessInterestStore` and then add a comment explaining why we expose the store to the API. This way its clear this is an explicit design decision for how the code all interacts.

store/src/sqlite/model.rs
```diff
@@ -687,6 +751,41 @@ impl iroh_bitswap::Store for ModelStore<Sha256a> {
     }
 }
 
+#[async_trait::async_trait]
+impl AccessModelStore for ModelStore<Sha256a> {
```
nathanielc, line 755: Similar comment here, Could we be explicit about the reason for implementing api crate traits here.

nathanielc (approved):
Looks Great! Two small requests to add some comments for clarity in our design patterns but no need for rereview.

api-server/api/openapi.yaml
```diff
@@ -178,4 +212,27 @@ components:
       - eventId
       title: A Ceramic Event
       type: object
+    EventFeed:
+      description: Ceramic event keys as part of a Ceramic Stream
+      example:
+        resumeToken: resumeToken
+        events:
+        - eventId: eventId
```
stbrody, line 220: nitpick: should we just call the fields in individual Events `id` and `data`, instead of `eventId` and `eventData`?  Save some bytes in the responses?

dav1do: hmm. yeah, probably. I didn't do that thinking it'd breaks all the existing routes, however it looks like we use this as the POST event payload and the response for the subscribe endpoint, which is now deprecated. I made a new type called `Event` that is just `id, data` , and renamed the existing one `EventDeprecated`.

dav1do: implemented in [f33455b](https://github.com/ceramicnetwork/rust-ceramic/pull/255/commits/f33455b25cedaf8ab191eb706e195817a441b88c)

dav1do: also updated `events/$eventId` to use this new format after merging main (#258)

PR #260, opened 2024-02-05
chore: fix false positive in checking generated servers

nathanielc:
There was a bug were failing to get the `git status` would exit the script but not fail the check. This is now fixed as well as the git permission issue that caused the original failure


ci-scripts/check_generated_server.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+# Script to check a generated server has not changed
+
```
smrz2001, line 3: Should we add a `set -e` as well?

smrz2001 (approved):
LGTM apart from one question/suggestion.

ci-scripts/check_generated_server.sh
AaronGoldman: https://www.shellcheck.net/

```bash
$ shellcheck myscript
 
Line 7:
DIR=$(dirname $0)
              ^-- [SC2086](https://www.shellcheck.net/wiki/SC2086) (info): Double quote to prevent globbing and word splitting.

Did you mean:
DIR=$(dirname "$0")
 
Line 8:
cd $DIR/..
^-- [SC2164](https://www.shellcheck.net/wiki/SC2164) (warning): Use 'cd ... || exit' or 'cd ... || return' in case cd fails.
   ^-- [SC2086](https://www.shellcheck.net/wiki/SC2086) (info): Double quote to prevent globbing and word splitting.

Did you mean:
cd "$DIR"/.. || exit
 
Line 14:
status=$(git status --porcelain $server_dir)
                                ^-- [SC2086](https://www.shellcheck.net/wiki/SC2086) (info): Double quote to prevent globbing and word splitting.

Did you mean:
status=$(git status --porcelain "$server_dir")
 
Line 15:
if [ $? -ne 0 ]
     ^-- [SC2181](https://www.shellcheck.net/wiki/SC2181) (style): Check exit code directly with e.g. 'if ! mycmd;', not indirectly with $?.
```

🤷‍♂️

ci-scripts/gen_api_server.sh
```diff
@@ -3,6 +3,8 @@
 # Script to generate api-server crate from OpenAPI definition.
 # Requires augeas/augtool to be installed
 
+set -e
+
 DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 cd $DIR/..
```
AaronGoldman, line 9: https://www.shellcheck.net/
```bash
$ shellcheck myscript
 
Line 9:
cd $DIR/..
   ^-- [SC2086](https://www.shellcheck.net/wiki/SC2086) (info): Double quote to prevent globbing and word splitting.

Did you mean:
cd "$DIR"/..
```

AaronGoldman (approved):

ci-scripts/check_generated_server.sh
nathanielc: Unfortunately no, that would have been simpler. However the `grep -v README.md` call can fail if the README didn't change (which is expected) and then we get a false negative. So we have to handle failures individually.

PR #263, opened 2024-02-07
fix: use try_from on recon keys

nathanielc:
This changes the recon::Key trait to use TryFrom<Vec<u8>> instead of simply From<_> in order to allow for validation of the bytes. Now it _should_ be impossible to construct either an Interest or an EventId that is not valid. Remembering that a fencepost is considered valid.

### DIFF
--- a/store/src/sqlite/model.rs
+++ b/store/src/sqlite/model.rs
@@ -33,6 +33,8 @@ struct BlockRow {
     bytes: Vec<u8>,
 }
 
+type EventIdError = <EventId as TryFrom<Vec<u8>>>::Error;
+
 impl<H> ModelStore<H>
 where
     H: AssociativeHash + std::convert::From<[u32; 8]>,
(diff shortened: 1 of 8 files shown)


store/src/sqlite/model.rs
```diff
@@ -628,10 +631,13 @@ where
...
             .await?;
-        Ok(rows.first().map(|row| {
-            let bytes: Vec<u8> = row.get(0);
-            EventId::from(bytes)
-        }))
+        Ok(rows
+            .first()
+            .map(|row| {
+                let bytes: Vec<u8> = row.get(0);
+                EventId::try_from(bytes)
```
AaronGoldman, line 638: Why not `EventId::try_from(bytes)?` instead of  `.transpose()?` after it is wrapped by the map?

store/src/sqlite/model.rs
```diff
@@ -709,8 +715,8 @@ where
             .await?;
         Ok(row
             .into_iter()
-            .map(|row| EventId::from(row.get::<Vec<u8>, _>(0)))
-            .collect())
+            .map(|row| EventId::try_from(row.get::<Vec<u8>, _>(0)))
```
AaronGoldman, line 718: Would `.map(|row| EventId::try_from(row.get::<Vec<u8>, _>(0))?)` prevent us from needing to change the next line?

store/src/sqlite/model.rs
nathanielc: Because we are inside a closure of the Option.map call. So `?` does not early return from the outer function but only the closure, which means the closure needs to return Result<EventId> either way which gives us Option<Result<EventId>> and we need to transpose.

store/src/sqlite/model.rs
nathanielc: No, for the same reason as above, the map closure has to return a Result<_> in either case and we have to propogate that Result to the outer function.

PR #264, opened 2024-02-07
fix: allow double insert of key/value pairs in Recon

nathanielc:
Previous behavior was to error if an event and its value were inserted twice. Now we simply drop a value if we already have a value. This assumes that keys always point to the same value which is the base assumption of recon itself.


AaronGoldman (approved):

store/src/sqlite/model.rs
```diff
@@ -1000,13 +1027,8 @@ mod test {
 
         // reject the second insert of same key with value. Do not override values
```
dav1do, line 1028: nit: this comment is wrong now

dav1do (approved):
LGTM. Is there any case we might get a different value? Just wondering if we'll ever want to log/warn/etc if somehow a value is trying to be changed.

PR #266, opened 2024-02-08
fix: store metrics under own registry (not recon)

dav1do:
The store/query metrics are now under their own subregistry (ceramic_store) rather than using recon. This makes sense based on the code layout, but was put off as it requires updating the performance tests as they look at this metric.


AaronGoldman (approved):
makes sense to me

nathanielc (approved):

PR #268, opened 2024-02-09
fix: moved api tests to a separate file

samika98:
Follow-up on the previous features. Noticed that it would be better to shorten the server.rs and move out test cases to a separate file.

This is just a followup, this PR does not change any business logic or adds any new logic.


api/src/lib.rs
```diff
@@ -7,3 +7,6 @@ pub use metrics::Metrics;
 pub use resume_token::ResumeToken;
 
 pub use server::{AccessInterestStore, AccessModelStore, Server};
+
+#[cfg(test)]
+mod test;
```
nathanielc, line 12: nit, its common to call this module `tests` as it contains many tests.

api/src/test.rs
```diff
@@ -0,0 +1,637 @@
...
+};
+use ceramic_core::{Cid, Interest};
+use ceramic_core::{EventId, Network, PeerId, StreamId};
+use mockall::{mock, predicate};
+use multibase::Base;
+use recon::{Key, Sha256a};
+use std::str::FromStr;
+use tracing_test::traced_test;
+
+mod tests {
```
nathanielc, line 22: you can remove this mod scoping now since the entire file is the tests module. We do not need another nested test module.

api/src/server.rs
```diff
@@ -340,641 +339,15 @@ where
     }
 }
 
-fn decode_event_id(value: &str) -> Result<EventId, ApiError> {
+pub(crate) fn decode_event_id(value: &str) -> Result<EventId, ApiError> {
```
nathanielc, line 342: why did these need to become crate public?

api/src/server.rs
```diff
@@ -340,641 +339,15 @@ where
...
     multibase::decode(value)
         .map_err(|err| ApiError(format!("multibase error: {err}")))?
         .1
         .try_into()
         .map_err(|err| ApiError(format!("invalid event id: {err}")))
 }
-fn decode_event_data(value: &str) -> Result<Vec<u8>, ApiError> {
+pub(crate) fn decode_event_data(value: &str) -> Result<Vec<u8>, ApiError> {
```
nathanielc, line 349: and this one?

nathanielc (changes requested):
A few small changes but overall looks good.

api/src/server.rs
dav1do: they are used by the tests which can't access them from the new file. I haven't really looked at whether that means there could be something semi-public that wraps this better (i.e. an input validator). same answer for `decode_event_data`

api/src/server.rs
samika98: This is being used in the test file as well as the BuildResponse

api/src/lib.rs
samika98: Appreciate the nits

dav1do (approved):

nathanielc (approved):

PR #269, opened 2024-02-09
feat: support new id/data event payload for event creation (POST /events)

dav1do:
I renamed the fields on the Event schema to save bytes but only supported that in responses to avoid breaking changes, but there was a path forward to support both for inputs without breaking anything. We do this for now, and can update callers to `id/data` instead of `event_id/event_data` and simplify the input validation at that time. 

We prefer `id/data` over `event_id/event_data` and don't allow mixing. I was going to allow either `id` field and either `data` field, but this ensures people are using either style correctly and completely, rather than a hybrid.

The implementation is ugly, as using `oneOf` schema generation doesn't work quite right. It generated 4 required fields on one struct, rather than an enum with 2 variants with 2 required fields each. So instead I crate a new type that is dedicated to parsing this input 😞.

Note: I skipped the subscribe endpoint as it's deprecated and interests should be used instead.

### DIFF
--- a/api-server/.openapi-generator/FILES
+++ b/api-server/.openapi-generator/FILES
@@ -6,6 +6,7 @@ api/openapi.yaml
 docs/Event.md
 docs/EventDeprecated.md
 docs/EventFeed.md
+docs/EventsPostRequest.md
 docs/Version.md
 docs/default_api.md
 examples/ca.pem
--- a/api-server/README.md
+++ b/api-server/README.md
@@ -15,7 +15,7 @@ To see how to make this your own, look here:
 [README]((https://openapi-generator.tech))
 
 - API version: 0.10.1
-- Build date: 2024-02-09T12:36:59.242763-07:00[America/Denver]
+- Build date: 2024-02-09T13:10:42.872308-07:00[America/Denver]
 
 
 
--- a/api-server/README.md
+++ b/api-server/README.md
@@ -15,7 +15,7 @@ To see how to make this your own, look here:
 [README]((https://openapi-generator.tech))
 
 - API version: 0.10.1
-- Build date: 2024-02-08T16:59:26.485961797Z[Etc/UTC]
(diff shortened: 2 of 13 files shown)


stbrody:
Do we need to do the work to support backwards compatibility?  Especially if it complicates the code a bunch?  I'm pretty sure the number of callers of this at this point is... zero?  It will be one once @stephhuynh18's PR merges.

stbrody:
oh I guess the existing perf and correctness tests are other users of this. So making the breaking change would break the tests until we could fix them.  But given what a small and trivial change to the tests it will be, maybe we can just prepare that update and merge both this and the change to the tests at basically the same time and be fine?

stbrody:
normally I'm very against breaking things and very for backwards compatibility. But given we're still pre-release, spending a bunch of time on backwards compat just feels like slowing ourselves down and making our lives complicated for not much benefit.  Would we be better off just ripping the bandaid off quickly?

nathanielc:
> normally I'm very against breaking things and very for backwards compatibility. But given we're still pre-release, spending a bunch of time on backwards compat just feels like slowing ourselves down and making our lives complicated for not much benefit. Would we be better off just ripping the bandaid off quickly?

In this case it won't make it through CI since CI uses the old client so we can't just break it. So the simplest way to get this out is to support both and then update CI and then remove the old method.

nathanielc (approved):

dav1do:
> > normally I'm very against breaking things and very for backwards compatibility. But given we're still pre-release, spending a bunch of time on backwards compat just feels like slowing ourselves down and making our lives complicated for not much benefit. Would we be better off just ripping the bandaid off quickly?
> 
> In this case it won't make it through CI since CI uses the old client so we can't just break it. So the simplest way to get this out is to support both and then update CI and then remove the old method.

I agree with all of this. I'll put up PRs for the e2e and keramik tests, and a follow up here to delete this awkardness. I really like having the green signal, and am okay doing the extra work to keep that in place while I clean up.

PR #270, opened 2024-02-09
feat: enable recon by default

nathanielc:
Now that we have broken compatibility with Kubo APIs the only purpose of ceramic-one is to run Recon. As such we should enable it by default as the binary doesn't do anything meaningful with it disabled.


smrz2001 (approved):
🚀🚀

PR #271, opened 2024-02-09
fix: honor PENDING_RANGES_LIMIT for initial ranges

nathanielc:
Prior to this change all initial interest ranges requests were sent to the remote potentially deadlocking the protocol. Now anytime the initiator sends a RangeRequest to the responder it always checks the pending ranges limit.


smrz2001 (approved):
:shipit:

PR #282, opened 2024-03-06
feat: inital mdbook docs for rust-ceramic

3benbox:
Checkout the branch, change to the ./rust-ceramic directory and run `mdbook serve --open` to view the contents.

Took the same directory name as kermaik, happy to move it if it's confusing.

Closes WS1-1515


rust-ceramic/src/SUMMARY.md
```diff
@@ -0,0 +1,15 @@
...
+- [Introduction](./introduction.md)
+- [Installation](./install.md)
+  - [Releases](./releases.md)
+  - [Source](./source.md)
+  - [Docker](./docker.md)
+- [Usage](./usage.md)
+- [Advanced](./advanced.md)
+  - [Configuration](./configuration.md)
+  - [Metrics](./metrics.md)
+- [help](./help.md)
```
nathanielc, line 12: nit , `help` -> `Help`

rust-ceramic/src/license.md
```diff
@@ -0,0 +1 @@
+# License
```
nathanielc, line 1: Is there supposed to be more content here?

rust-ceramic/src/source.md
```diff
@@ -0,0 +1,58 @@
...
+```bash
+make build
+```
+
+The generated binary will be located at `target/release/ceramic`.
+
+Verify it's functioning by running:
+
+```bash
+./target/release/ceramic --help
```
nathanielc, line 57: should this be `ceramic-one`?

rust-ceramic/src/docker.md
```diff
@@ -0,0 +1,8 @@
+# Docker
+
+Docker images for rust-ceramic are available from our AWS container registry.
+
+Currently, the latest release is `0.12.0`.
```
nathanielc, line 5: I'd prefer to to have to update this every time we cut a release. Can we either automate updating or word it so that we can just point at `latest` tag instead of the name of the release? 

Note the release script does have the ability to update this automatically we would just need to tell it about this file.

rust-ceramic/src/usage.md
```diff
@@ -0,0 +1,25 @@
+# Usage
```
nathanielc, line 1: Should we call out `ceramic-one --help` here? We have good usage docs built into the binary, I'd like to refer to them here.

rust-ceramic/src/configuration.md
```diff
@@ -0,0 +1,10 @@
+# Configuration
+
+Configuration is done via environment variables. The following environment variables are available:
+
+- `CERAMIC_ONE_NETWORK`: Ceramic network (default: `testnet-clay`)
+- `CERAMIC_ONE_STORE_DIR`: Path to storage directory (default: `~/.ceramic-one`)
+- `CERAMIC_ONE_BIND_ADDRESS`: Bind address of the metrics endpoint (default: `5001`)
+- `CERAMIC_ONE_METRICS_BIND_ADDRESS`: Bind address of the metrics endpoint (default: `9464`)
+- `CERAMIC_ONE_SWARM_ADDRESSES`: Listen address of the p2p swarm (default: `/ip4/0.0.0.0/tcp/4001`)
+- `CERAMIC_ONE_LOCAL_NETWORK_ID`
```
nathanielc, line 10: There are many more configuration env vars. All cli flags can be configured via the env. Should we highlight some important ones here and then refer to `ceramic-one daemon --help` for the rest?

Or if we could automate dumping the help test output here that would also work.

rust-ceramic/src/metrics.md
```diff
@@ -0,0 +1,10 @@
+# Metrics
+
+Available metrics are:
+- `ceramic_api_requests_total`: Total number of Ceramic API requests
+- `ceramic_one_info`: Information about the Ceramic node
+- `ceramic_store_key_insert_count_total`: Total number of keys inserted into the store
+- `ceramic_store_value_insert_count_total`: Total number of values inserted into the store
+- `ceramic_store_store_query_durations_[sum, buckets, count]`: Store query durations
+- `ceramic_api_request_durations_[sum, buckets, count]`: API request durations
+- `ceramic_api_api_info`: Information about the Ceramic API
```
nathanielc, line 10: Should we note that there are many more metrics and encourage operators to explore them via the `/metrics` endpoint? We can highlight these as the important ones but this is not an exhaustive list.

rust-ceramic/src/help.md
```diff
@@ -0,0 +1,167 @@
+```
```
nathanielc, line 1: I see we do have the help usage in the docs, great. Is this automated? Can we link to if from the places in the book that could benefit from it?

nathanielc (changes requested):
Looking good a few comments:

* I think we should name the directory simply `book` I find the redundant name confusing
* I want to make sure these docs are not immediately out of date as soon as we create them. To me that means either a) automate generating some content from the source (i.e. the help text) or b) Do not include any content that is already part of the binary but instead simply refer to it so it can't get out of sync. I would prefer option a but its more work so we can build up to that.
* How does this work in CI?
    *   Does this get deployed as a Github Pages book?
    *  Can we add a Makefile target to build the book?
    * Is the `book` directory the generated content? If so can we add a Makefile target to check its been generated correctly like we do for the api server crates?

rust-ceramic/src/SUMMARY.md
3benbox: Fixed

rust-ceramic/src/license.md
3benbox: Should be, but removed for now.

rust-ceramic/src/metrics.md
3benbox: Yeah, like the "help" I'd love a way to generate the metrics for the docs.

I don't see a way outside parsing all the source files.

Do you have any ideas how to make a rust command, or entrypoint, that would print the available metrics?

rust-ceramic/src/help.md
3benbox: Updated to generate a files that is included in the book.
CD would run the help commands and generate the help files that get included.
- `ceramic-one events slurp --help > src/ceramic-one-events-help.txt`
- `ceramic-one daemon --help > src/ceramic-one-daemon-help.txt`

rust-ceramic/src/configuration.md
3benbox: Yeah, linked back to the help for the full list, just the important ones here makes sense to me.

rust-ceramic/src/docker.md
3benbox: Now requires https://github.com/3box/rust-builder/pull/20

So that we can update the version number during a PR.

PR #283, opened 2024-03-08
feat: report error if event network does not match

nathanielc:
Prior to this change events written via the API that do not match the configured network of the node were stored but never synchronized as interests are specific to the configured network.

Now an error is returned from the API if the network of the event does not match the configured network.

NOTE: This will eventually be unneeded logic once event ids are removed from the API.


3benbox (approved):
LGTM

api/src/server.rs
```diff
@@ -307,6 +307,27 @@ where
 
         let event_id = decode_event_id(&id)?;
         let event_data = decode_event_data(&data)?;
+
+        if let Some(network_id) = event_id.network_id() {
```
samika98, line 311: nit : can we move this logic to network.rs add a method there and maybe remove multiple if statements for better readability
```
Suggestion to do something like this : 
fn process_event(id: &str, data: &str, self_network: Network) -> Result<EventsPostResponse> {
    let event_id = decode_event_id(id)?;
    let event_data = decode_event_data(data)?;
    let network_id = event_id.network_id().ok_or_else(|| anyhow!("event does not have network id"))?;
    let event_network = Network::try_from_id(network_id)?;

    if event_network != self_network {
        return Ok(EventsPostResponse::BadRequest(format!(
            "event network does not match node network: {} != {}",
            event_network, self_network
        )));
    }

    // Proceed with event processing...
}
```

samika98 (approved):
LGTM

nathanielc:
Closing, we are removing the event id from the API so this error case is no longer possible.

nathanielc:
https://github.com/ceramicnetwork/rust-ceramic/pull/335

PR #284, opened 2024-03-08
feat: use standard error response body for API 

dav1do:
Addresses WS1-1518.

All Recon API errors are now of the format `{"message": "some text"}` instead of just a string. In order to make sure I didn't miss anything, I moved all the logic inside the server struct impl, rather than the trait. This allows using `or_else(|err| Ok(resp::InternalServerError(err))`. We don't return the swagger ApiError struct in any cases. The diff looks a bit intimidating, but it's pretty much just moving some things around. The responses change from `<_,ApiError>` to `<_, ErrorResponse>` and the trait impl maps them.

### DIFF
--- a/api-server/.openapi-generator/FILES
+++ b/api-server/.openapi-generator/FILES
@@ -3,6 +3,7 @@
 Cargo.toml
 README.md
 api/openapi.yaml
+docs/ErrorResponse.md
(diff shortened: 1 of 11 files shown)


smrz2001 (approved):
:shipit:

PR #287, opened 2024-03-12
fix: remove loud warn from Recon

nathanielc:
Prior to this change we would emit a warning every time we connected to a peer that did not speak the recon protocol. As we connect to many peers for the DHT that do not speak Recon the warn was triggered a lot. We do not need the warn as its not generally helpful debugging information.


stbrody:
I thought we removed the DHT entirely?  Is there a reason we still need it for peer discovery?

PR #289, opened 2024-03-14
feat: validate time events

AaronGoldman:
This adds a command `ceramic-one events validate`
```sh
rust-ceramic> ./target/release/ceramic-one events validate -h
Usage: ceramic-one events validate [OPTIONS] --cid <CID> --ethereum-rpc-url <ETHEREUM_RPC_URL>

Options:
  -s, --store-dir <STORE_DIR>
          Path to storage directory [env: CERAMIC_ONE_STORE_DIR=]
  -c, --cid <CID>
          CID of the block to validate
  -e, --ethereum-rpc-url <ETHEREUM_RPC_URL>
          Ethereum RPC URL, e.g. ETHEREUM_RPC_URL=https://mainnet.infura.io/v3/<api_key> [env: ETHEREUM_RPC_URL=]
  -h, --help
          Print help
```

This logic will be used to validate events that come in by recon or API in a later PR.
When addressing the feadback I will sqwash and fixup the commit mesages.


one/src/cbor_value.rs
```diff
@@ -0,0 +1,285 @@
...
+}
+
+impl CborValue {
+    /// Parse bytes to a single CborValue
+    /// This will error on CBOR seq.
+    pub fn parse(bytes: &[u8]) -> Result<Self> {
+        let mut decoder = Decoder::new(bytes);
+        let c = CborValue::next(&mut decoder);
+        if decoder.position() != bytes.len() {
+            debug!("decoder not at end {}/{}", decoder.position(), bytes.len());
```
dav1do, line 77: should this be a warn (or at least info?) i guess we sort of have a "if you log it, you handled it rule" but it'd be nice if the error returned had a bit more info about why it happened (does this mean you somehow packed multiple cbor values into a single thing, or it was just bad input?).

one/src/cbor_value.rs
```diff
@@ -0,0 +1,285 @@
...
+        if let CborValue::Tag((Tag::Unassigned(inner_tag), boxed_value)) = &self {
+            if inner_tag == &tag {
+                return Some(boxed_value);
+            }
+        }
+        None
+    }
+
+    /// get a printable string of the type_name
+    pub fn type_name(&self) -> String {
```
dav1do, line 177: could this just return &'static str? I don't think there's a good way to use std::fmt::Display for this behavior, but I don't think we need to allocate the response.

one/src/cbor_value.rs
```diff
@@ -0,0 +1,285 @@
...
+    /// If the CborValue is a CborValue::Array get the CborValue at index.
+    /// If it is not a CborValue::Array or index is out of bounds returns None.
+    pub fn get_index(&self, index: usize) -> Option<&CborValue> {
+        self.as_array()?.get(index)
+    }
+
+    /// If the CborValue is a CborValue::Map get the CborValue at index.
+    /// If it is not a CborValue::Map or key is missing returns None.
+    pub fn get_key(&self, key: &str) -> Option<&CborValue> {
+        self.as_map()?.get(&CborValue::String(key.to_owned()))
```
dav1do, line 206: my general API sense is that you should try to avoid allocating internally as the "principle of least surprise". since you need an owned string, my opinion is we should take `key: String` but it's fine.

also, you could potentially use a `Cow<'_, str>` as the variant body

one/src/ethereum_rpc.rs
```diff
@@ -0,0 +1,178 @@
...
+    }
+}
+
+impl EthRpc for HttpEthRpc {
+    async fn root_time_by_transaction_cid(&self, cid: Cid) -> Result<RootTime> {
+        // transaction to blockHash, blockNumber, input
+        let tx_hash = format!("0x{}", hex::encode(cid.hash().digest()));
+        let json: serde_json::Value = self.eth_transaction_by_hash(&tx_hash).await?;
+        debug!("txByHash response: {}", json);
+        let Some(result) = json.get("result") else {
```
dav1do, line 105: I think you could use `.ok_or_else(|| anyhow!())?` instead if/else for these

one/src/ethereum_rpc.rs
```diff
@@ -0,0 +1,178 @@
...
+        let Some(block_number) = block_number.as_str() else {
+            return Err(anyhow!("missing field result.blockNumber was not a string"));
+        };
+        let Ok(block_number) = get_timestamp_from_hex_string(block_number) else {
+            return Err(anyhow!(
+                "missing field result.blockNumber was not a hex timestamp"
+            ));
+        };
+
+        let Some(input) = result.get("input") else {
```
dav1do, line 126: maybe it's worth defining a struct for the response and deserializing to that rather than getting keys out of the json map?

dav1do (commented):
A few thoughts on style, didn't quite have a chance to finish review today but wanted to leave these thoughts.

stbrody:
Outside of testing, is there a reason to expose this as a command?  When would a user or node operator run this command?  What would they be expected to do if the command failed or found invalid events in the local store?

one/src/lib.rs
```diff
@@ -317,7 +320,7 @@ impl Daemon {
         };
         info.apply_to_metrics_config(&mut metrics_config);
 
-        // Currently only an info metric is recoreded so we do not need to keep the handle to the
+        // Currently only an info metric is recorged so we do not need to keep the handle to the
```
stbrody, line 323: recorded?

one/src/events.rs
```diff
@@ -28,9 +35,25 @@ pub struct SlurpOpts {
...
+    #[arg(short, long, env = "CERAMIC_ONE_STORE_DIR")]
+    store_dir: Option<PathBuf>,
+
+    /// CID of the block to validate
+    #[arg(short, long)]
+    cid: Cid,
+
+    /// Ethereum RPC URL, e.g. ETHEREUM_RPC_URL=https://mainnet.infura.io/v3/<api_key>
+    #[arg(short, long, env = "ETHEREUM_RPC_URL")]
+    ethereum_rpc_url: String,
```
stbrody, line 50: this struct seems to represent the args for a single request to validate a single TimeEvent.  Copying around the eth_rpc_url for every single event we want to validate seems unnecessary?  Should we either store the eth_rpc_url as a member variable of a validator instance that is used for all validation requests, or take this by reference so we're not copying it for each validation request?

one/src/events.rs
```diff
@@ -48,17 +71,214 @@ async fn slurp(opts: SlurpOpts) -> Result<()> {
...
+            let path: String = time_event.path(&["path"]).try_into()?;
+
+            // If prev not in root then TimeEvent is not valid.
+            if !prev_in_root(prev, proof_root, path, block_store).await? {
+                return Err(anyhow!("prev {} not in root {}", prev, proof_root));
+            }
+
+            // if root in root_store return timestamp.
+            // note: at some point we will need a negative cache to exponentially backoff eth_getTransactionByHash
+            if let Ok(Some(timestamp)) = root_store.get(proof_root.hash().digest()).await {
```
stbrody, line 228: Here we're only comparing the root CID to what we validated previously and put into the root store, but it's possible that the last time we saw this root it came in a TimeEvent with a different txHash.  If we see a new TimeEvent where the txHash doesn't match the root given, but the root is one we've seen before, we'd consider that valid when it should be considered invalid.

I think the root_store needs to also store the txHash associated with that root, and we need to double-check it matches here before we assume we can re-use the data in the root store without validating directly against ethereum.

one/src/events.rs
```diff
@@ -156,3 +376,215 @@ async fn migrate_from_database(input_ceramic_db: PathBuf, store: SQLiteBlockStor
     );
     result.unwrap()
 }
+
+#[derive(Debug, Clone)]
+pub struct SQLiteRootStore {
+    pool: SqlitePool,
+}
+
+impl SQLiteRootStore {
```
stbrody, line 385: Can we add some comments here on what this is storing and how it is used please?

stbrody (changes requested):

AaronGoldman:
> Outside of testing, is there a reason to expose this as a command? When would a user or node operator run this command? What would they be expected to do if the command failed or found invalid events in the local store?

First, the main reason to make this PR is to have a place to get the validation corect before we use these functions in the normal Event flows.

I would expect this to be run when debuging. Some event is not being returned from the node and an operator wants to know why. One thing to check is that all the blocks are present and valid. This would ether conferm that validation is not the problem or tell the operator why they are getting invalid events. e.g. you did not get anchored in time.

in general I am imaganing the `./ceramic-one events *`  to be comands like slurping in blocks (backup/restore) or scanning for consistency(recheck an untrusted root store).

one/src/cbor_value.rs
AaronGoldman: I disagree asking to decode a thing and it not being valid is a normal thing that can happen. I would prefer not to log at all unless the caller has a reson to beleve that a failed parse is suspicions. 

What info would be helpfull for the Err?
The two main resons for a not at end error is ether it was `CBOR seq` a sequence of CBOR objects using the fact that CBOR knows its length to not need framing or it is just not a CBOR object at all so it is obviously not a formed CBOR object.

one/src/cbor_value.rs
AaronGoldman: I will return `&'static str`

one/src/cbor_value.rs
AaronGoldman: https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.get

```rs
pub fn get<Q>(&self, key: &Q) -> Option<&V>
where
    K: Borrow<Q> + Ord,
    Q: Ord + ?Sized,
```

If we can make an `==` for `&str` and `CborValue::String` we may be able to avoid making a `CborValue::String` to use `std::collections::BTreeMap.get(key)`.

one/src/events.rs
AaronGoldman: This makes sense the role of the `txHash` is to be a key into the root store. Whether that root store is SQLite or Ethereum Should not matter.

```sql
CREATE TABLE IF NOT EXISTS roots (
  txHash BLOB, 
  root BLOB, 
  timestamp INTEGER, 
  PRIMARY KEY(txHash)
);
```

one/src/cbor_value.rs
dav1do: Yeah, that's fair. I guess I wasn't sure if this was internal parsing and it should only fail if we did something wrong, in which case it's nice to know and have all the info about what, or if it's validating any data in which was we expect to fail regularly. I don't know what would be helpful, it's fine to close.

one/src/events.rs
stbrody: this seems fine for now when it's only being used in the CLI command which will just validate a single event and then shut down.  But once we start validating every incoming time event on the standard codepaths, I don't think we're going to want to copy the rpc url for every single time event we need to validate.

one/src/cbor_value.rs
dav1do: Yeah, we could do that with Deref but it's probably not worth it (will possibly have some surprising behavior). 

It's fine to close this, I was just commenting on the api signature generally. My opinion is to use `Into<String>` or `String` rather than `&str` and allocating internally. 

If we create and share/copy a lot of `String(String)` variants, changing it to `String(Cow<_,str>)` might help and not require a struct lifetime but I'm not sure how easy it'd be to work with. My point was just that you could create that from the `&str` without needing `to_owned()`.

one/src/events.rs
```diff
@@ -156,3 +376,215 @@ async fn migrate_from_database(input_ceramic_db: PathBuf, store: SQLiteBlockStor
...
+        Ok(sqlx::query("SELECT timestamp FROM roots WHERE root = ?;")
+            .bind(root)
+            .fetch_optional(self.pool.reader())
+            .await?
+            .map(|row| row.get::<'_, i64, _>(0)))
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
dav1do, line 427: are there more tests in js-ceramic we should add? it seems like we should have at least an invalid time event test

one/src/lib.rs
AaronGoldman: changed

dav1do (approved):
There is more work to do before this is "prod ready" but we're still working on it we can make breaking changes, so I think merging it and continuing to work on it this cycle is worthwhile.

dav1do (approved):

one/src/cbor_value.rs
```diff
@@ -0,0 +1,285 @@
...
+use enum_as_inner::EnumAsInner;
+use minicbor::{data::Tag, data::Type, Decoder};
+use ordered_float::OrderedFloat;
+use std::{collections::BTreeMap, ops::Index};
+use tracing::debug;
+
+/// CborValue
+/// a enum for working with dynamic values from decoding CBOR values.
+#[derive(Eq, PartialEq, Ord, PartialOrd, Clone, Debug, EnumAsInner)]
+pub enum CborValue {
```
stbrody, line 12: btw I'm pretty sure that there is already a library for working with Cbor objects.  Filecoin and IPC use CBOR extensively in Rust, I'm sure they have a library we can use.  Fine to stick with this for now to get this PR in, but we should look into if there's an existing library we can use for this.

one/src/ethereum_rpc.rs
```diff
@@ -0,0 +1,175 @@
...
+            }))
+            .send()
+            .await?
+            .json()
+            .await?)
+    }
+}
+
+impl EthRpc for HttpEthRpc {
+    async fn root_time_by_transaction_cid(&self, cid: Cid) -> Result<RootTime> {
```
stbrody, line 101: I don't usually see ethereum transaction ids encoded as CIDs. Any reason we're using a Cid for this and then converting to tx_hash instead of just passing around the tx_hash everywhere instead?

one/src/ethereum_rpc.rs
```diff
@@ -0,0 +1,175 @@
...
+            root,
+            block_hash: block_hash.to_string(),
+            timestamp,
+        })
+    }
+}
+
+fn get_root_from_input(input: &str) -> Result<Cid> {
+    if let Some(input) = input.strip_prefix("0x97ad09eb") {
+        // Strip "0x97ad09eb" from the input and convert it into a cidv1 - dag-cbor - (sha2-256 : 256)
```
stbrody, line 158: why are we stripping this out?  Can we use a named constant for this string and a comment for what it represents?

one/src/ethereum_rpc.rs
```diff
@@ -0,0 +1,175 @@
...
+        let timestamp = get_timestamp_from_hex_string(timestamp)?;
+        Ok(RootTime {
+            root,
+            block_hash: block_hash.to_string(),
+            timestamp,
+        })
+    }
+}
+
+fn get_root_from_input(input: &str) -> Result<Cid> {
```
stbrody, line 156: can we add a comment what this function is doing?

one/src/ethereum_rpc.rs
```diff
@@ -0,0 +1,175 @@
...
+            .context("missing field result.blockNumber was not a string")?;
+        let block_number = get_timestamp_from_hex_string(block_number)
+            .context("missing field result.blockNumber was not a hex timestamp")?;
+        let input = result.get("input").context("missing field result.input")?;
+        let input = input
+            .as_str()
+            .context("missing field result.input was not a string")?;
+        let root = get_root_from_input(input)?;
+
+        // Get the block with block_hash if latest block number minus result.number grater then 3 the block is stable.
```
stbrody, line 127: this is interesting.  This is a new check, I'm pretty sure js-ceramic doesn't do this when validating TimeEvents.

I'm not necessarily opposed, just surprised we're changing the logic here.  This is also going to create an extra call to the eth rpc endpoint and increase the latency of validating time events.  Is this something we want to do right now or maybe leave as a possible future improvement?

one/src/ethereum_rpc.rs
```diff
@@ -0,0 +1,175 @@
...
+        debug!("blockByHash response: {}", json);
+
+        let Some(timestamp) = json["result"]["timestamp"].as_str() else {
+            return Err(anyhow!(
+                "missing field result.timestamp or was not a string"
+            ));
+        };
+        let timestamp = get_timestamp_from_hex_string(timestamp)?;
+        Ok(RootTime {
+            root,
```
stbrody, line 149: I'm pretty sure this "root" field is the CID of the root of the merkle tree that was anchored in this transaction, but it's not immediately obvious from reading this function.  Some more comments throughout would help

one/src/events.rs
```diff
@@ -28,9 +35,25 @@ pub struct SlurpOpts {
     output_ceramic_path: Option<PathBuf>,
 }
 
+#[derive(Args, Debug)]
+pub struct ValidateOpts {
+    /// Path to storage directory
+    #[arg(short, long, env = "CERAMIC_ONE_STORE_DIR")]
+    store_dir: Option<PathBuf>,
+
+    /// CID of the block to validate
```
stbrody, line 44: of the block, or the transaction?

Can we call this arg something more descriptive than `cid` so it's clearer what it actually represents?  I have a bit of a pet peeve against variables with the same name as their type, it's just a missed opportunity to convey more information to the reader.

one/src/events.rs
```diff
@@ -48,17 +71,221 @@ async fn slurp(opts: SlurpOpts) -> Result<()> {
...
+                println!("UnsignedDataEvent({}) is not signed", opts.cid);
+            }
+            validate_data_event_payload(&opts.cid, &block_store, None).await?;
+        }
     }
     Ok(())
 }
 
+async fn validate_data_event_envelope(cid: &Cid, block_store: &SQLiteBlockStore) -> Result<String> {
+    let block = block_store.get(cid.to_owned()).await?.unwrap();
```
stbrody, line 150: nitpick, I'd actually prefer to leave all of this out of this PR, and change this function to just panic with "not yet implemented", then keep the data event validation in a separate PR so we're not doing too many things at once in a single PR.  I'm mostly focusing my efforts in this review on the time event validation, not going to read this function very closely right now, but it would probably deserve it's own proper review at some point.

one/src/events.rs
```diff
@@ -48,17 +71,221 @@ async fn slurp(opts: SlurpOpts) -> Result<()> {
...
+        }
+        return Ok(());
+    }
+    // If the CID is a DAG-CBOR, the event is either a Time Event or an unsigned Init Event. In this case, we'll pull
+    // the block from the store and use the presence of the "proof" field to determine whether this is a Time Event or
+    // an unsigned Init Event.
+    //
+    // When validating events from Recon, the block store will be a CARFileBlockStore that wraps the CAR file received
+    // over Recon.
+    if let Some(block) = block_store.get(opts.cid).await? {
```
stbrody, line 116: should we error or at least log a warning or something if the block isn't found in the block store?

one/src/events.rs
```diff
@@ -48,17 +71,221 @@ async fn slurp(opts: SlurpOpts) -> Result<()> {
...
+//
+// - If root not in local root store try to read it from txHash.
+// - Validated time is the time from the Root Store
+async fn validate_time_event(
+    cid: &Cid,
+    block_store: &SQLiteBlockStore,
+    root_store: &SQLiteRootStore,
+    eth_rpc: &impl EthRpc,
+) -> Result<i64> {
+    if let Some(block) = block_store.get(cid.to_owned()).await? {
```
stbrody, line 211: we already got the block from the block store in the caller, can we pass it in so we don't have to go back to the block store a second time?

one/src/events.rs
```diff
@@ -156,3 +383,236 @@ async fn migrate_from_database(input_ceramic_db: PathBuf, store: SQLiteBlockStor
...
+/// The EthereumRootStore is the authoritative root store but it is also immutable  
+/// once the blocks are final.
+/// We only pull each txHash once and store the root, blockHash and timestamp.
+/// After that Time Events validation can be done locally.
+#[derive(Debug, Clone)]
+pub struct SQLiteRootStore {
+    pool: SqlitePool,
+}
+
+impl SQLiteRootStore {
```
stbrody, line 398: Can we please add comments to clarify what "root" we are talking about here?  Especially once we have self-anchoring we are likely to have multiple "roots" - the root of my local tree plus the root of the tree submitted to ethereum.

stbrody (approved):
core logic looks correct.  Consider all comments optional

one/src/cbor_value.rs
AaronGoldman: Nathaniel is working on rust serde_dag_cbor to see where we can decode directly to nested structs.

one/src/ethereum_rpc.rs
AaronGoldman: We use CIDs to express the roots because that is what is in the 'proof' field of the Time Event. The problem is that this function needs a EthRpc to get the data from eth network but the fact that it is the Timestamp and the root that we want is ceramic logic not EthRPC logic so it should be moved to a function in Events that takes an EthRPC.

one/src/events.rs
AaronGoldman: I'll rename as `event`

PR #291, opened 2024-03-22
docs: added explicit deps to build steps

christianlavoie:
Minor doc update for shiny new computer setups wanting to compile using `cargo run`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,10 @@ Use the provided `Makefile` for basic actions to ensure your changes are ready f
     $ make check-fmt
     $ make test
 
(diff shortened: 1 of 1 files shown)


nathanielc (approved):
Looks good to me. We use conventional commit messages which is why CI is red. https://www.conventionalcommits.org/en/v1.0.0/#summary

Change the commit message to something like

docs: added explicit deps to build steps

With that CI should be happy. Thanks for the change.

nathanielc:
GitHub is annoying, it will use the PR title when there are multiple commits or use the commit message when there is just one commit. Since this change has a single commit the commit message is the bit being validated. Your PR title looks great, if you amend the commit with the same message CI should be happy.

PR #292, opened 2024-03-22
feat: migrations with new table format to de-dup blocks and support out of order delivery

dav1do:
Use sqlx for migrations that are run on startup. The tables have been reworked and this assumes there is no existing database on the node (we will likely need to nuke the volumes on the durable infra tests). Data migrations are handled via ceramic-one CLI commands, this is only to support schema migrations. 

- Database changes:
  - Rename tables (all prefixed `ceramic_one_`) and use a blocks table to normalize (particularly time) event data.
  - Renamed `key` columns to `id` to avoid reserved word, added `delivered` integer rather than relying on RowID which is sqlite specific and not reliable with vacuuming and deletion. The `value_retrieved` column was removed for a join on `event_block`.
  - Added `NOT NULL`, `DEFAULT`, `UNIQUE`, and `FOREIGN KEY` constraints
  
- General:
  - Deleted blockstore file and implemented blocks table in store crate.
  - Move rootstore to store crate and created migration.
  - Renamed model/Modelstore to event/EventStore
  - Use some query! macros and generated files for offline builds.
  - Switched to $ sql parameter binding to support postgres in future

- Future work:
  - Support CLI option to run migrations out of band (and support rollback one version)
  - Support dry-run for migrations
  
You can use `make check-queries` to apply migrations to your local sqlite file and generate the offline query data. That will be changed with postgres support (can't use query! since multiple URLs needed), but as an FYI it exists.

Closes [WS1-1580](https://linear.app/3boxlabs/issue/WS1-1580/setup-method-for-migrating-data-in-ceramic-one) and [WS1-1543](https://linear.app/3boxlabs/issue/WS1-1543/migrate-recon-storage-to-support-in-order-delivery)


p2p/src/node.rs
```diff
@@ -1639,6 +1639,7 @@ mod tests {
     }
 
     #[test(tokio::test)]
+    #[ignore]
```
dav1do, line 1642: should we keep this? it fails almost every time on osx but seems reliable on linux, haven't sorted out the issue yet.

migrations/20240321211121_block.up.sql
```diff
@@ -0,0 +1,7 @@
+-- Add up migration script here
+
```
dav1do, line 2: Do we want to create a dedicated schema rather than use public? Might make a change to another db (e.g. sqlite to postgres) easier if we were to ever support that.

migrations/20240321211311_event.up.sql
```diff
@@ -0,0 +1,18 @@
+-- Add up migration script here
+
+CREATE TABLE IF NOT EXISTS event (
+    id BLOB NOT NULL, -- network_id sort_value controller StreamID height event_cid
```
dav1do, line 4: do we want to normalize this to to something like the following to support a flat recon future? 

```sql
network_id integer not null,
sep_key text not null default 'model',
sep_value bytes not null,
controller text not null,
init_cid bytes not null
```

migrations/20240321211311_event.up.sql
```diff
@@ -0,0 +1,18 @@
...
+    cid BLOB UNIQUE, -- NOT NULL? the cid of the event as bytes no 0x00 prefix
```
dav1do, line 5: do we want to store event cid in this table or just use `select cid from event_block where id = $1`

should this be NOT NULL?

migrations/20240321211311_event.up.sql
```diff
@@ -0,0 +1,18 @@
...
+    ahash_0 INTEGER NOT NULL, -- the ahash is decomposed as [u32; 8]
+    ahash_1 INTEGER NOT NULL,
+    ahash_2 INTEGER NOT NULL,
+    ahash_3 INTEGER NOT NULL,
+    ahash_4 INTEGER NOT NULL,
+    ahash_5 INTEGER NOT NULL,
+    ahash_6 INTEGER NOT NULL,
+    ahash_7 INTEGER NOT NULL,
+    value_retrieved BOOL NOT NULL DEFAULT FALSE, -- bool with 0=false, 1=true. indicates if we have the value
```
dav1do, line 14: do we want to use `STRICT` tables? It removes bool and requires using integer and is generally more annoying. It's not recommended by the sqlite docs but could make a sqlite-> postgres migration easier. Shouldn't really be necessary since the rust data should be strongly typed so I left it off.

migrations/20240321211311_event.up.sql
dav1do: decided against this. it makes queries a lot more complicated. whatever we do now will likely require a recon protocol version bump and we are still a bit from landing at the flat recon implementation.

migrations/20240321211311_event.up.sql
dav1do: decided to use CID as the event PK that other tables reference and order_key as a unique index for recon splits

dav1do:
Sorry this got massive. [This branch](https://github.com/ceramicnetwork/rust-ceramic/compare/feat/ws1-1543-deliver-order-backup) is about what it was on a discovery/commit by commit basis. If it's too much, I can redo it with migrations, schema changes, query! use separate, but since we want to make these changes and the macros help validate, I just squashed it all into one when discussions landed on schema changes we agree on.

I'd suggest looking at the table definitions first. The queries were updated to match and keep the same functionality. The tests all pass with minor adjustments (the RowID -> delivered changes had the biggest impact).

one/src/events.rs
```diff
@@ -113,42 +115,45 @@ async fn validate(opts: ValidateOpts) -> Result<()> {
...
+            &HttpEthRpc::new(opts.ethereum_rpc_url),
+        )
+        .await;
+        match timestamp {
+            Ok(timestamp) => {
+                let rfc3339 = Utc
+                    .timestamp_opt(timestamp, 0)
+                    .unwrap()
+                    .to_rfc3339_opts(SecondsFormat::Secs, true);
+                println!(
```
dbcfd, line 134: log?

one/src/events.rs
```diff
@@ -200,93 +205,82 @@ async fn validate_data_event_payload(
...
+    let prev: Cid = time_event.path(&["prev"]).try_into()?;
+    let proof_block = block_store.get(&proof_cid).await?;
+    let proof_cbor = CborValue::parse(&proof_block.data)?;
+    let proof_root: Cid = proof_cbor.path(&["root"]).try_into()?;
+    let path: String = time_event.path(&["path"]).try_into()?;
+    let tx_hash_cid: Cid = proof_cbor.path(&["txHash"]).try_into()?;
+
+    // If prev not in root then TimeEvent is not valid.
+    if !prev_in_root(prev, proof_root, path, block_store).await? {
+        return Err(anyhow!("prev {} not in root {}", prev, proof_root));
```
dbcfd, line 229: anyhow::bail

store/src/sql/mod.rs
```diff
@@ -29,15 +31,16 @@ pub struct SqlitePool {
 impl SqlitePool {
     /// Connect to the sqlite database at the given path. Creates the database if it does not exist.
     /// Uses WAL journal mode.
-    pub async fn connect(path: impl AsRef<Path>) -> anyhow::Result<Self> {
+    pub async fn connect(path: impl AsRef<Path>, migrate: bool) -> anyhow::Result<Self> {
```
dbcfd, line 34: Can we enum the migrate flag? Or split to a connect and connect_with_migrate?

dav1do: Added an enum.. wrapped the memory pool in its own function so now we only call this from a few places in the binary rather than all over in test code

one/src/events.rs
dav1do: updated

one/src/events.rs
dav1do: we didn't have tracing in the cli yet.. I setup a subscriber now and changed almost all of the stdout/stderr messaging to use info/warn/error.

dbcfd (approved):

store/src/sql/mod.rs
```diff
@@ -1,14 +1,17 @@
+mod event;
```
AaronGoldman, line 1: I thought we standardized on `sql.rs` rather then `/sql/mod.rs` that the `/folder/mod.rs` pattern was the old way of laying out code?

AaronGoldman (approved):

.env.sample
```diff
@@ -0,0 +1,2 @@
+# DATABASE_URL="sqlite:://../ceramic_cicddb.sqlite" # relative to the store crate
+# SQLX_OFFLINE=true # overrides DATABASE_URL when set. should use .sqlx directory
```
smrz2001, line 2: minor: mind adding a new line?

.gitignore
```diff
@@ -10,3 +10,5 @@ peer*
 sync_test.sh
 
 openapi-generator-cli.jar
+.env
+ceramic_cicddb.sqlite*
```
smrz2001, line 14: minor: mind adding a new line?

migrations/sqlite/20240321211121_block.down.sql
```diff
@@ -0,0 +1,2 @@
+-- Add down migration script here
+DROP TABLE IF EXISTS "block";
```
smrz2001, line 2: minor: mind adding a new line?

migrations/sqlite/20240321211303_interest.down.sql
```diff
@@ -0,0 +1,3 @@
+-- Add down migration script here
+
+DROP TABLE IF EXISTS interest;
```
smrz2001, line 3: minor: mind adding a new line?

one/src/events.rs
```diff
@@ -64,32 +68,40 @@ async fn slurp(opts: SlurpOpts) -> Result<()> {
...
         "{} Opening output ceramic SQLite DB at: {}",
         Utc::now().to_rfc3339_opts(SecondsFormat::Secs, true),
         output_ceramic_path.display()
     );
 
-    let pool = SqlitePool::connect(output_ceramic_path).await.unwrap();
-    let block_store = SQLiteBlockStore::new(pool).await.unwrap();
+    let pool = SqlitePool::connect(output_ceramic_path, Migrations::Apply)
+        .await
+        .with_context(|| "Failed to connect to database")?;
```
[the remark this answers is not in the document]
dav1do: ah, yeah. not sure why I thought i needed with_context

migrations/sqlite/20240321211311_event.up.sql
```diff
@@ -0,0 +1,17 @@
+-- Add up migration script here
+
+CREATE TABLE IF NOT EXISTS "event" (
+    order_key BLOB NOT NULL UNIQUE, -- network_id sep_key sep_value controller stream_id event_cid
+    ahash_0 INTEGER NOT NULL, -- the ahash is decomposed as [u32; 3]
```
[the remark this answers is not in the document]
dav1do: good catch. i changed this while we were looking making it 96 bits and missed it

migrations/sqlite/20240321211121_block.up.sql
```diff
@@ -0,0 +1,7 @@
+-- Add up migration script here
+
+CREATE TABLE IF NOT EXISTS "block" (
```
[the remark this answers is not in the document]
dav1do: I like writing them as idempotent generally but it would be important to know if the table was wrong. 

I guess I could either remove the IF, or run a SELECT columna,..columnz from table; as a verify/assert that it worked.. if the columns don't match it should fail. sqitch offers a hook for verify but it's not quite the same in sqlx.

migrations/sqlite/20240321211121_block.down.sql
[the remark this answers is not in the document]
dav1do: sqlx requires every migration have a down if any have a down.. but we don't have to put anything in it. we can't change it later, so I was writing them in a way to revert completely if desired.. but they are definitely dangerous. I'm happy to put whatever we deem the right thing in the down statements. 

I wanted to enforce rollback to X (and in this case basically nothing) when we actually support the cli migrations in c1.

store/src/sql/mod.rs
dav1do: uhh.. maybe? I like the folder/mod.rs to control the exports, but I guess I can make store/src/sql.rs and it will be the same as the store/src/sql/mod.rs file. I kind of like keeping the module exports closer to the folder structure but I'll do whatever, especially if that's the preference

migrations/sqlite/20240321211303_interest.down.sql
dav1do: yep, will do

migrations/sqlite/20240321211121_block.up.sql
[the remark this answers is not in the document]
dav1do: I went with the SELECT to verify approach.. made sure they do fail with an invalid column

```
---- events::tests::test_validate_time_event stdout ----
thread 'events::tests::test_validate_time_event' panicked at one/src/events.rs:434:58:
called `Result::unwrap()` on an `Err` value: while executing migrations: error returned from database: (code: 1) no such column: tx_hash1```

File: store/src/sql/mod.rs [hunk 17]
```diff
@@ -26,18 +29,29 @@ pub struct SqlitePool {
     reader: sqlx::SqlitePool,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+/// Now to handle ounstanding database migrations.
```
smrz2001, line 33:
```suggestion
/// Now to handle outstanding database migrations.
```

smrz2001 (approved):
Just a minor typo, otherwise LGTM!

PR #294, opened 2024-03-26
fix: use TAG_LATEST

3benbox:
Only tag the workflow docker image with "latest" when it's a merge to main.

Currently we're getting branches tagged with latest.

### DIFF
--- a/.github/workflows/cd-to-infra.yml
+++ b/.github/workflows/cd-to-infra.yml
@@ -47,12 +47,12 @@ jobs:
           DEPLOY_TAG=$RELEASE_TAG
         fi
         # Tag with latest if this is a release (not pre-release)
-        if [[ "${{ github.event.action }}" == "published" && "${{ github.event.release.prerelease }}" == "true" ]]; then
-          TAG_LATEST=false
-        elif [[ "${{ github.event.action }}" == "released" ]]; then
+        if [[ "${{ github.event_name }}" == "push" ]]; then
           TAG_LATEST=true
+        elif
+          TAG_LATEST=false
         fi
-        make SHA="${{ github.SHA }}" SHA_TAG="$SHA_TAG" RELEASE_TAG="$RELEASE_TAG" publish-docker
+        make SHA="${{ github.SHA }}" SHA_TAG="$SHA_TAG" RELEASE_TAG="$RELEASE_TAG" TAG_LATEST="$TAG_LATEST" publish-docker
         echo "Deploy tag:"
         echo ${DEPLOY_TAG}
         echo "deploy_tag=${DEPLOY_TAG}" >> $GITHUB_OUTPUT
--- a/.github/workflows/cd-to-infra.yml
+++ b/.github/workflows/cd-to-infra.yml
@@ -46,7 +46,7 @@ jobs:
           # Use the release tag to deploy, if one is available.
           DEPLOY_TAG=$RELEASE_TAG
         fi
-        # Tag with latest if this is a release (not pre-release)
+        # Tag with latest if this is a push to master
         if [[ "${{ github.event_name }}" == "push" ]]; then
           TAG_LATEST=true
         elif
--- a/.github/workflows/cd-to-infra.yml
+++ b/.github/workflows/cd-to-infra.yml
@@ -46,6 +46,12 @@ jobs:
           # Use the release tag to deploy, if one is available.
           DEPLOY_TAG=$RELEASE_TAG
         fi
+        # Tag with latest if this is a release (not pre-release)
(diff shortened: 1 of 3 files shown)


.github/workflows/rust-build-and-test.yml
```diff
@@ -97,7 +97,13 @@ jobs:
...
           BUILD_TAG=$(echo ${{ github.sha }} | head -c 12)
         fi
-        make SHA_TAG="$BUILD_TAG" publish-docker
+        # Tag with latest if merge to main
+        if [ "${{ github.event_name }}" == "merge_group" ]; then
+          TAG_LATEST=true
+        else
+          TAG_LATEST=false
+        fi
+        make SHA_TAG="$BUILD_TAG" TAG_LATEST="$TAG_LATEST" publish-docker
```
smrz2001, line 106: Can revert this section, actually. The `cd-to-infra.yml` workflow detects pushes to `main` and so there you could add `TAG_LATEST="true"` in the `make` command there:
```
make SHA="${{ github.SHA }}" SHA_TAG="$SHA_TAG" RELEASE_TAG="$RELEASE_TAG" TAG_LATEST="true" publish-docker
```

ci-scripts/publish.sh
```diff
@@ -11,7 +11,6 @@
 # to login to docker. That password will be valid for 12h.
 
 docker buildx build --load -t 3box/ceramic-one .
-docker tag 3box/ceramic-one:latest public.ecr.aws/r5b3e0r5/3box/ceramic-one:latest
```
smrz2001, line 14: Good catch!!!

.github/workflows/rust-build-and-test.yml
3benbox: Removed!

smrz2001: Wait haha, I meant can keep the TAG_LATEST change in the publish script, revert changes to the test workflow yml, and then add the TAG_LATEST=true to the CD workflow.

3benbox: Ah, updated publish to set the latest tag when a release is done.

We don't need the TAG_LATEST anymore, yeah?

Updated for that.

.github/workflows/cd-to-infra.yml
```diff
@@ -46,6 +46,12 @@ jobs:
           # Use the release tag to deploy, if one is available.
           DEPLOY_TAG=$RELEASE_TAG
         fi
+        # Tag with latest if this is a release (not pre-release)
```
smrz2001, line 49: Can you switch this check around? Should tag `latest` only if this is a push to `main`, versus for a release.

.github/workflows/rust-build-and-test.yml
```diff
@@ -97,7 +97,7 @@ jobs:
         else
           BUILD_TAG=$(echo ${{ github.sha }} | head -c 12)
         fi
-        make SHA_TAG="$BUILD_TAG" publish-docker
+        make SHA_TAG="$BUILD_TAG" TAG_LATEST="false" publish-docker
```
smrz2001, line 100: Leaving this out would work as well, but being explicit might be more readable.

.github/workflows/cd-to-infra.yml
```diff
@@ -46,6 +46,12 @@ jobs:
...
+        if [[ "${{ github.event.action }}" == "published" && "${{ github.event.release.prerelease }}" == "true" ]]; then
+          TAG_LATEST=false
+        elif [[ "${{ github.event.action }}" == "released" ]]; then
+          TAG_LATEST=true
+        fi
         make SHA="${{ github.SHA }}" SHA_TAG="$SHA_TAG" RELEASE_TAG="$RELEASE_TAG" publish-docker
```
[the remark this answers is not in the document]
3benbox: doh

.github/workflows/cd-to-infra.yml
3benbox: updated!

.github/workflows/rust-build-and-test.yml
3benbox: I live in fear of defaults

.github/workflows/cd-to-infra.yml
```diff
@@ -46,7 +46,13 @@ jobs:
           # Use the release tag to deploy, if one is available.
           DEPLOY_TAG=$RELEASE_TAG
         fi
-        make SHA="${{ github.SHA }}" SHA_TAG="$SHA_TAG" RELEASE_TAG="$RELEASE_TAG" publish-docker
+        # Tag with latest if this is a release (not pre-release)
```
smrz2001, line 49: This comment can be updated as well.

smrz2001 (approved):
Just one editorial comment otherwise LGTM! 🚀

PR #299, opened 2024-03-28
feat: simplify event id

nathanielc:
This makes two changes to the event ID:

1. The event height is removed
2. The sep_value bytes are directly hashed instead of a concatenation of the sep_key|sep_value string.

We need to finish [WS2-3009](https://linear.app/3boxlabs/issue/WS2-3009/determine-if-we-can-remove-eventids-from-the-public-api-entirely) first this way we do not have to update the js event id library but can just delete it instead. Event Ids will be an implementation detail of Recon.

Fixes: WS2-3130
Fixes: WS2-3066

### DIFF
--- a/api/src/server.rs
+++ b/api/src/server.rs
@@ -361,8 +361,8 @@ where
 
     pub async fn get_events_sort_key_sort_value(
         &self,
-        sort_key: String,
-        sort_value: String,
+        sep_key: String,
+        sep_value: String,
         controller: Option<String>,
         stream_id: Option<String>,
         offset: Option<i32>,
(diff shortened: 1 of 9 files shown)


store/src/sqlite/model.rs
```diff
@@ -1154,13 +1134,53 @@ mod test {
         // No keys in range
         let ret = recon::Store::first_and_last(
```
dbcfd, line 1135: The comment seems to indicate that there should be no keys in range, but it looks like the test now returns some?

nathanielc: fixed

dbcfd (approved):

dav1do (approved):

dav1do:
Closing as this was included in #335

PR #300, opened 2024-04-01
feat: don't mark peers failed when we error during sync

dav1do:
We have a "transient" error type (for now) to indicate that we had an internal failure during sync and should retry, rather than mark the peer as failed, which terminates the conversation.

This also incorporates the sync_id changes in #304.

### DIFF
--- a/recon/src/libp2p/handler.rs
+++ b/recon/src/libp2p/handler.rs
@@ -201,7 +201,8 @@ where
                         }
                         Err(e) => {
                             // TODO: differentiate between transient and permanent errors
-                            // for now, we don't want to try to keep talking so we treat all errors as transient
+                            // for now, we want to keep talking so we treat all errors as transient
+                            // but in the future, we will want to shut down the connection if the error is permanent
                             return Poll::Ready(ConnectionHandlerEvent::NotifyBehaviour(
                                 FromHandler::TransientError(e),
                             ));
--- a/recon/src/libp2p/handler.rs
+++ b/recon/src/libp2p/handler.rs
@@ -123,6 +123,7 @@ pub enum FromBehaviour {
 pub enum FromHandler {
     Started { stream_set: StreamSet },
     Succeeded { stream_set: StreamSet },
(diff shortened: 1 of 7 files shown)


recon/src/libp2p/handler.rs
```diff
@@ -200,12 +201,30 @@ where
...
                         Err(e) => {
                             return Poll::Ready(ConnectionHandlerEvent::NotifyBehaviour(
-                                FromHandler::Failed(e),
+                                FromHandler::TransientError(e),
                             ))
                         }
                     }
                 }
             }
+            State::Inbound(stream) => {
```
AaronGoldman, line 210: Why are Inbound and Outbound different?

dav1do: My thinking was we want to mark inbound as errors still (potentially, since it might have been the peers fault making it failed makes sense). But outbound means we encountered a problem on our node, so we shouldn't mark the peer as failed.

recon/src/libp2p/handler.rs
```diff
@@ -200,12 +201,30 @@ where
...
+                if let Poll::Ready(result) = stream.poll_unpin(cx) {
+                    self.transition_state(State::Idle);
+                    match result {
+                        Ok(stream_set) => {
+                            return Poll::Ready(ConnectionHandlerEvent::NotifyBehaviour(
+                                FromHandler::Succeeded { stream_set },
+                            ));
+                        }
+                        Err(e) => {
+                            // we don't want to mark them down when we fail...
```
dav1do, line 220: remove this comment

smrz2001 (approved):
🚀

recon/src/metrics.rs
```diff
@@ -12,7 +12,7 @@ use prometheus_client::{
 };
 
 use crate::{
-    protocol::{InitiatorMessage, ResponderMessage},
```
smrz2001, line 15: Thanks for fixing this 😅

PR #302, opened 2024-04-05
feat: add postgres support in addition to sqlite for recon

dav1do:
All `Store` tests run with postgres and sqlite. Postgres tests no-op unless `PG_TESTS=1`, so that `cargo test` still passes even if you haven't set up a docker postgres image. In CI we can run them all in order to make sure the postgres type/query handling is valid. The `make check-migrations-ci` command handles setting up the dbs, applying migrations, running tests (will duplicate some tests with the current make tests command but tries to use the same parameters to avoid rebuildling things), and then cleaning up.

The database can be specified with `ceramic-one daemon --database-url $DB_URL` or `CERAMIC_ONE_DATABASE_URL="..." ceramic-one daemon`. Postgres or sqlite will be chosen based on the URL prefix, and defaults to putting a sqlite file in the "default directory" (see `DaemonOpts::default_directory`).

I also added `put()` to the `iron_bitswap::Store` trait as we were going through the concrete struct before. I couldn't get away from using concrete types, so I select the database and then use generics to bind them to the flavors of trait we need. We should be able to simplify/consolidate some of these traits now that we rely on `&self` rather than `&mut self` in recon::Store, but this PR is already big.


dav1do:
I attempted to get the postgres tests/docker running in CI and didn't get the networking right, so I reverted this branch back to where it was and I'm going to make a new branch to follow up with that.

dav1do:
> I attempted to get the postgres tests/docker running in CI and didn't get the networking right, so I reverted this branch back to where it was and I'm going to make a new branch to follow up with that.

#308 handles this.

dav1do:
> > I attempted to get the postgres tests/docker running in CI and didn't get the networking right, so I reverted this branch back to where it was and I'm going to make a new branch to follow up with that.
> 
> #308 handles this.

Which has now been incorporated into this branch... along with _one_ more change to capture store related metrics on the API side.

PR #303, opened 2024-04-05
feat: store bench

dbcfd:
Use criterion to benchmark store inserts

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,7 +61,7 @@ clap = { version = "4", features = ["derive", "env"] }
 clap_mangen = "0.2.2"
 console = { version = "0.15", default-features = false }
 console-subscriber = "0.1.7"
-criterion = "0.4"
+criterion2 = "0.7.0"
 crossterm = "0.25"
 ctrlc = "3.2.2"
 dag-jose = "0.1.3"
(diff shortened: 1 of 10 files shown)


Cargo.toml
```diff
@@ -201,3 +203,6 @@ authors = [
 ]
 license = "Apache-2.0/MIT"
 repository = "https://github.com/3box/rust-ceramic"
+
+[patch.crates-io]
+criterion2 = { git = "https://github.com/dbcfd/criterion2.rs.git", branch = "feat/async-setup" }
```
dav1do, line 208: why do we need a fork?

Cargo.toml
```diff
@@ -61,7 +61,7 @@ clap = { version = "4", features = ["derive", "env"] }
 clap_mangen = "0.2.2"
 console = { version = "0.15", default-features = false }
 console-subscriber = "0.1.7"
-criterion = "0.4"
+criterion2 = "0.7.0"
```
dav1do, line 64: why criterion2 vs criterion?

Cargo.toml
dbcfd: Needed the PR I put up merged and released.

Cargo.toml
dbcfd: criterion no longer seems to be actively  maintained

Cargo.toml
dav1do: It looks like it's been released? Mind updating it when you resolve the conflicts?

dav1do (approved):

PR #304, opened 2024-04-06
feat: add sync_id to recon conversations

smrz2001:


recon/src/protocol.rs
```diff
@@ -262,6 +264,7 @@ where
 
         self.role.close().await.context("closing stream")?;
         self.metrics.record(&ProtocolRun(start.elapsed()));
+        trace!("sync conversation complete: {}", self.role.sync_id());
```
dav1do, line 267: FYI that with tracing macros you can use `sync_id=%self.role.sync_id()` to get structured output with the field name in the logs.

recon/src/protocol.rs
```diff
@@ -353,13 +358,18 @@ where
         // Use a stack size large enough to handle the split factor of range requests.
         let ranges_stack = Vec::with_capacity(PENDING_RANGES_LIMIT * 10);
 
+        // Generate a random sync ID that fits in a JSON object and can be used to correlate Recon sync conversations
+        // across Recon nodes.
+        let sync_id = rand::thread_rng().gen_range(1..9007199254740991);
```
dav1do, line 363: You could use the u64::MAX constant here maybe. Also you can put underscores in numerics (eg 1_000_000) to make it read like commas if it's helpful. I don't think it matters here though.

dav1do (approved):

recon/src/protocol.rs
```diff
@@ -102,8 +103,8 @@ where
 /// Message that the initiator produces
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub enum InitiatorMessage<K: Key, H: AssociativeHash> {
-    /// Declares interestes to the responder.
-    InterestRequest(Vec<RangeOpen<K>>),
+    /// Declares interests to the responder.
+    InterestRequest((u64, Vec<RangeOpen<K>>)),
```
dav1do, line 107: Should we wrap any of these values in structs? Would make adding fields/behavior easier possibly...  but new messages with v2 or whatever might be better

dav1do (approved):

PR #307, opened 2024-04-11
chore: don't hang the executor

dbcfd:
Remove uses of std::thread and ensure the p2p poller is appropriately tied to a waker.


dav1do (approved):
I put the std task back as it looks to be required (??).. but it's swarm connect so it doesn't happen often and we wait for the task.

PR #308, opened 2024-04-16
chore: start postgres for tests during CI

dav1do:
Starts a postgres docker database during CI and run tests against it. 

I planned to use script to handle that, but ended up relying on the action's `services` to start it and the code to migrate it.


dav1do:
Closing as incorporated into #302

PR #310, opened 2024-04-18
chore: recon libp2p e2e tests

dav1do:
Use `libp2p_swarm_test` to write a few tests to drive the recon protocol through the libp2p swarm. They currently show how two peers in sync have 4 events each, and when the initiator or responder fails, an event is fired marking the peer as failed.

This identified that the previous "transient" error handling behavior was a mistake and buggy, and resulted in the connection being stalled. That has been removed. A failed conversation should be restarted from the beginning on a future loop. Next up is adjusting the recon/store error handling to avoid propagating all errors up as fatal, so certain errors could be captured and retried within recon, and libp2p would never hear of them.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6636,7 +6636,6 @@ dependencies = [
  "pretty",
  "prometheus-client",
  "quickcheck",
- "rand 0.8.5",
  "regex",
  "serde",
  "serde_cbor",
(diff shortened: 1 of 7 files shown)


nathanielc (approved):
Looks great. The tests are simple and easy to read!

PR #311, opened 2024-04-18
chore: add debug image builds

nathanielc:
This change builds two images a release and debug build. The same tags for the debug build are used as the release build with the addition of a `-debug` suffix. This also changes the deploy code to deploy the debug builds to infra. We can revert that change or parameterize it later.

### DIFF
--- a/.github/workflows/cd-to-infra.yml
+++ b/.github/workflows/cd-to-infra.yml
@@ -52,10 +52,11 @@ jobs:
         else
           TAG_LATEST=false
         fi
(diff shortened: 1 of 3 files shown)


3benbox (approved):
lgtm, we're doing debug on all envs, yeah?

nathanielc:
>we're doing debug on all envs, yeah?

Yes until we can figure it out.

stbrody:
Should we do debug on QA and dev but release build on clay on prod, so we have test coverage of the build that customers are most likely to be using?

nathanielc:
@stbrody We want release builds on all envs long term. This is a temporary departure so we can find and reproduce this hang. I want debug running on all envs so we increase our chances of catching it.

stbrody:
Okay, sounds good for now.  I actually wonder though if it might be nice to keep debug builds on dev and QA long-term, might make it easier to debug any future issues that pop up.

PR #317, opened 2024-04-22
chore: add variants for error handling

dav1do:
This adds a Error enum to the store and recon crates that allows for some error handling/decision making, rather than propagating all errors up as opaque failures. This doesn't actually handle or retry errors yet, just provides the types to do so. This is targeting #320 as it was originally on the postgres branch, and it simplified conflicts/rework (and I'm hoping that branch isn't too contentious, even though the commit is large).

### DIFF
--- a/recon/src/libp2p/tests.rs
+++ b/recon/src/libp2p/tests.rs
@@ -1,18 +1,15 @@
-use std::sync::{atomic::AtomicBool, Arc};
+use crate::{
+    libp2p::{stream_set::StreamSet, PeerEvent, PeerStatus},
+    AssociativeHash, BTreeStore, Error, FullInterests, HashCount, InsertResult, InterestProvider,
+    Key, Metrics, Recon, ReconItem, Result as ReconResult, Server, Store,
+};
 
-use anyhow::{bail, Result};
 use async_trait::async_trait;
 use ceramic_core::RangeOpen;
 use ceramic_metrics::init_local_tracing;
 use libp2p::{metrics::Registry, PeerId, Swarm};
 use libp2p_swarm_test::SwarmExt;
 use tracing::info;
-
-use crate::{
-    libp2p::{stream_set::StreamSet, PeerEvent, PeerStatus},
-    AssociativeHash, BTreeStore, FullInterests, HashCount, InsertResult, InterestProvider, Key,
-    Metrics, Recon, ReconItem, Server, Store,
-};
 
 fn start_recon<K, H, S, I>(recon: Recon<K, H, S, I>) -> crate::Client<K, H>
 where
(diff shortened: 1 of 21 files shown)


recon/src/libp2p/tests.rs
```diff
@@ -28,24 +25,44 @@ where
...
 
 impl<K, H> BTreeStoreErrors<K, H> {
-    fn set_error(&self, should_error: bool) {
-        self.should_error
-            .store(should_error, std::sync::atomic::Ordering::SeqCst);
+    fn set_error(&mut self, error: Error) {
+        self.error = Some(error);
+    }
+
+    fn return_error(&self) -> Result<(), Error> {
```
dbcfd, line 39: Maybe as_error, or get_error?

store/src/metrics.rs
```diff
@@ -159,7 +158,7 @@ impl<S> ceramic_api::AccessInterestStore for StoreMetricsMiddleware<S>
 where
     S: ceramic_api::AccessInterestStore,
 {
-    async fn insert(&self, key: Interest) -> Result<bool> {
+    async fn insert(&self, key: Interest) -> anyhow::Result<bool> {
```
dbcfd, line 161: Since we're in store, why does this not return a store error?

store/src/sql/sqlite/event.rs
```diff
@@ -1,38 +1,60 @@
...
+        store.init_delivered().await?;
+        Ok(store)
+    }
+
+    /// Creates an instance that doesn't share the global event counter with other instances (only for tests to avoid running serially)
+    #[allow(dead_code)]
+    pub(crate) async fn new_local(pool: SqlitePool) -> Result<Self> {
+        let store = SqliteEventStore {
+            pool,
+            test_counter: Some(Arc::new(AtomicI64::new(0))),
```
dbcfd, line 56: Can we mock this for tests instead, by mocking the delivered function?

store/src/sql/sqlite/event.rs
```diff
@@ -44,16 +66,22 @@ impl SqliteEventStore {
...
+            .ok_or_else(|| Error::new_fatal(anyhow!("More than i64::MAX delivered events!")))?;
+        if let Some(ref t) = self.test_counter {
+            t.fetch_max(max, std::sync::atomic::Ordering::SeqCst);
+        } else {
+            GLOBAL_COUNTER.fetch_max(max, std::sync::atomic::Ordering::SeqCst);
+        }
         Ok(())
     }
 
     // could change this to rely on time similar to a snowflake ID
```
dbcfd, line 78: Yes, looking at this, it needs to be a snowflake, since we can run out of events and will continue to error at that point.

dbcfd (changes requested):

store/src/sql/sqlite/event.rs
dav1do: Yeah, this field and function should be deleted.. it was related to postgres and sharing a single DB with things running in parallel and trying to assert highwater marks. But there are definitely better ways to manage that and it felt pretty janky.

I should probably rebase all of this off of main. #320 doesn't add a ton of short term value and some things like this leaked in 😞.

store/src/metrics.rs
dav1do: This trait is from the API crate and we're just implementing it here. Could define some enum for that crate (or a generic ceramic one) but I didn't think it was necessary here since we just convert all errors to 500 right now.

store/src/sql/sqlite/event.rs
dav1do: Maybe? I thought snowflakes were still limited to u64 items (and typically less since it uses some bits for node ID). This isn't intended to be globally unique and is just a value we set in our local sqlite DB to make sure we return them in the right order. Right now, the sqlite table will break if we go over u64 (integer column). Having a time component could be useful but I'm not sure a snowflake or similar is necessary.

store/src/sql/sqlite/event.rs
dav1do: Removed this field. I'm still targeting the other PR but I adjusted it to make it a bit easier to see what was refactored.. github wasn't showing the refactor + move nicely. It's still way too big 🙈

dbcfd (approved):

PR #320, opened 2024-04-24
chore: Refactor to to lay groundwork for postgres support

dav1do:
This was broken off from #302 and the postgres details were removed. The refactor was kept, including tests that use macros to generate tests for multiple backends (just one for now). The event/interest code wasn't altered in any practical way. Pieces were pulled out into the entities and test modules and it removes the `sqlx::query!` macro usage as they only work for a single backend database. The scripts were altered slightly to address this. The sqlite pool timeout was also updated. 

ceramic-one `lib.rs` monomorphizes the run function based on the database (over the 5 store traits that are required at compile time). This makes it more clear where a new type would be added, and how to change an existing type (and illustrates that we may want to use dynamic dispatch for the store).

I added `put` the bitswap store trait as we were going through the concrete type previously. It exposes a `database-url` flag (or `CERAMIC_ONE_DATABASE_URL` env var) to configure the sqlite location but still uses the same default.


dav1do:
The diff makes it look like they are entirely new files, even though they were committed as as a rename.. but anyway, you can look at [this commit](https://github.com/ceramicnetwork/rust-ceramic/pull/320/commits/d9abd9d725fad4d8420eccafb2d633570b8a01b1) to see refactor changes side by side.

store/src/sql/event.rs
```diff
@@ -332,63 +226,51 @@ impl SqliteEventStore {
...
-        sqlx::query!(
-            "INSERT INTO ceramic_one_event_block (event_cid, idx, root, block_multihash, codec) VALUES ($1, $2, $3, $4, $5) on conflict do nothing;",
-            id,
-            idx,
-            root,
-            multihash,
-            code,
-        )
+        sqlx::query(
+            "INSERT INTO ceramic_one_event_block (event_cid, idx, root, block_multihash, codec) VALUES ($1, $2, $3, $4, $5) on conflict do nothing;")
```
nathanielc, line 230: should this query be moved into the entities section like the others?

store/src/sql/event.rs
```diff
@@ -332,63 +226,51 @@ impl SqliteEventStore {
...
         let id = key.as_bytes();
         let delivered = self.get_delivered();
-        sqlx::query!(
-            "UPDATE ceramic_one_event SET delivered = $1 WHERE order_key = $2;",
-            delivered,
-            id
-        )
-        .execute(&mut **conn)
-        .await?;
+        sqlx::query("UPDATE ceramic_one_event SET delivered = $1 WHERE order_key = $2;")
```
nathanielc, line 244: same question here.

nathanielc (approved):
I really like this direction. Some questions below but nothing that needs a another review

store/src/sql/event.rs
dav1do: Good call. Crated an `EventBlockQuery` struct for it

store/src/sql/event.rs
dav1do: yep, good catch. moved into `EventQuery`

PR #324, opened 2024-04-29
fix: changes to improve sync and ingest performance

dav1do:
# Summary

I made a few changes to adjust how we insert data and how we flush recon messages (i.e. send them faster). 

- Flush the recon message sink more aggressively. This sends messages over the wire faster. We still "feed" messages while we have them in memory, but as soon as we're going to yield, we send them all rather than wait until we have some number ready.
- Parse the event data carfile outside of the database transaction lock. Clearly, reducing the amount of time we hold the critical resource is a good thing, especially as we have only one of these using sqlite.
- I added an api server task to batch inserts, this isn't really important under 300 tps, but in my benchmarks it helped keep things stable in the 1000-1500 tps range, and generally kept lock contention down quite a bit (visible in the order of magnitude decrase in the `api_insert_many` query duration - previously called `model_insert`). This didn't necessarily cause sync to speed up dramatically, as we're bounded on network, but it does seem like a worthwhile change and should be almost unnoticeable under lower loads.
- I renamed some metrics for clarity while we aren't yet live and fixed the local store insert benchmark to run again. Unfortunately, these changes didn't make much impact to the benchmark, which particularly surprised me with the carfile stuff, but it's possible the way they're set up doesn't cause the same lock contention as the api writes/recon sync and is simply bound by another resource.

### Flushing change

The flushing change helps with this graph. It took longer to exchange keys and we were seeing a decline in the graph without any obvious resource contention (this shows 3 peers syncing and how the messages exchanged was declining). 

image

After changing that, the line stays more steady (notice green and yellow - the blue and orange were another metric I added to help me understand this in the first place and no longer exists)
image

### 3 hour benchmark locally

It's worth calling out that the benchmark scenario is a pathological bad case for recon. We're inserting new streams for a single model/controller, and as we keep the peer in sync, each additional negotiation requires more work to identify the interleaved differences. There are meta things (like gratuitous push, adding a time component, and more) that we could do to address things like this, but for now it's enough to observe and write alternative scenarios to demonstrate different outcomes (things like api burstiness and writing to multiple models would likely be good for the recon sync rate).

With that in mind, the high (and stable) sync rate at the end of existing scenarios is simply the 0 case, and we're now seeing two peers that start more in sync, and therefore require a bit more work to identify the differences. 

I'm not certain what happened at 23:30, and it's worth running more tests to see if it shows up again (it might have been something as silly as my pc trying to sleep before I adjusted power settings).

image

#### Without API batching

The throughput is similar, but notice the 0.01 insert times (left middle)
image

#### With batching
Notice we're in the 0.004 range now (right middle: almost an order of magnitude better), even though the overall throughput/system doesn't look a ton different.

image

Results from a 30 minute test
image

### DIFF
--- a/api/src/server.rs
+++ b/api/src/server.rs
@@ -53,6 +53,12 @@ const EVENT_INSERT_QUEUE_SIZE: usize = 3;
 /// This is quite low, but in my benchmarking adding a longer interval just slowed ingest down, without changing contention noticeably.
 const FLUSH_INTERVAL_MS: u64 = 10;
 
+/// How long are we willing to wait to enqueue an insert to the database service loop before we tell the call it was full.
+const INSERT_ENQUEUE_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(1);
+/// How long are we willing to wait for the database service to respond to an insert request before we tell the caller it was too slow.
+/// Aborting and returning an error doesn't mean that the write won't be processed, only that the caller will get an error indicating it timed out.
+const INSERT_REQUEST_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
+
 // Helper to build responses consistent as we can't implement for the api_server::models directly
 pub struct BuildResponse {}
 impl BuildResponse {
@@ -391,7 +397,7 @@ where
         let event_data = decode_event_data(&event.data)?;
         let (tx, rx) = tokio::sync::oneshot::channel();
         tokio::time::timeout(
-            std::time::Duration::from_secs(3),
+            INSERT_ENQUEUE_TIMEOUT,
             self.insert_task.tx.send(EventInsert {
                 id: event_id,
                 data: event_data,
@@ -404,7 +410,7 @@ where
         .await?
         .map_err(|_| ErrorResponse::new("Database service not available".to_owned()))?;
 
-        let _new = tokio::time::timeout(std::time::Duration::from_secs(10), rx)
+        let _new = tokio::time::timeout(INSERT_REQUEST_TIMEOUT, rx)
             .await
             .map_err(|_| {
                 ErrorResponse::new("Timeout waiting for database service response".to_owned())
--- a/api/src/server.rs
+++ b/api/src/server.rs
@@ -281,15 +281,16 @@ where
             // i.e. validate before sending here and this is just an insert, we may want to process more at once.
             loop {
                 tokio::select! {
-                    _ = async {}, if events.len() >= EVENT_INSERT_QUEUE_SIZE => {
-                        Self::process_events(&mut events, &event_store).await;
-                    }
                     _ = interval.tick(), if !events.is_empty() => {
                         Self::process_events(&mut events, &event_store).await;
                     }
                     Some(req) = event_rx.recv() => {
                         events.push(req);
                     }
+                }
+                // make sure the events queue doesn't get too deep when we're under heavy load
+                if events.len() >= EVENT_INSERT_QUEUE_SIZE {
+                    Self::process_events(&mut events, &event_store).await;
                 }
             }
         })
--- a/api/src/server.rs
+++ b/api/src/server.rs
@@ -2,6 +2,7 @@
 
 #![allow(unused_imports)]
 
+use std::time::Duration;
 use std::{future::Future, ops::Range};
 use std::{marker::PhantomData, ops::RangeBounds};
 use std::{net::SocketAddr, ops::Bound};
@@ -43,6 +44,14 @@ use swagger::ApiError;
 use tikv_jemalloc_ctl::{epoch, stats};
 
 use crate::ResumeToken;
+
+/// When the incoming events queue has at least this many items, we'll store them.
+/// This imples when we're getting writes faster than the flush interval.
+const EVENT_INSERT_QUEUE_SIZE: usize = 3;
+/// How often we should flush the queue of events to the store. This applies when we have fewer than `EVENT_INSERT_QUEUE_SIZE` events,
+/// in order to avoid stalling a single write from being processed for too long, while still reducing contention when we have a lot of writes.
+/// This is quite low, but in my benchmarking adding a longer interval just slowed ingest down, without changing contention noticeably.
+const FLUSH_INTERVAL_MS: u64 = 10;
 
 // Helper to build responses consistent as we can't implement for the api_server::models directly
 pub struct BuildResponse {}
@@ -149,7 +158,8 @@ impl<S: AccessInterestStore> AccessInterestStore for Arc<S> {
 #[async_trait]
 pub trait AccessModelStore: Send + Sync {
     /// Returns (new_key, new_value) where true if was newly inserted, false if it already existed.
-    async fn insert(&self, key: EventId, value: Option<Vec<u8>>) -> Result<(bool, bool)>;
+    async fn insert_many(&self, items: &[(EventId, Option<Vec<u8>>)])
+        -> Result<(Vec<bool>, usize)>;
     async fn range_with_values(
         &self,
         start: &EventId,
@@ -171,8 +181,11 @@ pub trait AccessModelStore: Send + Sync {
 
 #[async_trait::async_trait]
 impl<S: AccessModelStore> AccessModelStore for Arc<S> {
-    async fn insert(&self, key: EventId, value: Option<Vec<u8>>) -> Result<(bool, bool)> {
-        self.as_ref().insert(key, value).await
+    async fn insert_many(
+        &self,
+        items: &[(EventId, Option<Vec<u8>>)],
+    ) -> Result<(Vec<bool>, usize)> {
+        self.as_ref().insert_many(items).await
     }
 
     async fn range_with_values(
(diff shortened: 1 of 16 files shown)


dav1do:
I was going to add a task to handle flushing and clearing the server insert task, but it's a bit annoying as the server gets moved and then erased into `ceramic_api_server::context::MakeAddContext<swagger::auth::MakeAllowAllAuthenticator<ceramic_api_server::server::MakeService<ceramic_api::MetricsMiddleware<ceramic_api::Server<_, ceramic_store::StoreMetricsMiddleware<std::sync::Arc<I1>>, ceramic_store::StoreMetricsMiddleware<std::sync::Arc<E1>>>>, _>, _>, swagger::EmptyContext>` so we can't easily call `server.shutdown()` after the graceful handler is triggered. 

There are ways to solve it, but I realized we don't really care.. if graceful shutdown is called, we'll empty what we can and then drop all the in flight events, which will return errors to all the clients and they will know their write wasn't successful. 

Sample of how I had implemented it

```rust

// This can be expanded to an enum of database writes, but for now we just have one.
struct InsertTask {
    handle: tokio::task::JoinHandle<()>,
    tx: tokio::sync::mpsc::Sender<EventInsert>,
    // only an option so we can take ownership to shutdown without taking ownership the entire server struct
    shutdown_tx: Mutex<Option<tokio::sync::oneshot::Sender<()>>>,
}

#[derive(Clone)]
pub struct Server<C, I, M> {
    peer_id: PeerId,
    network: Network,
    interest: I,
    model: Arc<M>,
    // If we need to restart this ever, we'll need a mutex. For now we want to avoid locking the channel
    // so we just keep track to gracefully shutdown, but if the task dies, the server is in a fatal error state.
    insert_task: Arc<InsertTask>,
    marker: PhantomData<C>,
}

impl<C, I, M> Server<C, I, M>
where
    I: AccessInterestStore,
    M: AccessModelStore + 'static,
{
    pub fn new(peer_id: PeerId, network: Network, interest: I, model: Arc<M>) -> Self {
        let (tx, event_rx) = tokio::sync::mpsc::channel::<EventInsert>(1024);
        let (shutdown_tx, shutdown_rx) = tokio::sync::oneshot::channel();
        let event_store = model.clone();

        let handle = Self::start_insert_task(event_store, event_rx, shutdown_rx);
        let insert_task = Arc::new(InsertTask {
            handle,
            tx,
            shutdown_tx: Mutex::new(Some(shutdown_tx)),
        });
        Server {
            peer_id,
            network,
            interest,
            model,
            insert_task,
            marker: PhantomData,
        }
    }

    fn start_insert_task(
        event_store: Arc<M>,
        mut event_rx: tokio::sync::mpsc::Receiver<EventInsert>,
        mut shutdown_rx: tokio::sync::oneshot::Receiver<()>,
    ) -> tokio::task::JoinHandle<()> {
        tokio::spawn(async move {
            let mut interval = tokio::time::interval(Duration::from_millis(FLUSH_INTERVAL_MS));
            let mut events = vec![];
            // could bias towards processing the queue of events over accepting more, but we'll
            // rely on the channel depth for backpressure. the goal is to keep the queue close to empty
            // without processing one at a time. when we stop parsing the carfile in the store
            // i.e. validate before sending here and this is just an insert, we may want to process more at once.
            loop {
                tokio::select! {
                    _ = async {}, if events.len() >= EVENT_INSERT_QUEUE_SIZE => {
                        Self::process_events(&mut events, &event_store).await;
                    }
                    _ = interval.tick(), if !events.is_empty() => {
                        Self::process_events(&mut events, &event_store).await;
                    }
                    _ = &mut shutdown_rx => {
                        event_rx.close();
                        while let Some(req) = event_rx.recv().await {
                            events.push(req);
                        }
                        if !events.is_empty() {
                            Self::process_events(&mut events, &event_store).await;
                        }
                        break;
                    }
                    Some(req) = event_rx.recv() => {
                        events.push(req);
                    }
                }
            }
        })
    }

    pub async fn shutdown(&self) -> Result<()> {
        let shutdown = self.insert_task.shutdown_tx.lock().unwrap().take();
        if let Some(tx) = shutdown {
            let _ = tx.send(());
        }
        let shutdown_duration = std::time::Duration::from_secs(5);
        let now = std::time::Instant::now();
        loop {
            if self.insert_task.handle.is_finished() {
                break;
            } else if now.elapsed() > shutdown_duration {
                tracing::warn!("Failed to shutdown insert task in time. Forcing shutdown.");
                self.insert_task.handle.abort();
                break;
            } else {
                tokio::time::sleep(Duration::from_millis(100)).await;
            }
        }
        Ok(())
    }
```

recon/src/protocol.rs
```diff
@@ -840,19 +835,18 @@ where
         for key in keys {
             if let Some(value) = self.recon.value_for_key(key.clone()).await? {
                 self.stream
-                    .feed((self.value_resp_fn)(
+                    .send((self.value_resp_fn)(
```
dav1do, line 838: use `send_all` here instead?

recon/src/protocol.rs
```diff
@@ -931,33 +925,12 @@ impl<S> SinkFlusher<S> {
         self.feed_count = 0;
         Ok(())
     }
-    async fn feed<T, E>(&mut self, message: T) -> Result<()>
```
dav1do, line 934: Is it possible that the additional sending is causing the additional sawing in the graph? I assumed that things are syncing faster, so we have to understand more ranges, but I'm wondering if there's somewhere that sending more/less together might help get bigger overlaps. 

It's also possible that we just didn't have tests demonstrating this time/number of keys synced, so we just eventually got to really fast 0 cases.

dav1do: I deleted this to simplify the api and avoid having to think about feed/flush.. but now you just have to collect in memory to send_all so it doesn't seem like an improvement.

recon/src/protocol.rs
dav1do: sticking with `send` until (if) we change the retrieval logic to work on many keys.. since we're yielding to find another key, we should send what we have

store/src/sql/entities/event.rs
```diff
@@ -87,20 +88,129 @@ impl ReconHash {
     }
 }
 
-#[derive(Debug, Clone, sqlx::FromRow)]
-pub struct EventValueRaw {
+use anyhow::anyhow;
+use multihash::{Code, MultihashDigest};
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, sqlx::Type)]
+pub struct BlockHash(Multihash);
```
dav1do, line 95: @stbrody we could do something similar for CID if we wanted to impl sqlx::Type to indicate it should be blob and how to use it in both `.bind()` and `query_as` FromRow responses.

api/src/server.rs
```diff
@@ -298,10 +389,28 @@ where
...
+                tx,
+            }),
+        )
+        .map_err(|_| {
+            ErrorResponse::new("Database service queue is too full to accept requests".to_owned())
+        })
+        .await?
+        .map_err(|_| ErrorResponse::new("Database service not available".to_owned()))?;
+
+        let _new = tokio::time::timeout(std::time::Duration::from_secs(10), rx)
```
nathanielc, line 407: should we make this a constant as well?

api/src/server.rs
```diff
@@ -298,10 +389,28 @@ where
     pub async fn post_events(&self, event: Event) -> Result<EventsPostResponse, ErrorResponse> {
         let event_id = decode_event_id(&event.id)?;
         let event_data = decode_event_data(&event.data)?;
-        self.model
-            .insert(event_id, Some(event_data))
+        let (tx, rx) = tokio::sync::oneshot::channel();
+        tokio::time::timeout(
+            std::time::Duration::from_secs(3),
```
nathanielc, line 394: Should we make this a constant? Why 3s here and 10s below?

api/src/tests.rs
```diff
@@ -96,6 +101,7 @@ impl AccessModelStore for MockReconModelTest {
 #[tokio::test]
 #[traced_test]
 async fn create_event() {
+    let _ = init_local_tracing();
```
nathanielc, line 104: Why do we need this? I thought `traced_test` would init the tracing bits?

nathanielc (approved):
LGTM I have a few questions but they should not block merging.

api/src/server.rs
dav1do: Good call, done.

api/src/server.rs
dav1do: Yeah, good call on constants. I changed this to 1 second and left the other at 10. The names hopefully help clarify a bit, but I was thinking this should be short because enqueuing to the channel should be immediate, and if it's full we shouldn't wait that long and just fail the request. Once we accept the write, we are willing to wait longer to hear it request succeeded... In practice, 10s is pretty high, but I was trying to avoid the "I got an error but my write succeeded" case for js-ceramic without leaving an infinite timeout... I'm open to alternate values.

api/src/tests.rs
dav1do: We probably don't.. I'm not really a fan of `traced_test` so far (or maybe I haven't read the docs). I think I got confused because I wasn't seeing anything because I forgot the `--nocapture` flag, and the `init_local_tracing` just works on RUST_LOG directly without needing that (and it prints nice formatted output).

dav1do: I removed traced_test from the api crate.. happy to go the other way if you have a strong preference.

nathanielc: I'd prefer `traced_test` as it should work even without `--nocapture` and uses RUST_LOG directly as well. We use it all over the code base so I'd prefer to keep using it. Can you reproduce the issue you had with it?

api/src/server.rs
```diff
@@ -202,30 +221,108 @@ impl<S: AccessModelStore> AccessModelStore for Arc<S> {
...
+        tokio::spawn(async move {
+            let mut interval = tokio::time::interval(Duration::from_millis(FLUSH_INTERVAL_MS));
+            let mut events = vec![];
+            // could bias towards processing the queue of events over accepting more, but we'll
+            // rely on the channel depth for backpressure. the goal is to keep the queue close to empty
+            // without processing one at a time. when we stop parsing the carfile in the store
+            // i.e. validate before sending here and this is just an insert, we may want to process more at once.
+            loop {
+                tokio::select! {
+                    _ = async {}, if events.len() >= EVENT_INSERT_QUEUE_SIZE => {
```
dbcfd, line 284: Can we just eliminate this match, and check after the select?

e.g. `select { tick, event }; if events.len() > process_events`

dav1do: lol yeah, good call. not sure why I didn't do that initially. Adding the `async {} if condition` felt... wrong.

api/src/tests.rs
dav1do: I put `traced_test` back for now

PR #326, opened 2024-04-29
feat: adds migration logic from IPFS 

nathanielc:


api/ceramic.yaml
```diff
@@ -101,41 +101,6 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorResponse"
-  /events/{event_id}:
```
stbrody, line 104: why removing this endpoint?  It's used in some of our test currently so removing it will break those tests

stbrody (changes requested):
this PR is really large and seems to contain a bunch of separate pieces and refactoring that could easily stand on their own separate from the migration command (such as moving where the logic to parse events into CARs lives, and adding logic for serializing events into CARs).  Can we split this up into some smaller PRs so it's easier to review?

nathanielc:
@stbrody Good call, I'll split this up. It grew gradually and I didn't realize how large it had gotten.

nathanielc:
We now have a sequence of PRs

https://github.com/ceramicnetwork/rust-ceramic/pull/378
https://github.com/ceramicnetwork/rust-ceramic/pull/379
https://github.com/ceramicnetwork/rust-ceramic/pull/380

Once they all merge this should be good for review. Its still large but now all refactors are separate PRs.

one/src/lib.rs
```diff
@@ -47,9 +43,9 @@ struct Cli {
 enum Command {
     /// Run a daemon process
     Daemon(DaemonOpts),
-    /// Event store tools
+    /// Perform various migrations
     #[command(subcommand)]
-    Events(events::EventsCommand),
+    Migrations(migrations::EventsCommand),
```
stbrody, line 48: is this going to get confusing with the database migrations that @dav1do is working on right now?

one/src/migrations.rs
```diff
@@ -0,0 +1,157 @@
...
+#[derive(Args, Debug)]
+pub struct FromIpfsOpts {
+    /// The path to the ipfs_repo [eg: ~/.ipfs/blocks]
+    #[clap(long, short, value_parser)]
+    input_ipfs_path: PathBuf,
+
+    /// The database to connect to e.g.
+    ///
+    /// `sqlite:///path/to/file.sqlite3` or `sqlite://:memory:` or `sqlite://~/.ceramic-one/db.sqlite3`
+    #[clap(long, short, default_value = "sqlite://~/.ceramic-one/db.sqlite3")]
```
stbrody, line 32: should we make a constant to ensure that the default used here is the same as the default used for the daemon?

one/src/lib.rs
```diff
@@ -150,6 +142,16 @@ struct DaemonOpts {
...
 
+    #[command(flatten)]
+    db_opts: DBOpts,
+}
+
+#[derive(Args, Debug)]
+struct DBOpts {
+    /// Path to storage directory
+    #[arg(short, long, env = "CERAMIC_ONE_STORE_DIR")]
+    store_dir: Option<PathBuf>,
```
stbrody, line 153: doesn't this have a default value?  I don't think it need to be specified currently for the daemon to start up - why isn't the default specified here?

one/src/migrations.rs
```diff
@@ -0,0 +1,157 @@
...
+    output_database_url: String,
+
+    /// Unique key used to find other Ceramic peers via the DHT
+    #[arg(long, default_value = "testnet-clay", env = "CERAMIC_ONE_NETWORK")]
```
stbrody, line 36: not for this PR, but as a general question while I'm thinking about it: we've talked about changing the default network for a while to something other than Clay.  Most likely to an isolated local node.  Is that a change we want to make before GA?

one/src/migrations.rs
```diff
@@ -0,0 +1,157 @@
...
+            database_url: Some(value.output_database_url.clone()),
+        }
+    }
+}
+
+pub async fn events(cmd: EventsCommand) -> Result<()> {
+    if let Err(e) = init_local_tracing() {
+        eprintln!("Failed to initialize tracing: {}", e);
+    }
+    // TODO Cleanup
```
stbrody, line 57: either remove or add more detail and link to a ticket?

one/src/migrations.rs
```diff
@@ -0,0 +1,157 @@
...
+        return Err(anyhow!(
+            "block data does not match hash: path={}",
+            block_path.display()
+        ));
+    }
+    // If we can decode the block as a JWS envelope then we can assume the block is dag-jose
+    // encoded.
+    let result: Result<unvalidated::signed::Envelope, _> = serde_ipld_dagcbor::from_slice(&blob);
+    let cid = if result.is_ok() {
+        Cid::new_v1(0x85, hash)
```
stbrody, line 136: can we name these constants so it's clear to a reader what they represent?

stbrody (changes requested):

one/src/lib.rs
nathanielc: No, I expect he will plug into this top level command. The intent is that operators can expect to find all types of migration commands in one place regardless of if they are within the same db or across dbs.

one/src/lib.rs
nathanielc: The default is system dependent as its the user's home directory which changes based on the user and the OS. Also note this is not a change in behavior but a refactor to enable the migrations cmd to share this default logic.

service/src/event/migration.rs
```diff
@@ -0,0 +1,326 @@
...
+}
+
+impl Migrator {
+    pub async fn new(
+        network: Network,
+        blocks: impl Stream<Item = Result<BoxedBlock>>,
+    ) -> Result<Self> {
+        let blocks = blocks
+            .map(|block| block.map(|block| (block.cid(), block)))
+            .try_collect::<BTreeMap<Cid, BoxedBlock>>()
```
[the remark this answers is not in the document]
nathanielc: Blocks are lazy loaded. This logic reads only metadata (i.e. cid and file path) for each block. We should easily be able to fit metadata for even our largest dbs in memory.

one/src/migrations.rs
nathanielc: That came over from previous code and since I do not know what it means I'll just delete it.

one/src/migrations.rs
nathanielc: This is unfortunately not the same concept. The daemon has a combination of the store path and a db url that define where to find the db. The migration cmd doesn't have a concept of a store path. So it has a sane default.

We should figure how to reconcile these differences as they are confusing.

nathanielc: @dav1do Any ideas on how to simplify this?

dav1do: We could probably remove the `--database-url` and rely on the store directory exclusively as we have no plans to support postgres or alternative databases. It did use it for some testing yesterday because I had custom database names, but we probably don't want to allow people to rename the file at this point. And it's kind of bugged because if you pass a database-url it might not be in your store-dir at all, and then it's just weird (you could even pass `sqlite//:memory:` if you were so inclined).

For now, we should probably just no-op it and print a warning. I was just noodling about removing the store crate from ceramic-one and going through the service crate for everything (like give me a service using sqlite or something). 

I'm not sure how/if we want to specify database files/types, but I think we should probably say the file name/format is up to us and you can only provide a directory. Not sure how likely it is, but I could imagine creating multiple sqlite files for like every ~100gb of events or something. Probably not how we'd want to shard things, but the point is mostly that it's not really an option if we let people pick the sqlite file name and expect only one.

dav1do: Maybe I kind of missed the point in that response 😅 but the daemon has  database-url to specify and it will not be modified or moved from that location, or it will be called `db.sqlite3` in one of these 3 places:

1. path from options
2. path $HOME/.ceramic-one
3. pwd/.ceramic-one

IMO specifying is something we don't want (even though I'm the one that added it 🤦).

nathanielc: https://github.com/ceramicnetwork/rust-ceramic/pull/384

one/src/migrations.rs
```diff
@@ -0,0 +1,155 @@
...
+#[derive(Subcommand, Debug)]
+pub enum EventsCommand {
+    /// Migrate raw event blocks from IPFS.
+    FromIpfs(FromIpfsOpts),
+}
+
+#[derive(Args, Debug)]
+pub struct FromIpfsOpts {
+    /// The path to the ipfs_repo [eg: ~/.ipfs/blocks]
+    #[clap(long, short, value_parser, env = "CERAMIC_ONE_INPPUT_IPFS_PATH")]
```
smrz2001, line 26:
```suggestion
    #[clap(long, short, value_parser, env = "CERAMIC_ONE_INPUT_IPFS_PATH")]
```

service/src/tests/migration.rs
```diff
@@ -0,0 +1,309 @@
...
+            .stream()
+            .map_err(|err| anyhow!("{err}"))
+            .map(|block| block.map(|(cid, data)| Box::new(InMemBlock { cid, data }) as BoxedBlock));
+        if let Some(prev_stream) = stream.take() {
+            stream = Some(prev_stream.chain(s).boxed());
+        } else {
+            stream = Some(s.boxed());
+        }
+    }
+    //TODO return empty stream
```
dav1do, line 46: What's the reason to do this? Seems like it's working for the tests now?

dav1do (commented):
I think this looks really good. A couple questions/comments but nothing substantial. I'm going to leave it unapproved since others have been invested/reviewing and are more familiar with the event structure than I am, but happy to hit the green button tomorrow if needed 🚀.

service/src/tests/migration.rs
nathanielc: Yeah, this was more for completeness than anything. But no tests need this code path so I'll remove the todo so its clear its not broken.

stbrody (approved):
lgtm

service/src/event/migration.rs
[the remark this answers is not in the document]
nathanielc: Added some explicit comments to the Block trait methods.

dav1do (approved):

PR #327, opened 2024-04-30
feat: Update events GET API to return event CID instead of EventID

stbrody:


stbrody:
This is passing unit tests, but I still want to do some manual testing to see that the http endpoint actually works correctly end-to-end.  The only unit test that needed updating was using a mock AccessModelStore.  Do we have any tests for the http API that actually hit a real server and would actually exercise the real queries to the SQL database?

Anyway sending this out for review early even as I keep testing it, just to get early feedback that I'm going in the right direction.

store/src/sql/entities/query.rs
```diff
@@ -34,8 +36,21 @@ impl EventQuery {
             ORDER BY eb.idx;"#
     }
 
+    /// Requires binding 1 parameter. Finds the `EventValueRaw` values needed to rebuild the event
+    /// Looks up the event by the root CID of the event.
+    /// TODO(stbrody): Does it need to return order_key at all?  Can the join on event table be removed?
```
dav1do, line 41: It needs to return whatever you want it to be 😄.. right now it's deserialized into this struct so it does need order_key, but if we can remove the need for event ID that can maybe be dropped (not sure since recon relies so heavily on it)

```
#[derive(Debug, Clone, sqlx::FromRow)]
pub struct EventValueRaw {
    pub order_key: Vec<u8>,
    pub codec: i64,
    pub root: bool,
    pub idx: i32,
    pub multihash: Vec<u8>,
    pub bytes: Vec<u8>,
}```

File: store/src/sql/entities/query.rs [hunk 2]
```diff
@@ -24,7 +24,9 @@ pub struct EventQuery;
 
 impl EventQuery {
     /// Requires binding 1 parameter. Finds the `EventValueRaw` values needed to rebuild the event
-    pub fn value_blocks_one() -> &'static str {
+    /// Looks up the event by the EventID (ie order_key).
+    /// TODO(stbrody): fields selected don't match how it's used?
```
dav1do, line 28: The block stuff can probably be consolidated a bit as the BlockRow/EventValueRaw stuff overlaps. I adjusted it up somewhat in #324 (not sure I can say "cleaned it up"), but I'm sort of waiting for the CeramicStore/CeramicService stuff and expect the sql layer to get a new set of traits. I think the oddness stems from serving as both an ipfs block store and a carfile store and not dividing the queries clearly yet.

This is deserialized into BlockRow which uses [this](https://github.com/ceramicnetwork/rust-ceramic/blob/dadeea0366b20b8f5babfca58ed68e055bf8fbda/store/src/sql/entities/block.rs#L15) FromRow (order key is not needed in that one but was when it was being used to build the carfile as well)

Regarding the CeramicStore, here's what I'm thinking (at least for event store) but I'm open to other approaches. 

recon::Store is implemented by CeramicStore. CeramicStore operates on bytes and returns `CeramicEvents/Commits/Validated/etc` that the redtree/sqlite can accept and persist. Then we aren't parsing bytes -> carfile -> bytes -> carfile repeatedly and we can divide up the block store/event store structs/queries a bit more cleanly. That might mean the `impl recon::Store` for sqlite/redtree etc change to some EventStore trait or something that takes a real ceramic event and not just bytes, and the only valid `recon::Store<Key=EventId, Hash=Sha256a>` option would be the Ceramic event version.

dav1do (commented):
Nice, a couple comments based on your TODO questions.

In terms of tests, we have the correctness tests run on PR which does hit the database. For unit tests, the api tests only use the mock interface. We could crate [integration tests](https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html) in the project but the API crate can't depend on store (store depends on API to implement the trait), so you can't do it in the api tests as they stand.

You can test the store functions in the store crate with a real (in memory) sqlite database. If they pass and meet the interface, and the api tests pass with the interface, we've considered it sufficient for unit testing so far.

api/src/server.rs
```diff
@@ -324,17 +347,17 @@ where
 
     pub async fn get_events_event_id(
         &self,
-        event_id: String,
+        event_cid: String,
     ) -> Result<EventsEventIdGetResponse, ErrorResponse> {
-        let decoded_event_id = decode_event_id(&event_id)?;
-        match self.model.value_for_key(&decoded_event_id).await {
+        let decoded_event_cid = decode_cid(&event_cid)?;
```
dav1do, line 352: we could update this to return `Ok(EventsEventIdGetResponse::BadResponse("Invalid CID"))` rather than a 500 .. but it's not doing it yet so it's fine to leave too

dav1do: oh, I have an idea for backward compatibility too.. we could try as a CID and if fails, as an EventID. This would keep the tests passing and give us a "debug" route if we have recon keys.

I was going to suggest deleting the `value_for_order_key` trait behavior, since we're not exposing an API for that.. but if we have it, supporting both would be nice (at least for a while).

store/src/sql/entities/query.rs
stbrody: So this BlockRow serialization is what confused me.  The FromRow you linked seems to be looking for `multihash` and `codec` fields and parsing those into a Cid, but the table definition for the events table just has a field called `cid` that is a Blob.

stbrody: oh wait, I think maybe I see?  The `multihash` and `codec` columns are coming from the `event_block` table, not the `event` table.  But the event_block table also has an `event_cid` column.  What's the relationship between the `event_cid` column and the `multihash` and `codec` columns in the `event_block` table?  Are they redundant of each other?

store/src/sql/entities/query.rs
stbrody: Where is it being used to populate an `EventValueRaw`?  I only see it being turned into a `BlockRow`

store/src/sql/entities/query.rs
dav1do: So the `FromRow` stuff is query specific depending on the return columns from the statement. Typically you use  derive and then the return columns match the struct fields, but you can write a custom one and you get the query results and can pull out each response value by name, decode it somehow, and then build the struct.

That's how the `BlockRow` and `EventValueRaw` both use the same query but build different types.. not the cleanest way to do it and it will likely be refactored. The `query_as()` function automatically calls FromRow on the response from the query for the type you hint based on whether it's `fetch_one`/`fetch_optional`/`fetch_all` for `T`/`Option<T>`/`Vec<T>`.

Here's an example using the derive vs a tuple of types that all have a valid Decode operation. Since CID doesn't impl Decode from sqlite blob, I used FromRow in our code to get the bytes and then build a CID to put into a struct. This just uses byte arrays:

```rust
    // assuming each column has a Decode impl for the database. 
    // since we're going from BLOB -> Vec<u8>, INTEGER -> i64 and INTEGER -> u32 it's fine for sqlite 
    // for postgres, this should fail to compile as u32 doesn't exist in pg so no decode impl would be found.
    // it would also fail in sqlite if the data in the column is negative or overflows
    let event_block: (Vec<u8>, i64, u32) =
        sqlx::query_as("SELECT event_cid, codec, idx FROM ceramic_one_event_block ORDER limit 1")
            .fetch_one(&mut conn)
            .await?;
    // or create a type
    #[derive(sqlx::FromRow)]
    // each name must match the return in the query. 
    struct TempBlock {
        event_cid: Vec<u8>,
        codec: i64,
        idx: u32,
    }

    let event_blocks: Vec<TempBlock> =
    sqlx::query_as("SELECT event_cid, codec, idx FROM ceramic_one_event_block ORDER limit 10")
        .fetch_all(&mut conn)
        .await?;
```

> oh wait, I think maybe I see? The multihash and codec columns are coming from the event_block table, not the event table. But the event_block table also has an event_cid column. What's the relationship between the event_cid column and the multihash and codec columns in the event_block table? Are they redundant of each other?

the `block` table stores the blocks, the `event` table stores the data in the "recon hashable" way, and the `event_block` table is the join that tells you which blocks go in the carfile and in which order for the event. I had blocks have CID instead of multihash originally, but Aaron suggested using the multihash instead and the event_block should contain the extra CID info (unless I have this wrong now). I believe the `event_cid` should be the first root of the carfile, and we just pull it out to index it so we can join things back up, and when we stop trusting the value from the API and build it all from the value, it will be more obvious where things come from and what's duplicative/indexed.

`event 1->* event_block *<-1 block`  `1->*` is supposed to mean one-to-many 😅

store/src/sql/entities/query.rs
dav1do: Oh, I think you're right.. so they are different queries but it's these two returning basically the same columns.

```rust
let blocks: Vec<BlockRow> = sqlx::query_as(EventQuery::value_blocks_one()) // value_blocks_by_order_key_one
let all_blocks: Vec<EventBlockRaw> = sqlx::query_as(EventQuery::value_blocks_many()) // value_blocks_by_cid_one
```

store/src/sql/entities/query.rs
```diff
@@ -34,8 +36,21 @@ impl EventQuery {
             ORDER BY eb.idx;"#
     }
 
-    /// Requires binding 4 parameters. Finds the `EventBlockRaw` values needed to rebuild the event
-    pub fn value_blocks_many() -> &'static str {
+    /// Requires binding 1 parameter. Finds the `EventValueRaw` values needed to rebuild the event
+    /// Looks up the event by the root CID of the event.
+    /// TODO(stbrody): No need to return order_key?
```
dav1do, line 41: yeah, it looks like you're right. you only use this to get a BlockRow so it should order_key shouldn't be necessary.

dav1do (approved):
Nice, looks good. Sorry if I was confusing you (and myself) about the query stuff. It does look like you had it right. :shipit:

stbrody:
PR has been approved and included in the unified branch: https://github.com/ceramicnetwork/rust-ceramic/pull/335

PR #328, opened 2024-04-30
feat: add serde methods for Events

nathanielc:
The unvalidated module contains structures and can serialize and deserialze into Ceramic Events (i.e. dagcbor IPLD blocks).

Additionally a method to construct an EventId from the raw data is added.


event/src/unvalidated.rs
```diff
@@ -0,0 +1,242 @@
...
+    }
+    /// Get the data
+    pub fn data(&self) -> &Ipld {
+        &self.data
+    }
+}
+
+/// Headers for a data event
+#[derive(Debug, Serialize, Deserialize)]
+pub struct DataHeader {
```
dbcfd, line 138: Feels like we're duplicating what is in https://github.com/ceramicnetwork/rust-ceramic/blob/main/event/src/args.rs#L88

event/src/unvalidated.rs
```diff
@@ -0,0 +1,242 @@
...
+    }
+}
+
+/// Payload of a data event
+#[derive(Debug, Serialize, Deserialize)]
+pub struct DataPayload {
+    id: Cid,
+    prev: Cid,
+    header: Option<DataHeader>,
+    data: Ipld,
```
dbcfd, line 114: This seems like a specialization of https://github.com/ceramicnetwork/rust-ceramic/blob/main/event/src/args.rs#L99

dbcfd (changes requested):

event/src/unvalidated.rs
nathanielc: @dbcfd Good call, I'll make both bits of the code use the same structures. As I see it the `args` module can consume the `unvalidated` module to construct the events. This way the structures still live in the `unvalidated` module. That said I'll make sure to bring the generalizations from the existing structures into the `unvalidated` structs. Sound good?

event/src/unvalidated.rs
```diff
@@ -0,0 +1,242 @@
...
+    /// Get the data
+    pub fn data(&self) -> &Ipld {
+        &self.data
+    }
+}
+
+/// Headers for a data event
+#[derive(Debug, Serialize, Deserialize)]
+pub struct DataHeader {
+    controllers: Option<Vec<String>>,
```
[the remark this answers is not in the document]
stbrody: Yeah, I'm pretty sure `shouldIndex` is the only thing that can actually show up in the metadata for a data event currently: https://github.com/ceramicnetwork/js-ceramic/blob/be8d6d9c24509c2bd6bcae43314554a3b850b1f8/packages/stream-model-instance/src/model-instance-document.ts#L281-L286

store/src/event.rs
```diff
@@ -0,0 +1,142 @@
...
+                            .map_err(Error::new_fatal)?,
+                        payload.id(),
+                        &event_cid,
+                    ))
+                }
+            }
+        }
+        unvalidated::Event::Time(event) => {
+            let init = event.id();
+            // TODO should time event CAR files contain their init event?
```
stbrody, line 100: wouldn't this same problem exist for data events too?  I think you just have to have received the init event independently before you can even really start parsing or validating data and time events.

stbrody (commented):
left a few comments, mostly around what fields are actually present in the headers for various event types.

store/src/event.rs
```diff
@@ -0,0 +1,142 @@
...
+use ceramic_event::unvalidated;
+use cid::Cid;
+use iroh_car::CarReader;
+use tokio::io::AsyncRead;
+
+use crate::{Error, Result};
+
+// Helper function to construct an event Id from CAR data of an event
+// We should likely move this closer to where its needed but this is good enough for now.
+pub(crate) async fn event_id_from_car<R>(network: Network, reader: R) -> Result<EventId>
```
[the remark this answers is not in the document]
nathanielc: The CAR file contains other blocks that we want to parse in order to get all the data. So having access to the entire CAR file is important as part of the process of constructing the EventID. 

I briefly considered using an event based parser (i.e. SAX XML parser) and then let there be various consumers of the parser events to build up the state it needs. However that felt overly complex for the needs here as its easy to ensure we only read/parse a block once.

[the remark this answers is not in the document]
stbrody: huh?  I don't understand your concern about needing the CAR file.  My suggestion was that we fully parse everything out of the CAR file and into an Event struct with everything we'll ever need to access for the Event, and that that Event struct has a `get_event_id()` method on it.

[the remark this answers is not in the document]
nathanielc: Constructing an event ID from an event has two dependencies. 

1. The blocks in the CAR file
2. The init block in the store

The CAR file for the simplest case has two blocks that we need, the envelope and the payload. The Event type we have created represents the raw structure of the IPLD data. Specifically the Event enum represents the envelope while the Payload enum represents the payload block. We do not have a reference from the Event enum to the Payload enum. They are loosely connected via the CID in the envelope.

 In order the construct the event ID from an update event for example we do the following steps:
 
 1. Decode the CAR file and get the CID of the first root.
 2. Decode the block from the CAR for that CID into an Event (i.e. envelope)
 3. Get the CID of the payload from the of the Event (envelope)
 4. Decode the payload block from the CAR 
 5. Get the CID of the init event from the payload
 6. Read the init event from the store and decode as an Event
 7. Get the payload CID from the init Event
 8. Read the payload block from the store and decode
 9. Get the controller and model from the init payload header
 
 These steps require database reads and access to the CAR file in order to read blocks that exist in the CAR skipping a database read.
 
 Given that the Event type and its friends are data structs, meaning simply contain deserialzed data and do not contain references to each other it makes sense to me to have a function not attached to the Event type that can do this operation when provided with the dependencies.
 
 >  My suggestion was that we fully parse everything out of the CAR file and into an Event struct with everything we'll ever need to access for the Event
 
 How would you do this? Does this imply that we treat CIDs as references ? If so does that mean you can't construct an event without materializing the entire log of events back to init for that event since that is where the trail of references end? 
 
 Happy to discuss on a call if I am not making sense. TL;DR I want to decouple these data structure from how they are used because they can be used in many contexts which may not all need to do the same operations and therefore do not have the same dependencies.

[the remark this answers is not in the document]
stbrody: I wouldn't imagine materializing the entire log, just all the blocks that are a part of this event (so following all links *except* for the `prev` link), plus any extra metadata we need such as the controller and model.

As discussed on discord though, happy to leave this for a future optimization.

api/src/server/event.rs
```diff
@@ -0,0 +1,331 @@
...
+}
+
+// Helper function to get the block first from the CAR file data or otherwise from the store.
+async fn get_block<S: AccessModelStore>(
+    cid: &Cid,
+    car_blocks: &HashMap<Cid, Vec<u8>>,
+    store: &S,
+) -> Result<Vec<u8>> {
+    if let Some(bytes) = car_blocks.get(cid) {
+        Ok(bytes.clone())
```
stbrody, line 95: is there any way we could move the bytes instead of cloning them?  Just worried that these blocks could be somewhat large in some cases, would be nice to avoid copying them if possible.

api/src/server/event.rs
```diff
@@ -0,0 +1,331 @@
...
+    use std::str::FromStr;
+
+    use super::*;
+
+    use async_trait::async_trait;
+    use expect_test::{expect, Expect};
+    use mockall::{mock, predicate};
+    use tracing_test::traced_test;
+
+    const INIT_EVENT_CID: &str = "bagcqcerar2aga7747dm6fota3iipogz4q55gkaamcx2weebs6emvtvie2oha";
```
stbrody, line 115: can you include a reference to how these things were generated?  Ideally a script that's checked in to the repo?  I hate having tests with hard-coded data that no one knows how to recreate in case anything changes and the structures need to be updated.

event/src/args.rs
```diff
@@ -40,40 +45,58 @@ impl<'a, S: Signer> EventArgs<'a, S> {
...
             signer,
             controllers: vec![signer.id()],
-            parent,
+            model: parent,
             sep: SEP,
         }
     }
 
     /// Parent stream id
     pub fn parent(&self) -> &StreamId {
```
stbrody, line 54: thanks for updating parent->model here!  Should we also update the name of this accessor method while we're at it?

event/src/args.rs
```diff
@@ -84,141 +107,66 @@ impl<'a, S: Signer> EventArgs<'a, S> {
...
-    prev: Option<Cid>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    id: Option<Cid>,
-}
-
-fn create_unique() -> Result<Vec<u32>> {
+fn create_unique() -> Result<Bytes> {
     let mut rng = rand::thread_rng();
-    let mut unique = [0u32; 12];
+    //TODO is u8; 12 OK?
```
stbrody, line 112: how do we plan to address this TODO?  Can you link a ticket number where we will figure this out?

event/src/bytes.rs
```diff
@@ -0,0 +1,69 @@
+//! Thin wrapper on a Vec<u8> that correctly serializes and deserializes as bytes as opposed to a
+//! sequence of u8 integers.
+use serde::{
+    de::{self, Visitor},
+    Deserialize, Serialize,
+};
+
+/// Sequence of byte values.
+#[derive(PartialEq, Default, Debug)]
+pub struct Bytes(Vec<u8>);
```
stbrody, line 10: isn't there already a `Bytes` type from the `bytes` crate?  Do we really need to make our own?

event/src/unvalidated.rs
```diff
@@ -0,0 +1,249 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::bytes::Bytes;
+
+/// Ceramic Event
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum Event {
+    /// Time event in a stream
+    Time(Box<TimeEvent>),
```
stbrody, line 16: why do TimeEvents need to be boxed but not SignedEvents/JWS?

event/src/unvalidated.rs
```diff
@@ -0,0 +1,249 @@
...
+}
+
+/// Signed event
+#[derive(Debug, Serialize, Deserialize)]
+pub struct JWS {
+    payload: Bytes,
+    signatures: Vec<Signature>,
+}
+
+impl JWS {
```
stbrody, line 28: why are we parsing out the link Cid and capability Cid on demand every time they are asked for?  Why doesn't the JWS struct just deserialize the data into Cids upfront so that then all callers can just access them directly when needed?

I guess this is becoming a pattern of questions for me.  I feel like we keep leaning towards very lazy approaches to deserialization but I'm worried this is going to make it very easy to wind up parsing the same data over and over again.  It feels like it would be better convention to always try to parse things at the edges once then pass around strongly-typed structures throughout all of our internal code.

event/src/unvalidated.rs
```diff
@@ -0,0 +1,242 @@
...
+    pub fn data(&self) -> Option<&Ipld> {
+        self.data.as_ref()
+    }
+}
+
+/// Headers for an init event
+#[derive(Debug, Serialize, Deserialize)]
+pub struct InitHeader {
+    controllers: Vec<String>,
+    sep: String,
```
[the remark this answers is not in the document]
stbrody: nevermind, looks like `sep` is explicitly specified in both Models and MIDs, and has been since before ComposeDB launched.

store/src/sql/event.rs
```diff
@@ -338,6 +338,14 @@ impl SqliteEventStore {
             .await?;
         Ok(())
     }
+
+    async fn get_block(&self, cid: &Cid) -> Result<Option<Vec<u8>>> {
+        let block: Option<BlockBytes> = sqlx::query_as(BlockQuery::get())
+            .bind(cid.hash().to_bytes())
+            .fetch_optional(self.pool.reader())
```
stbrody, line 345: why is this `fetch_optional` when everything else in this file uses `fetch_all`?  I don't actually know what either of those functions do, so mostly just looking to increase my own understanding of how this all works.

event/src/unvalidated.rs
```diff
@@ -0,0 +1,249 @@
...
+    /// Get the data
+    pub fn data(&self) -> &D {
+        &self.data
+    }
+}
+
+/// Headers for a data event
+#[derive(Debug, Serialize, Deserialize)]
+pub struct DataHeader {
+    controllers: Option<Vec<String>>,
```
stbrody, line 158: controllers should never show up in a DataHeader because changing the controller isn't allowed.

event/src/unvalidated.rs
```diff
@@ -0,0 +1,249 @@
...
+    }
+    /// Get the data
+    pub fn data(&self) -> &D {
+        &self.data
+    }
+}
+
+/// Headers for a data event
+#[derive(Debug, Serialize, Deserialize)]
+pub struct DataHeader {
```
stbrody, line 157: `shouldIndex` can show up in a DataHeader.  Do we need to specify that field explicitly if it can show up, even if rust-ceramic never needs to inspect it?

api/src/server/event.rs
nathanielc: Not easily, however these bytes come from the CAR file so they are already in memory so we are duplicating them for a short period of time. I expect this is sufficient for now. But if we determine its an issue we can look into reference counting the data or similar.

event/src/args.rs
nathanielc: What can the unique value be in an event? Previously it was treated as u32s but I was told its always opaque bytes. If we can get a clear answer to this question then we can remove the TODO.

event/src/bytes.rs
nathanielc: Yes, this is about serde of the Bytes to be raw bytes not a sequence of u8s. The IPLD codecs are naive and assume a Vec<T> maps to an IPLD List of T. We need to special case where T is u8 so that a Vec<u8> maps to IPLD bytes type.

store/src/sql/event.rs
nathanielc: fetch_all returns a Vec,T> that may be empty of no data exists for the query. fetch_optional returns Option<T>. If fetch_optional find more than one value an error is returned. If we use fetch_all we have to explicitly check  that the vector.len <= 1.  In this case we expect at most one value to match the query so fetch_optional is the most appropriate method to use.

event/src/unvalidated.rs
nathanielc: Yes, what data type is it? A boolean?

event/src/unvalidated.rs
nathanielc: TimeEvent is large ~312 bytes and JWS is small < 8 bytes. So putting TimeEvent in a Box allows both variants to be about the same size making the enum more efficient. Honestly this is a bit of a premature optimization but clippy complained so I didn't fight it.

event/src/args.rs
stbrody: You can see both ways this can be generated here: https://github.com/ceramicnetwork/js-ceramic/blob/0b6ff7c825ef915cd24aea4591904f710cae253f/packages/stream-model-instance/src/model-instance-document.ts#L458-L466

event/src/unvalidated.rs
stbrody: except the JWS payload can be up to 256kb so there's no guarantee the JWS is actually small either.

At the least can we add some comments explaining the reasoning of whatever we decide to go with?

event/src/unvalidated.rs
stbrody: yep, boolean

event/src/unvalidated.rs
stbrody: okay, I think this makes sense at least for now.  Still slightly worried about it, but we can optimize later if need be.

api/src/server/event.rs
nathanielc: @stephhuynh18 Can you give me some insight on how these were created? My assumption was that you started up js-ceramic created a stream and gave it an update and waited for it be anchored and then grabbed the CAR data from the logs? 

@stbrody Agreed we need a way to generate this data. I don't think we have a simple way to do so now. As such I was most interested in getting real world data so we can have more confidence that the code works as intended.

event/src/unvalidated.rs
nathanielc: Yes but those bytes will be allocated on the heap via the Vec<u8> of the payload. The JWS struct is basically just two pointers.

nathanielc: In contrast the TimeEvent struct has three CIDs which are 64+ bytes each plus the string bits. So that struct is ~312 bytes (according to the Rust compiler). I'll add a comment.

event/src/args.rs
nathanielc: Ok looks like we can always assume its a sequence of bytes. Will remove the TODO, we really need to specify these details in the spec.

stbrody (approved):
lgtm

nathanielc:
Closing this work has been done

PR #329, opened 2024-05-01
chore: add bad request response to get events API

dav1do:
Add bad request response to GET events API and improve event creation 400 response.


stbrody (approved):
lgtm

PR #330, opened 2024-05-02
chore: SQL queries for event blocks dont need order_key

stbrody:


stbrody:
huh, @dav1do any idea why github doesn't seem to be running any tests against this PR?

dav1do:
> huh, @dav1do any idea why github doesn't seem to be running any tests against this PR?

I believe the action is set up to not run unless it targets main.. should probably change that.

dav1do (approved):
Here is the [workflow](https://github.com/ceramicnetwork/rust-ceramic/blob/6113064021239e78a124c9fec383b8a01cc03031/.github/workflows/rust-build-and-test.yml#L5) pointing to main PRs.. could probably remove the target but I'd have to review the action triggers to see what's allowed.

stbrody:
PR has been approved and included in the unified branch: https://github.com/ceramicnetwork/rust-ceramic/pull/335

PR #337, opened 2024-05-06
feat: Update experimental API for getting a range of events to return event root CID, not EventID

stbrody:


store/src/sql/event.rs
```diff
@@ -552,10 +552,11 @@ impl ceramic_api::AccessModelStore for SqliteEventStore {
         end: &EventId,
         offset: usize,
         limit: usize,
-    ) -> anyhow::Result<Vec<(EventId, Vec<u8>)>> {
+    ) -> anyhow::Result<Vec<(Cid, Vec<u8>)>> {
         let res = self
             .range_with_values_int(start, end, offset, limit)
-            .await?;
+            .await?
+            .map(|(event_id, value)| (event_id.cid().unwrap(), value));
```
stbrody, line 559: how bad do yall think this is?  It's a bit inefficient.  Another approach would be to update the SQL query to return the CID and then thread it through here so we get the CID from the SQL result instead of by parsing the eventID.  The problem with that approach is that the `range_with_values` method exists for both the `AccessModelStore` and `Recon::Store` traits, and the later needs it to continue to return EventIds not Cids, so if I want to take that approach I will need to add a new method with a different name that is only used by the AccessModelStore and returns Cids.  I can definitely make that change though if you think it's the better way to go.  @dav1do @nathanielc

dav1do: I'm not worried about the performance of this right now, but I wonder if we can/should change EventId to be more complex than a `Vec<u8>` wrapper? It requires a CID to build, so returning None is pretty annoying to have all these unwraps.. and if it can't be None, it seems like it shouldn't return an option. Seeing the `unwrap` is definitely a red flag.. and usually something you want to comment like you would unsafe. Crashing the entire process over bad data would be _bad_, even if that data _should_ be impossible. 

My instinct here is to return remove them (or just even just error) even though they should never exist, but I'd rather EventId be more complex and actually store the its parts and then have a `to_vec()` method. Curious to know the reason it's like this before changing it though.

```rust
        let res = self
            .range_with_values_int(start, end, offset, limit)
            .await?
            .filter_map(|(event_id, value)| {
                if let Some(cid) = event_id.cid() {
                    Some((cid, value))
                } else {
                    warn!(?event_id, "Invalid event ID missing CID");
                    None
                }
            });
```

Your point about whether we should thread CID through is interesting... I don't think you need to do it now, but it'd be nice to draw the boundary between recon and the event store more clearly so we aren't parsing the CID in/out of the EventId on both sides. 

I'm looking at that as part of IOD and consuming these changes. My current expectation is that the store will no longer implement any of the existing traits directly, and the ceramic service (or whatever it's called) will be the one that turns bytes into events and then calls into the store, so the store could take a ValidatedEvent or whatever makes the input/output easier to work with. 

My general philosophy is simple, correct, performant. Rust makes it more obvious than most languages that you're not doing something "the most optimal" way, but it's hard to know how bad it is without profiling. Rust and LLVM are pretty good about removing clones and using cache, so outside of the bad algorithm/obvious stuff, I try to avoid optimizing in favor of simplicity/clarity until it's proven to be an issue.

nathanielc: Lets not change the EventId yet, its a thin wrapper around a Vec<u8> so it can lexicographically sort. The option returned is that it is technically possible to construct an EventId from bytes that do not decode correctly into the CID. So while annoying I say we explicitly handle the case where there is no CID in the API code. 

Something like:

```
.map(|(event_id, value)| Ok((event_id.cid().ok_or_else(|| anyhow!("EventId does not have an event CID")?, value)))?;
```


Since we are gradually reducing the surface area of the EventId type itself, we can likely change it or remove it entirely from the API soon. Given that I say we wait on any more refactors of its type.

nathanielc (approved):
LGTM once we resolve the conversation re invalid event ids

store/src/sql/event.rs
```diff
@@ -552,11 +552,18 @@ impl ceramic_api::AccessModelStore for SqliteEventStore {
...
-            .await?;
-        Ok(res.collect())
+    ) -> anyhow::Result<Vec<(Cid, Vec<u8>)>> {
+        self.range_with_values_int(start, end, offset, limit)
+            .await?
+            .map(|(event_id, value)| {
+                Ok((
+                    event_id
+                        .cid()
+                        .ok_or_else(|| anyhow!("EventId does not have an event CID"))?,
```
stbrody, line 562: okay, updated this to handle if the CID is missing.  It compiles, but I'm not totally sure it's correct (or if it is I'm not entirely sure *why* it's correct).  It was actually kind of surprising to me that this compiled.

Previously, the map call returned a Map iterator structure where the values inside were a tuple of `(Cid, Vec<u8>)`.  Now the values returned in the Map are Results of that tuple (ie `Result<(Cid, Vec<ui>)>`).

I would have expected calling `collect()` on an `Iterator<Result<T>>` to return a `Vec<Result<T>>`.  But it looks like instead it is returning a `Result<Vec<T>>`?  Is it just taking the first non-ok Result and promoting that up to the top-level error?  That's what I want, but it wasn't/isn't clear to me that that's what `collect()` is doing.  How can I learn how this works?

stbrody: oh wow, okay I found this page that indicates that collect is indeed doing what I want: https://doc.rust-lang.org/rust-by-example/error/iter_result.html#fail-the-entire-operation-with-collect

This is actually quite nice, but feels unintuitive to me.  Any advice on how I can build an intuition (or get my IDE to tell me) around these types of behaviors in Rust would be appreciated

dav1do: tbh, I don't think it is super intuitive and I'm kind of surprised it works. I didn't think you could use `?` in closures so I need to review what's happening as well. The fact that [result implements FromIterator](https://doc.rust-lang.org/src/core/result.rs.html#1934) is just syntax sugar and imo kind of odd, but it is nice to avoid all the `transpose` and `map_err` business. 

I tend to use `filter_map` the other suggestions in that doc to log every error when it makes sense. The issue with the "abort immediately" approach here is that it will *never* succeed. The event data is garbage, so someone is going to have to manually fix it... the `filter_map` approach would let it return something, but we shouldn't be storing this data in the first place, and if we do have it, someone or something is going to need to intervene.

nathanielc: My intuition for it is that `collect` can construct various types from iterators. As its common to want to construct a vector from an iterator of results and fail on the first one  the Result type implements FromIteraor as David mentions. That is part one of this code, you can collect an iterator of results into a result of a collection failing on the first error.


Part two is that we explicitly return a Result from the `map` closure. There is nothing special about closures or the `?` syntax so using it within a closure is reasonable to me. To _help_ the closure understand we want to return a result we make the `Ok` variant explicit and then use `?` where needed.



Rust idoms aside why do we want to only log this error? Seems like if we have bad data we should error the request? Maybe I am missing something.

nathanielc (approved):

store/src/sql/event.rs
dav1do: Yeah, good call that ? is fine if you return a result. 

So I guess my argument for not an error is that this is only returning data we already stored, not preventing the data from being stored in the first place. So it'll never possible to stop erroring until someone goes and deletes something. IMO skipping bad data in the range would be reasonable instead of just failing until someone intervenes (or we could just delete the data because we don't expect it in our table) but it shouldn't happen.

dav1do (approved):

PR #341, opened 2024-05-06
feat: Make event creation endpoint infer EventID from event CAR file data

stbrody:


api/src/tests.rs
```diff
@@ -105,45 +105,7 @@ impl AccessModelStore for MockReconModelTest {
         Ok(None)
     }
 }
-#[tokio::test]
-async fn create_event() {
```
stbrody, line 109: let me know if you think it's worth it to try to revive this test.  Probably what I would do is take the hard-coded constants used in the unit tests in `api/src/server/event.rs` and pull them out into their own file and use them in both the unit tests in that file as well as in this unit test here.  It just feels like since the underlying store is mocked out anyway, this test isn't actually testing a whole lot of real functionality to begin with.  The actual coverage will come from the integration tests.

stbrody: oh wait, looking more deeply there is some fancy logic in the server code around passing the event into a channel to be processed by another Task.  That's probably worth having unit test coverage of.  Okay I'll go ahead with pulling those constants out and bringing this test back to life

api/src/server.rs
```diff
@@ -409,15 +410,24 @@ where
         ))
     }
 
+    // TODO(stbrody): Only take the event data, don't take an id field at all since its unused.
     pub async fn post_events(&self, event: Event) -> Result<EventsPostResponse, ErrorResponse> {
```
dav1do, line 414: we might need to change this endpoint to accept something else (e.g. EventInput) that is just a data field or make eventID optional, but that would be weird for the get endpoints.

stbrody: yep, probably going to need to introduce a new type.

I'm going to do that change in a separate PR as a follow-up

api/src/tests.rs
```diff
@@ -53,79 +119,70 @@ impl AccessInterestStore for MockReconInterestTest {
...
-        "did:key:zGs1Det7LHNeu7DXT4nvoYrPfj3n6g7d6bj2K4AMXEvg1", // cspell:disable-line
-        &Cid::from_str("baejbeihyr3kf77etqdccjfoc33dmko2ijyugn6qk6yucfkioasjssz3bbu").unwrap(), // cspell:disable-line
-        &Cid::from_str("baejbeicqtpe5si4qvbffs2s7vtbk5ccbsfg6owmpidfj3zeluqz4hlnz6m").unwrap(), // cspell:disable-line
-    );
-    let event_id_str = multibase::encode(Base::Base16Lower, event_id.to_bytes());
-    let event_data = "f".to_string();
+    let network = Network::Mainnet;
+    let expected_event_id = EventId::try_from(hex::decode(DATA_EVENT_ID).unwrap()).unwrap();
+
+    // Remove whitespace from event CAR file
```
dav1do, line 173: do we need to include the whitespace at all?

api/src/tests.rs
```diff
@@ -53,79 +119,70 @@ impl AccessInterestStore for MockReconInterestTest {
...
+        async fn value_for_cid(&self, key: &Cid) -> Result<Option<Vec<u8>>>;
+        async fn events_since_highwater_mark(
+            &self,
+            highwater: i64,
+            limit: i64,
+        ) -> Result<(i64, Vec<Cid>)>;
+        async fn get_block(& self, cid: &Cid) -> Result<Option<Vec<u8>>>;
     }
 }
-#[async_trait]
```
dav1do, line 73: thanks for cleaning this up

dav1do (approved):

PR #343, opened 2024-05-07
feat: adjust recon bounds to be lower inclusive and always store value

nathanielc:
The change consists of two related pieces:

1. Change the recon protocol to not discover keys without also knowing
   the value.
2. Change the semantics of the protocol to be lower bound inclusive.

These two changes together greatly reduce the complexity of the
protocol and reduce the complexity of implementing the red tree in the
future.


store/src/metrics.rs
```diff
@@ -61,13 +53,7 @@ impl Metrics {
 
         register!(
             key_insert_count,
```
dav1do, line 55: do we want to change the name in prometheus too?

store/src/sql/event.rs
```diff
@@ -119,52 +95,47 @@ impl SqliteEventStore {
...
         for (idx, item) in to_add.into_iter().enumerate() {
             let new_key = self.insert_key_int(&item.order_key, &mut tx).await?;
-            for block in item.blocks.iter() {
-                self.insert_event_block_int(block, &mut tx).await?;
-                self.mark_ready_to_deliver(&item.order_key, &mut tx).await?;
+            if new_key {
+                // Only add the blocks if this is a new key, otherwise ignore the value.
+                for block in item.blocks.iter() {
+                    self.insert_event_block_int(block, &mut tx).await?;
+                    self.mark_ready_to_deliver(&item.order_key, &mut tx).await?;
```
dav1do, line 132: `mark_ready_to_deliver` should be pulled out of the loop, but I have it in the IOD changes I'm working on

recon/src/recon/tests.rs
```diff
@@ -1514,80 +1501,6 @@ async fn two_in_sync() {
     .await
 }
 
-#[test(tokio::test)]
-async fn dog_missing_sync() {
```
dav1do, line 1518: is there a reason these tests were deleted?

recon/src/recon/tests.rs
```diff
@@ -1810,97 +1735,141 @@ async fn alternating() {
             dog: [a: A, c: C, d: D, f: F, h: H, j: J, l: L, n: N, p: P, q: Q, s: S, u: U, w: W, y: Y, z: Z]
         -> range_req({𝚨 h(a, b, c, e, g, i, k, m, o, p, r, t, v, x, z)#15 𝛀 })
             cat: [a: A, b: B, c: C, e: E, g: G, i: I, k: K, m: M, o: O, p: P, r: R, t: T, v: V, x: X, z: Z]
-        <- range_resp({𝚨 h(a, c, d, f, h, j, l)#7 n}, {n h(p, q, s, u, w, y, z)#7 𝛀 })
+        <- range_resp({𝚨 h(a, c, d, f, h, j, l)#7 n}, {n h(n, p, q, s, u, w, y, z)#8 𝛀 })
```
dav1do, line 1738: This test (`alternating`) seems to have been hit hardest in terms of round trip increases but most stayed the same or dropped a bit (e.g. `subset_interest`).

dav1do (commented):
This is awesome! I wish it were easier to reason about the impact, the test changes are substantial but comparing outcomes and roundtrips makes me think it's acceptable and correct. The simplification and code deletion is a huge benefit.

I'll take care of the easy comments I added (often related to things I did 😅).

recon/src/recon/tests.rs
nathanielc: Yes, they tested if synced work when a peer knew about a key but not a value. Since that is not longer a valid state the tests are not needed.

store/src/metrics.rs
nathanielc: yep, renamed

recon/src/recon.rs
```diff
@@ -490,65 +458,62 @@ pub trait Store {
...
-    async fn first(
-        &self,
-        left_fencepost: &Self::Key,
-        right_fencepost: &Self::Key,
-    ) -> Result<Option<Self::Key>> {
-        Ok(self
-            .range(left_fencepost, right_fencepost, 0, 1)
-            .await?
-            .next())
+    async fn first(&self, range: Range<&Self::Key>) -> Result<Option<Self::Key>> {
```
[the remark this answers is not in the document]
nathanielc: You are right, and in fact I think these are now unused by Recon. Going to push up a commit that removes them.

recon/src/recon/tests.rs
nathanielc: I had a shower thought on how I might reduce roundtrips. I may try it out today...

p2p/src/node.rs
```diff
@@ -1274,7 +1274,7 @@ mod tests {
         }
         async fn keys_with_missing_values(
```
dav1do, line 1275: can we delete this field from the trait?

recon/src/client.rs
```diff
@@ -98,14 +98,6 @@ where
...
-    /// Report all keys in the range that are missing a value
-    pub async fn keys_with_missing_values(&self, range: RangeOpen<K>) -> Result<Vec<K>> {
-        let (ret, rx) = oneshot::channel();
-        self.sender
-            .send(Request::KeysWithMissingValues { range, ret })
-            .await?;
-        rx.await?
-    }
     /// Report the local nodes interests.
     pub async fn interests(&self) -> Result<Vec<RangeOpen<K>>> {
```
dav1do, line 102: Do we want to keep `RangeOpen` around? It seems like we're just turning it into a normal range now but it seems to be working fine.

dav1do (approved):
A couple small comments/questions but overall I think this is great.

dav1do:
> A couple small comments/questions but overall I think this is great.

I ran a local benchmark (write on 0, sync to 1) and it seems to performing similarly to what we were seeing before with key/value messages (and importantly they ended up in sync).
image

recon/src/client.rs
nathanielc: Yes, RangeOpen is still useful for communicating interests. Its another way of saying interests boundaries are always fenceposts.

PR #344, opened 2024-05-07
feat: in order delivery of ceramic events

dav1do:
This implements in order event delivery. That means that we must have the event the `prev` CID points to before we deliver it to any clients on the feed API. We currently assume one prev, so this will get more involved with multi-prev. When accepting new events via API, we error inline if we don't have the history, but for recon, we send them to a background task to be updated when it's possible. In this case, we still insert them as we want to synchronize, we just wait to notify anyone via the feed.

On startup we try to process all undelivered events (i.e. if we shutdown with some in memory). ~~We also do batches on an interval, and grab undelivered events if the queue isn't too full (this logic seems flawed, particularly if we keep rediscovering the same events we can't deliver).~~ We send new events to the task and if anything is unblocked, we'll mark them as delivered so they go out on the feed.

The current implementation may require loading an entire stream history into memory to find the required delivered event, and will not remove it unless a first event is found. Currently, there are not any indexes about ceramic event data on disk and everything is opaque bytes, so we have to parse them. If could query id/prev for an event, we could reduce that. Also, and worse, pending events are kept in memory until a prev is discovered. Generally, this number will be low, but if we're very out of sync with a peer and there are many streams with updates, we could potentially discover them in the worst possible order and have to keep many things in memory. The total number incoming at once is limited by the channel, but we can keep putting them in memory without expiring or delivering them currently.

This depends on a refactor to clean up the store crate types and public API. All the traits/higher level crates now depend on the `service` crate directly and it implements the traits for API, recon, IPFS. That is in #354. 

It also is now interleaved with the changes to update to the new event types (#355 / [add-builder](https://github.com/ceramicnetwork/rust-ceramic/tree/add-builder)) which will need to be sorted out/cleaned up. I have a squashed version + a commit I added with some docs/pub.

Follow ups:
- Add metrics for queue depth and other things
- Be smarter: 
  - Keep track of dropped/missed tasks and find them later
  - Expire events in memory after some time and prune them
  - Limit the total number of events we can ever have in memory (requires better finding/keeping track of dropped)


service/src/event/ordering_task.rs
```diff
@@ -0,0 +1,730 @@
...
+                    info!("Server dropped the ordering task. Processing once more before exiting...");
+                    let _ = state
+                        .process_events(&pool)
+                        .await
+                        .map_err(Self::log_error);
+                    return;
+                }
+            };
+
+            if state
```
dav1do, line 143: Remove this or do it very infrequently. We should process anything blocked when we hear about it through the `rx_new` channel. 

As thing get worse (i.e. we have a ton of streams we can't deliver), this will loop over them all doing a lot of pointless work we can't make progress on. It may catch things that were missed somehow, but there are better ways to deal with this.

service/src/event/service.rs
```diff
@@ -32,4 +69,222 @@ impl CeramicEventService {
...
+        }
+
+        Ok(res)
+    }
+
+    async fn verify_history_inline(
+        &self,
+        insert_after_history_check: Vec<(DeliverableMetadata, EventInsertable)>,
+        insert_now: &mut Vec<EventInsertable>,
+        notify_new: &mut Vec<DeliveredEvent>,
```
dav1do, line 230: this is annoying to duplicate. we need to know about the init/stream ID in one of these and not the other. I think it can be simplified by adding a bit more ceramic  info to things after parsing blocks (it was very much just recon K,V before)

service/src/tests/ordering.rs
```diff
@@ -0,0 +1,247 @@
...
+        "radius": 1,
+        "red": 2,
+        "green": 3,
+        "blue": 4,
+        "raw": data2.as_slice(),
+    });
+
+    let init = init_event(model, &signer).await;
+    let (event_id, car) = (
+        random_event_id(Some(&init.envelope_cid().to_string())),
```
nathanielc, line 79: nit, this might read a little clearer as 

```
event_id_builder().with_event(&init.envelope_cid().to_string()).build()
```

This way its clear that no random ID is actually being generated.

nathanielc: Also not sure if it matters but this event ID is wrong as well. The event_id's contain the model but we do not use the model we generated above to construct the event id, instead a hard coded model is used https://github.com/ceramicnetwork/rust-ceramic/blob/50cc8baa101ba56c5c1bea31573cb0378d36c800/service/src/tests/mod.rs#L28

service/src/event/service.rs
```diff
@@ -32,4 +70,222 @@ impl CeramicEventService {
...
+        };
+        let meta = maybe_init_prev.map(|(cid, prev)| DeliverableMetadata {
+            init_cid: cid,
+            prev,
+        });
+        Ok((insertable, meta))
+    }
+
+    #[tracing::instrument(skip(self, items), level = tracing::Level::DEBUG, fields(items = items.len()))]
+    pub(crate) async fn insert_events_from_carfiles<'a>(
```
nathanielc, line 124: Could we make this function private and expose (i.e. `pub(crate)`) two functions that make it clear if history is required? Reading the test code its easy to miss the true, false boolean argument.

Maybe function names like `insert_events_ordered` and `insert_event_unordered`, or something else related to the name `history`? Not really picky on the name but having two different names that make it clear how it behaves would help in reading the code.

nathanielc (approved):
Looks great, a few small changes requested and some questions to make sure I understand things. However no need for re-review

service/src/event/service.rs
```diff
@@ -121,71 +122,40 @@ impl CeramicEventService {
...
-            } else {
-                insertable.deliverable(true);
-                notify_task_new.push(DeliveredEvent::new(
-                    insertable.body.cid,
-                    insertable.body.cid,
-                ));
-                insert_now.push(insertable);
-            }
+    #[tracing::instrument(skip(self, items), level = tracing::Level::DEBUG, fields(items = items.len()))]
+    pub(crate) async fn insert_events_from_carfiles_remote_history<'a>(
```
nathanielc, line 143: Can you add a similar comment to this function ? 

```
    /// This function is used to insert events from a carfile NOT requiring that the history is local to the node and expecting to learn the history eventually.
```

dav1do: yep, good call

PR #345, opened 2024-05-07
fix: parse unsigned init events

nathanielc:
Parses unsigned init events and can construct an EventId from its car file.

### DIFF
--- a/api/src/server/event.rs
+++ b/api/src/server/event.rs
@@ -34,9 +34,13 @@ where
         car_blocks.insert(cid, bytes);
     }
     let event_bytes = get_block(&event_cid, &car_blocks, store).await?;
-    let event: unvalidated::Event =
+    let event: unvalidated::Event<Ipld> =
         serde_ipld_dagcbor::from_slice(&event_bytes).context("decoding event")?;
-    let init_id = match event {
+    let (init_id, payload) = match event {
+        unvalidated::Event::Time(event) => (
+            event.id(),
+            get_init_event_payload(&event.id(), &car_blocks, store).await?,
+        ),
         unvalidated::Event::Signed(event) => {
             let link = event
                 .link()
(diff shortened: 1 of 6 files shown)


dav1do (approved):
:shipit: 

Looks like the root description changes are missing again. Not sure where that got dropped but a simple server gen should add it back. Maybe #338 merged here without the corresponding generation?

event/src/unvalidated.rs
```diff
@@ -11,13 +11,15 @@ use crate::bytes::Bytes;
...
-pub enum Event {
+pub enum Event<D> {
     /// Time event in a stream
     // NOTE: TimeEvent has several CIDs so its a relatively large struct (~312 bytes according to
     // the compiler). Therefore we box it here to keep the Event enum small.
     Time(Box<TimeEvent>),
     /// Signed event in a stream
     Signed(JWS),
+    /// Unsigned event in a stream
+    Unsigned(InitPayload<D>),
```
stbrody, line 22: not important or worth changing in this PR, but could the payload *ever* be anything other than Ipld?  Is there a reason to keep this generic?

api/src/server/event.rs
```diff
@@ -34,9 +34,13 @@ where
         car_blocks.insert(cid, bytes);
     }
     let event_bytes = get_block(&event_cid, &car_blocks, store).await?;
-    let event: unvalidated::Event =
+    let event: unvalidated::Event<Ipld> =
         serde_ipld_dagcbor::from_slice(&event_bytes).context("decoding event")?;
-    let init_id = match event {
+    let (init_id, payload) = match event {
```
stbrody, line 39: nitpick, can we call this `init_payload` so it's clear it's not the payload of the event itself?

stbrody (approved):
lgtm

event/src/unvalidated.rs
nathanielc: The idea is if you know what it is say in specific application code you can directly deserialize into that value instead of first deserializing into IPLD and then reconstructing the application specific data from the IPLD data. This was how the Event API was setup initially for something that @dbcfd was using. Not sure what specific use case he needed it for but that is the idea.

PR #349, opened 2024-05-08
fix: lower correctness test network ttl

3benbox:
Correctness tests should be cleaned up sooner.


.github/workflows/rust-build-and-test.yml
```diff
@@ -137,6 +137,7 @@ jobs:
           curl -LO https://raw.githubusercontent.com/3box/ceramic-tests/main/networks/basic-rust.yaml
           ./bin/yq -i '.spec.ceramic[0].ipfs.rust.image = strenv(RUST_CERAMIC_IMAGE)' basic-rust.yaml
           ./bin/yq -i '.metadata.name = strenv(KERAMIK_NETWORK_NAME)' basic-rust.yaml
+          ./bin/yq -i '.spec.ttlSeconds = 3600' basic-rust.yaml
```
smrz2001, line 140: Could move this to the call to the test driver at line 160 😊

smrz2001 (approved):
:shipit:

PR #354, opened 2024-05-16
refactor: add service crate

dav1do:
This is the refactor to add a "service" crate to support business logic like event validation. Pulled out of #344 and will adjust it to target this branch. Mostly moves things around, some renaming/adjusting. Also modified the server insert background task to not panic after shutdown signal and `recv_many`.


service/src/tests/event.rs
```diff
@@ -39,140 +39,6 @@ macro_rules! test_with_dbs {
     }
 }
 
-test_with_dbs!(
```
dav1do, line 42: moved to store crate because we use predefined event ID ranges to assert the hash. at this layer we can't build a random body for a CID but we can insert raw data at the store crate and run the queries there

nathanielc (approved):

PR #355, opened 2024-05-16
feat: Enforce no unexpected fields in events

stbrody:
This PR temporarily removes the recently-added Builder structs, so as to focus first on getting the underlying Event types right and ensure all fields in the Events are specified explicitly.  Once this is set we can add back the builders in a way that preserves type safety better and makes it impossible to build invalid events.


event/src/unvalidated/payload/signed.rs
```diff
@@ -6,6 +6,7 @@ use std::collections::BTreeMap;
 
 /// A signed event payload.
 #[derive(Debug, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
```
stbrody, line 9: this is the only one I'm not sure of.  I believe this struct represents the JWS envelope and I don't actually know if it's safe to assume that that doesn't have extra fields.  I think I'm leaning towards removing the deny_unknown_fields tag on this struct since the JWS envelope format isn't really a part of the Ceramic protocol spec so feels like maybe a place where we should be looser in what we accept.  Curious for other's thoughts here.

stbrody: Wait, actually I'm confused.  There's also the `SignedEvent` struct that has a `Jws` struct: https://github.com/ceramicnetwork/rust-ceramic/blob/c803ac49ec7b136a3aca42927c5696649238841a/event/src/unvalidated/ext/signed_event.rs#L17

@dbcfd - what's the relationship between this signed::Payload struct and the SignedEvent struct?  Are they both used or can one be deleted?

dbcfd: SignedEvent is what js-ceramic requires, hence it is in ext. signed::Payload is what comes from js-ceramic when it is written to a car file.

stbrody:
Note I have this test of sending extra fields in ceramic-tests: https://github.com/3box/ceramic-tests/pull/129.  Not sure how to run that test against a version of rust-ceramic from this branch...

stbrody:
closing.  Event changes will be included in https://github.com/ceramicnetwork/rust-ceramic/pull/357.  Adding the serde directive to restrict unexpected fields will come in a follow-up PR

PR #356, opened 2024-05-16
fix: recon protocol hang with large diffs

nathanielc:
If two nodes each had 2K+ events that the other node did not have it was possible for the protocol to deadlock as both nodes tried to write those events without read the values from the other node.

The fix was accomplished by splitting the protocol into two loops, a read loop and a write loop. They communicate with each other using message passing. The code now has both a clean separation of concerns as well as the behavior that we concurrently read and write from the network so we do not enter a state where we are exclusively writing to the network.

### DIFF
--- a/recon/src/protocol.rs
+++ b/recon/src/protocol.rs
@@ -191,7 +191,7 @@ where
         metrics.clone(),
     );
 
-    let read = read(sync_id.clone(), stream, role, to_writer_tx, metrics.clone());
+    let read = read(sync_id, stream, role, to_writer_tx, metrics.clone());
 
     // In a recon conversation there are 4 futures being polled:
     //
@@ -682,9 +682,8 @@ where
                     .range(range.first, range.last, 0, usize::MAX)
                     .await?;
                 for key in keys {
-                    if let Some(value) = recon.value_for_key(key.clone()).await? {
-                        yield Value { key, value };
-                    }
+                    let value = recon.value_for_key(key.clone()).await?.ok_or_else(|| anyhow!("recon key does not have a value: key={}", key))?;
+                    yield Value { key, value };
                 }
             }
         }
--- a/recon/src/protocol.rs
+++ b/recon/src/protocol.rs
@@ -218,7 +218,7 @@ where
 {
     if let Some(sync_id) = &sync_id {
         // Record sync_id on the tracing span
-        tracing::Span::current().record("sync_id", &sync_id);
+        tracing::Span::current().record("sync_id", sync_id);
     }
 
     pin_mut!(stream);
@@ -277,7 +277,7 @@ where
 {
     if let Some(sync_id) = &sync_id {
         // Record sync_id on the tracing span
-        tracing::Span::current().record("sync_id", &sync_id);
+        tracing::Span::current().record("sync_id", sync_id);
     }
     pin_mut!(sink);
 
@@ -418,7 +418,7 @@ where
                     .send(ToWriter::SendAll(
                         self.common
                             .process_remote_missing_ranges(ranges.clone())
-                            .map(move |value| value.map(|value| InitiatorMessage::Value(value)))
+                            .map(|value| value.map(InitiatorMessage::Value))
                             .boxed(),
                     ))
                     .await
@@ -441,9 +441,7 @@ where
         self.pending_ranges += ranges.len();
         to_writer
             .send(ToWriter::SendWIPLimited(
-                ranges
-                    .map(move |range| InitiatorMessage::RangeRequest(range))
-                    .collect(),
+                ranges.map(InitiatorMessage::RangeRequest).collect(),
             ))
             .await
             .map_err(|err| anyhow!("{err}"))?;
@@ -556,7 +554,7 @@ where
                     .send(ToWriter::SendAll(
                         self.common
                             .process_remote_missing_ranges(ranges.clone())
-                            .map(move |value| value.map(|value| ResponderMessage::Value(value)))
+                            .map(move |value| value.map(ResponderMessage::Value))
                             // Send the range hash after we have sent all keys so the remote learns we are in
                             // sync.
                             .chain(once(Ok(ResponderMessage::RangeResponse(ranges))))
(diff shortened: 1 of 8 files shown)


recon/src/protocol.rs
```diff
@@ -619,38 +660,25 @@ where
...
     }
     // The remote is missing all keys in the range send them over.
-    async fn process_remote_missing_range(
+    fn process_remote_missing_ranges(
         &mut self,
-        range: &RangeHash<R::Key, R::Hash>,
-    ) -> Result<()> {
-        // TODO: This logic has two potential failure modes we need to test them
-        // 1. We allocate memory of all keys in the range, this can be very large.
-        // 2. We spend a lot of time writing out to the stream but not reading from the stream.
```
nathanielc, line 628: Both of these points are addressed with this change.

recon/src/protocol.rs
```diff
@@ -142,275 +139,360 @@ where
...
+        sync_id.clone(),
+        sink.sink_map_err(anyhow::Error::from),
+        to_writer_rx,
+        init,
+        role.finish(),
+        PENDING_RANGES_LIMIT,
+        metrics.clone(),
+    );
+
+    let read = read(sync_id.clone(), stream, role, to_writer_tx, metrics.clone());
```
dav1do, line 194: nit: don't think we need to clone sync_id in this line but it probably doesn't actually happen

recon/src/protocol.rs
```diff
@@ -142,275 +139,360 @@ where
...
+                // Send all message that fit within the work in progress limit.
+                if in_progress < wip_limit {
+                    let capacity = wip_limit - in_progress;
+                    let mut sent = false;
+                    for message in messages.drain(..std::cmp::min(messages.len(), capacity)) {
+                        in_progress += 1;
+                        sent = true;
+                        trace!(?message, "feeding message");
+                        metrics.record(&MessageSent(&message));
+                        sink.feed(ReconMessage::new(sync_id.clone(), message))
```
dav1do, line 342: I'm curious how feed/flush vs send/send_all behaves. This seems like the natural way to do this rather than keep things in memory, but I'd rather have a synchronous feed API and then async flush().. if I'm giving up control I might as well `send()`, no?

Anyway, before this seemed to be related to the slowing message exchange in conversations. There definitely was a flush after X setting coming into play in our case, but I guess this is a long winded way of saying maybe we should just call send (or map and send_all if we can)? 

If we feed and it's full, we might wait a while without actually being able to send things... I guess we could flush every X items or something, and maybe it's silly to worry about now. 

Basically, my opinion is that using the feed API makes sense, but given how abstracted everything is through libp2p/os/etc to the actual I/O, I think it might be better to just hit send and let the lower layers buffer things and keep things flowing steadily.

recon/src/protocol.rs
```diff
@@ -422,166 +504,134 @@ where
...
+        Ok(RemoteStatus::Active)
     }
 }
 
-type ResponderValueResponseFn<K, H> =
-    fn(Option<String>, Value<K>) -> ReconMessage<ResponderMessage<K, H>>;
-
 // Responder implements the [`Role`] where it responds to incoming requests.
-struct Responder<R, S>
+struct Responder<R>
```
dav1do, line 541: ❤️ love it ❤️

recon/src/protocol.rs
```diff
@@ -619,38 +660,25 @@ where
...
+        let recon = self.recon.clone();
+        try_stream! {
+            for range in ranges {
+                // TODO this holds all keys in memory
+                // Paginate or otherwise stream the keys and values back out
+                let keys = recon
+                    .range(range.first, range.last, 0, usize::MAX)
                     .await?;
+                for key in keys {
+                    if let Some(value) = recon.value_for_key(key.clone()).await? {
```
dav1do, line 676: If there isn't a value should we be concerned at all now that we should discover them together or am I forgetting about something?

recon/src/protocol.rs
```diff
@@ -594,22 +644,13 @@ where
...
+struct Common<R: Recon> {
     recon: R,
-
-    value_resp_fn: V,
-
-    sync_id: Option<String>,
 }
 
-impl<R, S, E, In, Out, V> Common<R, S, V>
+impl<R> Common<R>
```
dav1do, line 651: YES 💪

dav1do (approved):
A few comments / questions but I think this looks great!

recon/src/protocol.rs
nathanielc: > Basically, my opinion is that using the feed API makes sense, but given how abstracted everything is through libp2p/os/etc to the actual I/O, I think it might be better to just hit send and let the lower layers buffer things and keep things flowing steadily.

My understanding of the API is that using `send` prevents any lower layers from buffering as it calls flush down the layers of abstraction ensuring the message has been sent all the way to the remote node.

From the docs on feed:

```
A future that completes after the given item has been received
by the sink.

Unlike `send`, the returned future does not flush the sink.
It is the caller's responsibility to ensure all pending items
are processed, which can be done via `flush` or `close`.
```

From the docs on send:

```
A future that completes after the given item has been fully processed
into the sink, including flushing.

Note that, **because of the flushing requirement, it is usually better
to batch together items to send via `feed` or `send_all`,
rather than flushing between each item.**

```

As recommended when we use `send_all` or `feed` depending on if we have a stream of messages or just a set of messages in memory. Additionally we always call `flush` after doing a batch of `feed` calls so its not possible that a message is not flushed and therefore stuck in pending land. 

This differs from how we did this previously because we used feed on a counter so its was possible a message landed on the wrong side of the count and wasn't flushed to the other node. This cannot happen with this code change as we always flush after calling feed.

dav1do: We discussed and no changes needed 😄 and I agree the loop/flush every time now versus the count before is good. I just wanted to call out the sink handling as an area we might want to keep in mind, but intuitively, and according to the docs, using feed like this is the right thing to do.

recon/src/protocol.rs
nathanielc: Good call, going to convert this into an error.

PR #357, opened 2024-05-17
feat: Update event types and builder APIs to prevent building events that dont align with the protocol spec

stbrody:


core/src/stream_id.rs
```diff
@@ -101,12 +101,13 @@ impl StreamId {
...
 
     /// Convert the stream id to a vector
-    pub fn to_vec(&self) -> anyhow::Result<Vec<u8>> {
+    pub fn to_vec(&self) -> Vec<u8> {
         // Use self.len() here when we have cid@0.10
         let buf = Vec::new();
         let mut writer = std::io::BufWriter::new(buf);
-        self.write(&mut writer)?;
-        Ok(writer.into_inner()?)
+        // safe to unwrap because self.write should never fail.
```
stbrody, line 108: @nathanielc let me know if this is what you had in mind or if there's more I should do to handle this error.

nathanielc: LGTM

nathanielc (approved):

event/src/unvalidated/builder.rs
```diff
@@ -1,343 +1,324 @@
...
-impl<T> Additional for DataBuilder<T> {
-    fn with_additional(mut self, key: String, value: Value) -> Self {
-        self.additional.insert(key, value);
-        self
+/// State with data added, also supports the optional should_index field.
+pub struct DataBuilderWithData<D> {
+    id: Cid,
+    prev: Cid,
+    data: D,
+    should_index: Option<bool>,
```
dav1do, line 163: we should add `with_sep` to data builder too, right? we need sep/sepValue to build a recon event ID.

dav1do (approved):

event/src/unvalidated/builder.rs
stbrody: Data events don't have the separator in their header.  So the separator needs to be fetched from the init event to build the event ID.

dav1do: oh right. I knew that.. and still think it's annoying 😅

PR #358, opened 2024-05-17
feat: Disallow unexpected fields in events

stbrody:


event/src/unvalidated/event.rs
```diff
@@ -50,6 +50,7 @@ impl<D> From<signed::Envelope> for RawEvent<D> {
 
 /// Time event
 #[derive(Debug, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)] // TODO Should this be here?
```
stbrody, line 53: Time events feel different than data and init events because they'll generally be created by the protocol (C1 nodes or the CAS) rather than by users of the protocol.  So maybe it makes sense to be looser in what we expect here so it's easier for us to make additive changes to the events in the future without breaking older nodes?

On the other hand, nothing stops users from making their own TimeEvents and publishing them, in which case issues with field name collisions could still happen.

I'm conflicted, probably leaning slightly towards not adding this directive and keeping validation looser here, but could be easily swayed if anyone thinks it's worth locking it down now.

event/src/unvalidated/signed.rs
```diff
@@ -116,6 +116,7 @@ impl<D: serde::Serialize> Event<D> {
 
 /// A signed event envelope.
 #[derive(Debug, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)] // TODO Should this be here?
```
stbrody, line 119: since these structures come more from the JWS spec than from anything we control, it seems like maybe we'd be better off being loose here in what we accept

stbrody (commented):
Made the PR with adding the directive to restrict extra fields every place that it *might* make sense to put it (that I could find at least), but I left a few comments of places where it might be better to not actually add it.  Would like to get some additional opinions on those places.

event/src/unvalidated/event.rs
```diff
@@ -18,7 +18,7 @@ pub enum Event<D> {
 
 /// Ceramic Event as it is encoded in the protocol.
 #[derive(Serialize, Deserialize)]
-#[serde(untagged)]
+#[serde(untagged, deny_unknown_fields)]
```
dav1do, line 21: I'm not convinced we want to be this strict. I'm trying to articulate/think through it so bear with me. Some of it is probably that the events/schemas aren't self describing. 

I think doing this will basically make it impossible to be an old node on the network, as no one will be able to talk to you if we ever added a new field (you'll reject every event they send you).

I think it may make sense to have 2 versions (one for API ingest and one for sync). We could reject anything coming in on the API outright, but we shouldn't do that to peers if they are running a newer version of the protocol than we are. It feels like this makes everything a hard fork. I'd be happy to be wrong about this so shoot me down 😄.

stbrody: Agreed this is a problem, going to need a different approach

stbrody:
Closing this PR.  Will put up a new one that only enforces the no extra fields constraint for events received via the http api, but continues to accept events with unexpected fields if they are learned via recon.

PR #362, opened 2024-05-20
chore: Deserialize CAR into Event before extracting EventID

stbrody:
There should be no functional changes here, just code refactoring.

This PR splits the `event_id_from_car` function into two phases: first parse the CAR data into an Event struct, solely based on the data in the car without any reliance on the blockstore.  Then, given the Event struct and access to the blockstore, get the EventID.  This is setting the groundwork to be able to add the code that re-serializes the Event struct back into a CAR file to compare the bytes and ensure that no extra fields are present.

### DIFF
--- a/api/src/server/event.rs
+++ b/api/src/server/event.rs
@@ -10,19 +10,26 @@ use tracing::debug;
 
 use crate::AccessModelStore;
 
-// Helper function to construct an event Id from CAR data of an event
-// We should likely move this closer to where its needed but this is good enough for now.
+// Helper function to construct an event ID from CAR data of an event coming in via the HTTP api.
 pub async fn event_id_from_car<R, S>(network: Network, reader: R, store: &S) -> Result<EventId>
 where
     R: AsyncRead + Send + Unpin,
     S: AccessModelStore,
+{
+    let (event_cid, event) = event_from_car(reader).await?;
+    event_id_for_event(event_cid, event, network, store).await
+}
+
+async fn event_from_car<R>(reader: R) -> Result<(Cid, unvalidated::Event<Ipld>)>
+where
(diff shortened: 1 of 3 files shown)


event/src/unvalidated/signed.rs
```diff
@@ -36,6 +36,26 @@ impl<D: serde::Serialize> Event<D> {
         )
     }
 
+    /// Factory for building an Event.
+    pub fn new(
```
dav1do, line 40: I started debating if we want to have a public new function like this (or if we could calculate the CIDs ourselves to avoid some potential errors), but we're going to have to turn it into a validated event at some point, which will involve checking all the CIDs and signatures, so it seems fine.

api/src/server/event.rs
```diff
@@ -10,76 +10,138 @@ use tracing::debug;
...
         }
-        unvalidated::RawEvent::Unsigned(event) => (event_cid, event),
-    };
+        unvalidated::RawEvent::Unsigned(event) => {
+            Ok((event_cid, unvalidated::Event::Unsigned(event)))
+        }
+    }
+}
 
+async fn event_id_for_event<S>(
```
dav1do, line 77: It might be nice to consolidate some of the behavior here with the service crate, which is now implementing `AccessModelStore` (aside: that should really be renamed to events or something) to have a shared `store -> &[u8] -> event` flow but no need to do it now.

dav1do (approved):
LGTM. A couple musings but nothing to do now.

api/src/server/event.rs
stbrody: on the naming piece, I through together https://github.com/ceramicnetwork/rust-ceramic/pull/365 to rename the traits to InterestStore and EventStore.

PR #364, opened 2024-05-21
feat: adjust prom metrics to capture path info and status codes 

dav1do:
Added a middleware to capture all http requests status code, path and method in prometheus. In order to avoid blowing up prom with millions of labels, we rewrite a few known routes. This deletes the previous trait middleware metric handling and just relies on this for kubo and api http servers. I left store metrics alone as they provide additional useful visibility.

Example: `rate(http_api_requests_total{k8s_pod_name="ceramic-0-0"}[$__rate_interval])` includes capturing 500s when something goes wrong (like someone deleting the db for lulz)

image


one/src/http.rs
```diff
@@ -126,3 +131,95 @@ where
...
+        &mut self,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: Request<Body>) -> Self::Future {
+        let start_time = Instant::now();
+        let method = req.method().clone();
+        let path = req.uri().path().to_string(); // strip query args
```
dav1do, line 164: remove comment (path doesn't have query params)

one/src/http_metrics.rs
```diff
@@ -15,34 +11,30 @@ use prometheus_client::{
...
+    method: String,
+    path: String,
+    status: u16,
+}
+
+impl HttpMetrics {
     /// Register and construct Metrics
     pub fn register(registry: &mut Registry) -> Self {
-        let sub_registry = registry.sub_registry_with_prefix("ceramic_api");
+        let sub_registry = registry.sub_registry_with_prefix("http_api");
```
dav1do, line 37: open to a better name (e.g. ceramic_api) but I tried to avoid the existing sub registries I was going to delete

one/src/lib.rs
```diff
@@ -497,32 +498,32 @@ impl Daemon {
...
+        let kubo_rpc_service =
+            http::MakeMetricsService::new(kubo_rpc_service, http_metrics.clone());
+        let ceramic_service = http::MakeMetricsService::new(ceramic_service, http_metrics);
 
         // Compose both services
         let service = http::MakePrefixService::new(
             ("/ceramic/".to_string(), ceramic_service),
             ("/api/v0/".to_string(), kubo_rpc_service),
         );
+        // let service = http::MakeMetricsService::new(service, Arc::new(http_metrics));
```
dav1do, line 526: clean up

3benbox (approved):
Amazing!

PR #365, opened 2024-05-22
chore: Rename store traits to InterestStore and EventStore

stbrody:


api/src/server.rs
```diff
@@ -130,8 +130,9 @@ impl TryFrom<models::Interest> for ValidatedInterest {
     }
 }
 
+/// Trait for accessing persistent storage of Interests
 #[async_trait]
-pub trait AccessInterestStore: Send + Sync {
+pub trait InterestStore: Send + Sync {
```
stbrody, line 135: do you know why these traits are in the `api` crate to begin with?  Wouldn't these more logically belong to the `store` crate?

dav1do: you can't have circular crate dependencies, so the crate owns/requires the behavior declares the trait it needs and whoever can implement and provide it. in our case, the service crate implements this trait, using the api and store crate as its dependencies. then `one` is able to pull them all in and compose them together. recon also declares traits that are implemented by service. this way, you rely on the interface without knowing about the implementation.

it does get a bit annoying trying to write tests in this crate because we can't use a store. we could create integration tests (a tests folder) and put tests in there if we wanted to go `api -> service -> store` in a test.

dav1do (approved):
Thanks for doing this. I don't think we even need both traits anymore and we could just make one API trait that has all the behaviors we need. We used to have generic bounds that made it more necessary, but I don't think we need to make that change now.

PR #367, opened 2024-05-23
fix: recon protocol stall with no shared interests

dav1do:
It was possible for the conversation to stall if the peers shared no interests. Now the initiator will send a finished message immediately and the conversation will end, allowing a future iteration to discover new overlapping interests.

### DIFF
--- a/recon/src/protocol.rs
+++ b/recon/src/protocol.rs
@@ -18,7 +18,7 @@ use ceramic_core::RangeOpen;
 use ceramic_metrics::Recorder;
 use futures::{pin_mut, stream::BoxStream, Sink, SinkExt, Stream, StreamExt, TryStreamExt};
 use serde::{Deserialize, Serialize};
-use tokio::{join, sync::mpsc, time::Instant};
+use tokio::{sync::mpsc, time::Instant};
 use tokio_stream::once;
 use tracing::{instrument, trace, Level};
 use uuid::Uuid;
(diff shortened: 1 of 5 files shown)


recon/src/protocol.rs
```diff
@@ -497,17 +496,26 @@ where
...
-                            .recon
-                            .initial_range(interest)
-                            .await
-                            .context("querying initial range")?,
-                    );
+                if interests.is_empty() {
+                    to_writer
+                        .send(ToWriter::Finish)
+                        .await
+                        .map_err(|err| anyhow!("{err}"))
```
dav1do, line 503: going to remove this `map_err` (copilot is a bad pilot sometimes)

dav1do: jk, it's a good pilot and this is needed 😊

stbrody (approved):

recon/src/protocol.rs
```diff
@@ -497,17 +496,26 @@ where
...
+                        .context("sending finish")?;
+                    return Ok(RemoteStatus::Finished);
```
dav1do, line 505: do we want to do an early return on the other exit step too? [that is, here](https://github.com/ceramicnetwork/rust-ceramic/pull/367/files#diff-7c57e2a3f12a04af8303d7883cea6c88b855ecea5ecb0356653b3d4a2b4945f8R539)

nathanielc: no we do not want any early returns from this

nathanielc (approved):

nathanielc (approved):

PR #369, opened 2024-05-27
fix: ensure all segments of stream history are connected and delivered at once

dav1do:
With long streams, it was possible to stop delivering events when we had partial chains even though we had everything we needed on the node. We didn't trust that we'd get every message on the channel and try to reconstruct whatever exists at the time (which may be things that were just written but we haven't yet heard about) so it was possible to basically race with ourself putting pieces together and stopping. Now we discover and connect all the events in the chain so that they exist on the node, not necessarily when we hear of the last event on the channel. This adds a test to this repo that was failing quite consistently (though I did leave `while cargo test recon_lots_of_streams; do :; done` running for a while to be sure this helped). This issue is also exposed in the new [js-ceramic test](https://github.com/ceramicnetwork/js-ceramic/pull/3205).

The implementation is not very efficient, particularly for long streams, but may perform badly in a different way with many shorter streams, especially if we're very out of sync. It mostly comes down to our inability to make certain queries. I'm going to write up a migration proposal to add more stream information to the database tables so we can answer ceramic questions such as "what are all the events for the stream" and "what is the current tip of this stream". We could also include other header info such as controller, sep and sep value. This is likely required to help with event validation (we always need to find the init event information) and would simplify this code, improve its performance, make the start up discovery logic cleaner, as well as opening up flat recon concepts and more sophisticated node queries.

Sorry that the commits aren't very targeted, I ended up moving some things around and the diff isn't super clear. I pulled some of this history/ordering info out of the `InsertEventProcessor` struct in the service functions themselves, and then call into OrderState with the events to build the state and return Ok/Err. This requires that init events are already stored as the task only deals with events that have a prev, so there was a change to insert any init event before validating history, which could result in a partial write of the events requested, but the events written are valid. I moved the event discovery into the `StreamEvents` struct and there is a loop to recurse if we find pieces of the chain we can connect, to make sure we link and deliver everything that exists. And then I realized that all of this complexity is just a band aid around not knowing any stream information.


dav1do:
Closing as this is a mess. I'm going to do resolve this with the changes to store stream data in ceramic one (aes-87) that simplifies and improves it.

PR #373, opened 2024-06-06
chore: rename prometheus registry and metrics for better consistency

dav1do:
This will break existing dashboards (of which there are few well defined examples I believe), but it will make building new ones much easier as all metrics will start with `ceramic_one_` now. This also renamed `http_api_http_api` to `ceramic_one_http_api_info` and `ceramic_store_store_query_durations` will now be `ceramic_one_store_query_durations`, and other store metrics will be under `ceramic_one_store`.

The existing tokio, recon, libp2p, bitswap subregistries are all under ceramic_one now:
image

### DIFF
--- a/metrics/src/core.rs
+++ b/metrics/src/core.rs
@@ -16,7 +16,7 @@ pub(crate) struct Core {
 
 impl Default for Core {
     fn default() -> Self {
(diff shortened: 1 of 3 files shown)


smrz2001 (approved):
🥳

PR #375, opened 2024-06-10
feat: capture stream metadata in database and use for IOD

dav1do:
We now store stream metadata in the database so we can find all events for a stream. This simplifies the in order delivery behavior. We do make more database queries now but use less memory to try to keep track of events. This could be more efficient, but for now I went with simple and hopefully good enough, especially while streams are short. If streams start being longer, this may not be a very good approach.

We run a task to update deliverable status for streams. It gets notifications when an event for a recon event is inserted and is currently deliverable, and it tries to sort any other pieces out and update them. It includes start up operation to load streams and mark them deliverable as well. Finally, there's a data migration run on daemon start to populate the new tables.

This should fix the test in [js-ceramic](https://github.com/ceramicnetwork/js-ceramic/pull/3205). It requires #381  and #382.


migrations/sqlite/20240530125008_event_header.up.sql
```diff
@@ -0,0 +1,9 @@
+-- Add up migration script here
+CREATE TABLE IF NOT EXISTS "ceramic_one_event_header" (
```
nathanielc, line 2: How does this table get populated for existing dbs?

nathanielc (commented):
Overall looking good

migrations/sqlite/20240530125008_event_header.up.sql
dav1do: I'm going to tackle that part today 😅. I'm planning to follow the kubo approach and basically add a CLI flag to not prompt, otherwise ask/exit if they should be applied. It will require reading every event, parsing, and storing the headers so it can't (easily) be done via a sql statement. I'm open to other approaches... was thinking I'd create a table like below and then query/update at start.

```sql
create table ceramic_one_data_migrations (
    version text unique not null,
    started_at timestamp not null default current_timestamp,
    completed_at timestamp
  );
  ```

File: event/src/unvalidated/payload/init.rs [hunk 2]
```diff
@@ -26,7 +26,7 @@ impl<D> Payload<D> {
 }
 
 /// Headers for an init event
-#[derive(Default, Serialize, Deserialize)]
+#[derive(Default, Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
```
dav1do, line 29: call out

store/src/sql/access/event.rs
```diff
@@ -7,33 +7,94 @@ use std::{
...
+    /// Create a new delivered event
+    fn new(cid: Cid, stream_cid: StreamCid) -> Self {
+        Self { cid, stream_cid }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+/// The result of inserting events into the database
+pub struct InsertResult {
+    /// True for new keys, false for existing keys. In same order as the input. BAH THIS ISN'T TRUE
```
dav1do, line 46: Clean up

store/src/sql/access/event.rs
```diff
@@ -87,41 +199,61 @@ impl CeramicOneEvent {
...
+    /// For local writes (i.e. over the API), `require_history` should be true. For events discovered over recon, it should be false.
     pub async fn insert_many(
         pool: &SqlitePool,
         to_add: &[EventInsertable],
+        require_history: bool,
     ) -> Result<InsertResult> {
         let mut new_keys = vec![false; to_add.len()];
+        let mut delivered = Vec::with_capacity(to_add.len());
+        let mut undelivered = Vec::with_capacity(to_add.len());
+        // TODO: this changes the order of the keys so the response is not in the same order as the input as claimed
```
dav1do, line 220: call out / follow up

store/src/sql/access/stream.rs
```diff
@@ -0,0 +1,147 @@
...
+    },
+    Error, Result, SqlitePool, SqliteTransaction,
+};
+
+/// Access to the stream and related tables. Generally querying events as a stream.
+pub struct CeramicOneStream {}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+/// Represents a stream event in a way that allows ordering it in the stream. It is metadata and not the event payload itself.
+pub struct StreamCommit {
```
dav1do, line 17: open to a better name

store/src/sql/access/stream.rs
```diff
@@ -0,0 +1,147 @@
...
+            })
+            .collect();
+
+        Ok(res)
+    }
+
+    /// Load streams with undelivered events to see if they need to be delivered now.
+    pub async fn load_stream_cids_with_undelivered_events(
+        pool: &SqlitePool,
+        last_cid: StreamCid,
```
stbrody, line 77: Can you add a comment for what is `last_cid`?

store/src/sql/access/stream.rs
```diff
@@ -0,0 +1,147 @@
...
+use crate::{
+    sql::entities::{
+        EventHeader, EventHeaderRow, EventType, IncompleteStream, StreamCid, StreamCommitRow,
+        StreamRow,
+    },
+    Error, Result, SqlitePool, SqliteTransaction,
+};
+
+/// Access to the stream and related tables. Generally querying events as a stream.
+pub struct CeramicOneStream {}
```
stbrody, line 13: this name makes me think the type represents a single Stream, rather than representing the subsystem for accessing and working with Streams.

Can we call it `CeramicStreamStore` or something like that?

store/src/sql/entities/event.rs
```diff
@@ -129,13 +151,71 @@ impl EventInsertableBody {
...
+        let mut blocks = vec![]; // use a hashset and store the CID instead of recalculating?
         while let Some((cid, data)) = reader.next_block().await.map_err(Error::new_app)? {
             let ebr = EventBlockRaw::try_new(&event_cid, idx, roots.contains(&cid), cid, data)
                 .map_err(Error::from)?;
             blocks.push(ebr);
             idx += 1;
         }
-        Ok(Self::new(event_cid, blocks))
+
+        let ev_block = blocks
```
stbrody, line 162: this is a really long function, can we pull this part out into a helper?

stbrody: Actually, can you just use `event_from_car` for all this parsing?  https://github.com/ceramicnetwork/rust-ceramic/blob/107be36d27c73ae96a07614f69503596c48d4f10/api/src/server/event.rs#L23

store/src/sql/entities/event_header.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+impl EventHeaderRow {
+    pub fn insert() -> &'static str {
+        "INSERT INTO ceramic_one_event_header (cid, stream_cid, event_type, prev) VALUES ($1, $2, $3, $4)"
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum EventHeader {
```
stbrody, line 25: I don't love using the word "header" here when there is a literal "header" field in events that is different than what this type represents.  Maybe we could use the word "metadata" for this concept instead?  I don't think the protocol/rust-ceramic has a concept of "metadata" yet.  js-ceramic/the aggregated MIDs have metadata, but that's at a higher level, so I think it's okay giving the word a different meaning at this layer for the core protocol.

store/src/sql/entities/stream.rs
```diff
@@ -0,0 +1,61 @@
...
+            e.delivered IS NOT NULL as "delivered" 
+        FROM ceramic_one_stream s
+            JOIN ceramic_one_event_header eh on eh.stream_cid = s.cid
+            JOIN ceramic_one_event e on e.cid = eh.cid
+        WHERE s.cid = $1"#
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct IncompleteStream {
```
stbrody, line 39: comment on the type please.  This PR introduces a lot of new types and I'm worried about how confusing it will be to keep track of them all and how they're different.

I believe this type represents a stream where we know about some events but there are holes in the history meaning some of the events are still undeliverable - is that right?  Can we add a comment to that effect?

store/src/sql/entities/stream.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+impl IncompleteStream {
+    pub fn fetch_all_with_undelivered() -> &'static str {
+        r#"
+        SELECT DISTINCT s.cid as "stream_cid" 
+        FROM ceramic_one_stream s
+            JOIN ceramic_one_event_header eh on eh.stream_cid = s.cid
+            JOIN ceramic_one_event e on e.cid = eh.cid
+        WHERE e.delivered is NULL and s.cid > $1
```
stbrody, line 50: comment for what the arguments are?

service/src/event/ordering_task.rs
```diff
@@ -1,158 +1,108 @@
...
-                    if new > 0 {
-                        modified = Some(newly_added_buf.into_iter().map(|ev| ev.init_cid).collect::<HashSet<InitCid>>());
+                _new = rx_stream_update.recv_many(&mut updated_streams, 100) => {
+                    debug!(?updated_streams, "incoming updates!");
+                    for stream in updated_streams {
+                        state.add_stream(stream);
                     }
                 }
+                _ = interval.tick() => {
+                    // For this and the else branch, we allow the loop to complete and if state has anything to process, it will be processed.
```
stbrody, line 80: this comment confused me at first.  The loop doesn't complete in this case, we continue looping.  We just fall through this select! statement to the bottom of the loop body, then we loop back to the beginning.  Can we clarify this comment?

service/src/event/ordering_task.rs
```diff
@@ -1,158 +1,108 @@
...
+                    .process_delivered_events(&pool, delivered_events)
                     .await
                     .map_err(Self::log_error)
                     .is_err()
             {
                 return;
             }
-            last_processed = std::time::Instant::now();
+
+            if state
```
stbrody, line 98: looking over every stream every 10 minutes makes me nervous.  I'm worried it is going to paper over bugs in the IOD logic so we never notice them, users will just have occasional sporadic high latency for certain events to be delivered, in a way that will be hard to notice or debug.

It shouldn't be necessary to look at all events, except once at startup.  After that we should be able process streams only in reaction to incoming events, without anything slipping through the cracks.  If there's a bug in our logic, I'd rather us notice it because an event *never* gets delivered, vs the event getting delivered 10 minutes late when this loop happens to notice it.

service/src/event/ordering_task.rs
```diff
@@ -175,599 +125,292 @@ impl OrderingTask {
...
+            .filter_map(|(cid, event)| if event.delivered { Some(cid) } else { None })
+    }
+
+    fn order_events(&mut self, start_with: EventCid) -> VecDeque<Cid> {
+        let mut deliverable = VecDeque::with_capacity(self.prev_map.len());
+
+        deliverable.push_back(start_with);
+        self.remove_by_event_cid(&start_with);
+        let mut tip = start_with;
+        // technically, could be in deliverable set as well if the stream is forking?
```
stbrody, line 197: we should definitely make sure we have tests that cover the case of stream forks and multiple histories each with gaps that get filled in at different times.

stbrody (changes requested):
Haven't finished reviewing `ordering_task.rs` or `service.rs` yet, but there's a lot of comments already I want to get out and my brain is starting to turn mushy reading all this code so I think this is all I can get through today.  Will plan to revisit tomorrow.

store/src/sql/access/event.rs
```diff
@@ -61,6 +123,56 @@ impl CeramicOneEvent {
...
+            }
+        }
+
+        for event in insert_check_history {
+            match &event.body.header {
+                EventHeader::Init { .. } => {
+                    unreachable!("Init events should have been filtered out")
+                }
+                EventHeader::Data { cid, prev, .. } | EventHeader::Time { cid, prev, .. } => {
+                    // check for prev in this set and fallback to database
```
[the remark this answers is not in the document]
dav1do: Hmm, I think you're right. Good catch! I should probably write a better test and make sure 😄. This would fail if we somehow got 3 writes from the API for a single stream out of order. For recon discovered events, we would miss marking them immediately (but wouldn't error), and the stream would get sent to the task and it would find them, but it shouldn't need to.

store/src/sql/access/event.rs
```diff
@@ -61,6 +123,56 @@ impl CeramicOneEvent {
...
         }
     }
+
+    /// Splits the evtents into deliverable and non-deliverable events. Non-deliverable events are checked against the database to see if they can be delivered.
+    /// If `require_history` is true, an error is returned if any event is missing history, oherwise they will not be marked as deliverable.
+    async fn with_deliverable_marker<'a>(
+        events: &'a [EventInsertable],
+        pool: &SqlitePool,
+        require_history: bool,
+    ) -> Result<Vec<(bool, &'a EventInsertable)>> {
```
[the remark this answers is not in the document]
dav1do: yeah, I 100% agree with this. I discussed a bit with Nathaniel and we want to make the an event from the event crate part of the public interface, but some of it comes out of event validation and the migrations stuff he's working on right now.

I made a lot of types to wrap specific behavior (like a group of CIDs rather than a tuple) or a specific database query result but it merits more thought and clean up.

store/src/sql/access/event.rs
```diff
@@ -87,41 +199,61 @@ impl CeramicOneEvent {
...
+        // TODO: this changes the order of the keys so the response is not in the same order as the input as claimed
+        // we currently throw this result away in recon but the recon::Store trait claims that it will be met. would be nice to remove that claim
+        let to_add = Self::with_deliverable_marker(to_add, pool, require_history).await?;
         let mut tx = pool.begin_tx().await.map_err(Error::from)?;
 
-        for (idx, item) in to_add.iter().enumerate() {
-            let new_key = Self::insert_key(&mut tx, &item.order_key).await?;
+        for (idx, (deliverable, item)) in to_add.iter().enumerate() {
+            let new_key = Self::insert_key(&mut tx, &item.order_key, *deliverable).await?;
+            let candiadate = CandidateEvent::new(item.cid(), item.stream_cid());
```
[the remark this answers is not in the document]
dav1do: thanks :). I renamed this locally in the migration to populate the new tables and will push it tonight even though it's not quite ready (i.e. working)

store/src/sql/access/stream.rs
dav1do: Yeah, that makes sense. These access structs were intended to be a sort of grouping of higher level functionality than just single table/entity writes but the names are pretty much copied straight from the tables which is a misnomer for sure.

store/src/sql/access/stream.rs
dav1do: yeah, I 100% was just trying to avoid `event` and used some js-ceramic (and git) language. I like `StreamEventMetadata` or similar - I think I even named it something like that at first but changed it because I didn't want to see Event again 🙈.

store/src/sql/access/stream.rs
dav1do: Yeah, good call. I was going to call it `cid_cursor` or something but wasn't sure how to best convey the "start after this CID" semantic.

store/src/sql/entities/event.rs
dav1do: yeah, this behavior should be in the event crate and not here. we're duplicating it in api and store to avoid circular dependencies as they originally had slightly different concerns but as we've hardened the Event types, it probably makes sense to provide this behavior in one place.

store/src/sql/entities/event_header.rs
dav1do: I'm on board with that. I originally named the table `event_metadata` and think I want to rename it back as event_header is unclear.

store/src/sql/entities/stream.rs
dav1do: yeah, I will review/consolidate the types. This is basically a wrapper around a CID for serialization to support a specific query and is only public to the module.

store/src/sql/access/event.rs
```diff
@@ -87,41 +199,61 @@ impl CeramicOneEvent {
...
+                // the insert failed so we didn't mark it as deliverable.. is this possible?
+                if !new_key {
+                    Self::mark_ready_to_deliver(&mut tx, &item.cid()).await?;
+                }
+            } else {
+                undelivered.push(candiadate);
+            }
             if new_key {
                 for block in item.body.blocks.iter() {
                     CeramicOneBlock::insert(&mut tx, block.multihash.inner(), &block.bytes).await?;
```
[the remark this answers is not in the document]
dav1do: everything is wrapped in a single (large) transaction so it gets rolled back together

migrations/sqlite/20240530125008_event_header.up.sql
```diff
@@ -0,0 +1,9 @@
...
+    "cid" BLOB NOT NULL, -- event cid
+    "event_type" INTEGER NOT NULL, -- enum EventType: Init, Data, Time
+    "stream_cid" BLOB NOT NULL,  -- id field in header. can't have FK because stream may not exist until we discover it but should reference ceramic_one_stream(cid)
```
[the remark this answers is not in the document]
dav1do: foreign key. the insert fails if we haven't discovered the init event yet if it's a foreign key, but it _should_ join on the table in the same way a foreign key would.

migrations/sqlite/20240529202212_stream.up.sql
```diff
@@ -0,0 +1,8 @@
+-- Add up migration script here
+CREATE TABLE IF NOT EXISTS "ceramic_one_stream" (
+    "cid" BLOB NOT NULL, -- init event cid
+    "sep" TEXT NOT NULL,
```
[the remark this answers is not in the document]
dav1do: yeah, I wanted to discuss controller and whether we want to fully normalize it out. it's easy to capture for just the init event here, but we may want to include it for every event as well. I wasn't sure if we wanted many to many or just a single one. having it normalized for every event makes it possible to do "flat recon" things but it would still be a decent amount of work to implement.

migrations/sqlite/20240530125008_event_header.up.sql
```diff
@@ -0,0 +1,9 @@
...
+    "prev" BLOB, -- prev event cid (can't have FK because node may not know about prev event)
+    PRIMARY KEY(cid),
+    FOREIGN KEY(cid) REFERENCES ceramic_one_event(cid)
```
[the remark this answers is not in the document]
dav1do: I didn't want to make that table any wider as we do large scans and we have to load the entire row off disk even if we only read some of the columns. Since these types are pretty small it probably doesn't matter, but I kind of expected `event_metadata` (probably going with that since `event_header` is silly) to keep growing and I didn't want to end up with a 25 column event table.

If I could redefine these tables (wow, that was quick), I'd rename event -> event_hash and make event_header/event_metadata the event table and put delivered and all the new data in there.

service/src/tests/mod.rs
```diff
@@ -202,33 +174,44 @@ async fn get_events_with_model(model: &StreamId) -> [(EventId, Vec<u8>); 3] {
...
 
-    [
-        (event_id, car),
-        (data_id, data_car),
-        (data_id_2, data_car_2),
-    ]
+    let mut events = Vec::with_capacity(number);
+    events.push((event_id, car));
+    let mut prev = init_cid;
+    for _ in 0..number {
```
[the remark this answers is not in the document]
dav1do: this loops from 0 to number-1. I tend to use the playground to sanity check if I'm ever unsure ([for example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e71e0b4d7b0bd8f9f4f422ebc916b192))

dav1do (commented):
Thanks for the thorough review Spencer! I pretty much agree with everything you mentioned. Here's my summary of what I think I should do:

- fix the bugs 😄 
- rename `ceramic_one_event_header` to `ceramic_one_event_metadata` (or possibly just put the data in the event table but I'd rather not as I wrote in that comment)
- remove `commit` from the codebase and just refer to events
- review and rename/consolidate types. A number of them were specific to queries for serialization or grouping a few CIDs and useful data without relying on a tuple just for a bit of sanity, but it could be a lot cleaner. most of them aren't public outside of the module but also may not be necessary.
  - this includes relying on something from the event crate and possibly sharing carfile parsing with the api server 
- clean up comments. some were just missed/cruft/typos, some were me reasoning about why something was correct/okay but don't add any value and are a bit confusing now

And there's possibly more I missed but I'll go over each one as I change things tomorrow.

migrations/sqlite/20240529202212_stream.up.sql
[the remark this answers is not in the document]
stbrody: I think for now we should assume that there is only ever a single controller, and that the controller for a stream never changes.  Both of those are true today, and if either were to ever change it would require work across the entire stack, so I think requiring a database migration if/when that were to happen would be okay.  For now let's keep it simple.

migrations/sqlite/20240530125008_event_header.up.sql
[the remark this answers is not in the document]
stbrody: okay fair enough.  Was also just talking to @smrz2001 about the fact that for self-anchoring we'll probably want to track whether we learned of an event via the API or via Recon.  That might be something we want to add to this metadata table as well.

service/src/tests/mod.rs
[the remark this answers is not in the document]
stbrody: right, which would create `number` events in the loop, plus the one init event created outside the loop

store/src/sql/access/event.rs
```diff
@@ -7,33 +7,94 @@ use std::{
...
         sqlite::SqliteTransaction,
     },
-    CeramicOneBlock, CeramicOneEventBlock, Error, Result, SqlitePool,
+    CeramicOneBlock, CeramicOneEventBlock, CeramicOneStream, Error, Result, SqlitePool,
 };
 
 static GLOBAL_COUNTER: AtomicI64 = AtomicI64::new(0);
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+/// An event that has been delivered to the client. Generally returned from a batch of inserts.
```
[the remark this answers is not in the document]
dav1do: Reworked things quite a bit.. this is called `InsertedEvent` now

store/src/sql/entities/stream.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+impl StreamRow {
+    pub fn insert() -> &'static str {
+        "INSERT INTO ceramic_one_stream (cid, sep, sep_value) VALUES ($1, $2, $3) returning cid"
+    }
+}
+
+#[derive(Debug, Clone, sqlx::FromRow)]
+pub struct StreamCommitRow {
```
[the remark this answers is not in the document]
dav1do: Removed all instances of commit. EventHeaderRow has been changed to just be a container for queries related to the entity

service/src/event/ordering_task.rs
dav1do: removed most of the comments and took out the select!. hopefully it's clearer now, it's still a bit rough

service/src/event/ordering_task.rs
dav1do: yeah, agreed. the trigger was only reviewing things in memory we failed to process for some reason, but I removed that in favor of explicitly trying anything left over at the beginning of the loop

service/src/tests/mod.rs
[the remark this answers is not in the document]
dav1do: ah, I see what you were asking. yeah, it does. I renamed things to make it more clear (like `get_init_plus_n_events_with_model` 😄)

PR #378, opened 2024-06-13
feat: add builder for time events

nathanielc:


event/src/unvalidated/event.rs
```diff
@@ -48,17 +53,85 @@ impl<D> From<signed::Envelope> for RawEvent<D> {
...
+            let block_bytes = serde_ipld_dagcbor::to_vec(&block)?;
+            let block_cid = cid_from_dag_cbor(&block_bytes);
+            writer.write(block_cid, block_bytes).await?;
+        }
+        writer.finish().await?;
+        Ok(car)
+    }
+}
+/// Raw Time Event as it is encoded in the protocol.
 /// TODO: Parse out all the blocks for the full witness proof.
```
stbrody, line 115: this TODO can probably be removed now?

event/src/unvalidated/builder.rs
```diff
@@ -203,9 +212,151 @@ impl<D> DataBuilder<DataBuilderWithData<D>> {
...
+            .expect("should always be at least one edge");
+        let prev = leaf_edge
+            .get(*leaf_index)?
+            .ok_or_else(|| anyhow!("leaf index should always exist"))?;
+        let prev = match prev {
+            Ipld::Link(prev) => *prev,
+            _ => bail!("leaf indexed value should always be a Cid"),
+        };
+        let root_bytes = serde_ipld_dagcbor::to_vec(root)?;
+        let root_cid = cid_from_dag_cbor(&root_bytes);
```
stbrody, line 335: I don't understand how this is working.

The root is an IPLD array with two elements, each the CID of the two child nodes in the tree.  But the proof only has one of the child nodes in the path.  So then how can you compute the root CID without knowing the CID of the other child of the root that *isn't* a part of the path for this TimeEvent?


EDIT: Nevermind I understand now after looking at how this code is used in the unit test.  See my comment above, but I found the names and comments on the functions for building up the path confusing and lead to me not understanding how they were actually used until I saw the unit test.

event/src/unvalidated/builder.rs
```diff
@@ -203,9 +212,151 @@ impl<D> DataBuilder<DataBuilderWithData<D>> {
...
+    chain_id: String,
+    tx_hash: Cid,
+    tx_type: String,
+}
+impl TimeBuilderState for TimeBuilderWithTx {}
+
+impl TimeBuilder<TimeBuilderWithTx> {
+    /// Specify the root of the proof.
+    /// The index is an index into the edge that should be followed.
+    pub fn with_root(self, index: usize, edge: Ipld) -> TimeBuilder<TimeBuilderWithRoot> {
```
stbrody, line 275: the comments and names here confused me.  The thing called `edge` here isn't really the edge since it contains data not part of the proof path.  It really actually just is the IPLD encoding of the root node.  Same for the `with_edge` function below, the `edge` argument is really the entire IPLD object for this intermediate tree node.  Then the `index` is really the edge.  Can we rename `edge` to `node` and `index` to `edge_index`, or something like that?  Here and also for the `with_edge` function?

event/src/unvalidated/builder.rs
```diff
@@ -319,15 +468,42 @@ mod tests {
...
+            .with_root(0, ipld! {[prev, Ipld::Null, metadata_cid]})
+            .build()
+            .unwrap();
+
+        let event_car_str = multibase::encode(
+            multibase::Base::Base64Url,
+            event.encode_car().await.unwrap(),
+        );
+        assert_eq!(TIME_EVENT_CAR, event_car_str);
+    }
```
stbrody, line 508: it would be nice to have a unit test showing how this works with a tree that has intermediate nodes not just the root.  But low priority and not worth delaying the rest of the stack of PRs this is depending on, so feel free to just add a TODO for now.

stbrody (approved):

event/src/unvalidated/builder.rs
nathanielc: This is good feedback, I wasn't really sure how to name these bits. I'll rename them.

event/src/unvalidated/builder.rs
nathanielc: FWIW the migration tests do exercise the code paths for 10 deep trees.

PR #379, opened 2024-06-13
refactor: consolidate car serde logic into event crate

nathanielc:
Additionally we now have clear roundtrip tests


stbrody (approved):
nice

PR #380, opened 2024-06-13
feat: add cacao serde to unvalidated::Events

nathanielc:


event/src/unvalidated/event.rs
```diff
@@ -119,6 +119,19 @@ where
...
+                        let capability_bytes =
+                            car_blocks.get(&capability_cid).ok_or_else(|| {
+                                anyhow!("Signed Event CAR data missing block for capability")
+                            })?;
+                        let capability: signed::cacao::Capability =
+                            serde_ipld_dagcbor::from_slice(capability_bytes)
+                                .context("decoding capability")?;
+                        Ok((capability_cid, capability))
+                    })
+                    .transpose()?;
```
stbrody, line 134: what is `transpose()` doing?  I'm having a hard time googling for any docs on it.

event/src/unvalidated/signed/mod.rs
```diff
@@ -8,16 +10,19 @@ use ipld_core::ipld::Ipld;
...
 use super::{cid_from_dag_cbor, cid_from_dag_jose};
 
 /// Materialized signed Event.
 pub struct Event<D> {
     envelope: Envelope,
-    payload: Payload<D>,
     envelope_cid: Cid,
+    payload: Payload<D>,
     payload_cid: Cid,
+    capability: Option<(Cid, Capability)>,
```
stbrody, line 25: should we be consistent with the other two and either change this to two fields (capability and capability_cid) or change the envelope and payload fields to tuples?

event/src/unvalidated/signed/cacao.rs
```diff
@@ -0,0 +1,155 @@
...
+    /// Get the signature
+    pub fn signature(&self) -> &Signature {
+        &self.signature
+    }
+}
+/// Header for a CACAO
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Header {
+    #[serde(rename = "t")]
+    r#type: String,
```
stbrody, line 36: `r#type`?  Why the weird field name?  Can we also comment what this field is?

stbrody (approved):
a few small questions though generally looks good

event/src/unvalidated/event.rs
nathanielc: Its this https://doc.rust-lang.org/std/option/enum.Option.html#method.transpose

event/src/unvalidated/signed/mod.rs
nathanielc: I don't want two Options because they would be always correlated (i.e. both None or both Some). But for the required fields I like the clarity of named fields

event/src/unvalidated/signed/cacao.rs
nathanielc: `type` is a keyword in rust but you can use the `r#` syntax to still use it. I prefer using the full names for things vs abbreviations. As for a comment there isn't much in the spec about its purpose, the getter for the field has a simple comment.

PR #382, opened 2024-06-14
feat: add process to track and run data migrations

dav1do:
The upcoming changes to normalize some of event payload requires running a data migration a la nosql style. This provides tables to track the daemon version and the state of migrations. They can be run by calling the daemon with the `--migrate-data` flag. If there are required migrations, the daemon will fail to start.


store/src/migration.rs
```diff
@@ -0,0 +1,212 @@
...
+            required_migrations,
+            pool,
+        })
+    }
+
+    /// Determines whether migrations are needed. Will mark all migrations as "complete" on a fresh install.
+    pub async fn needs_migration(&self) -> Result<bool> {
+        let new_install = self.is_new_install().await?;
+
+        if new_install {
```
stbrody, line 90: nitpick: this feels like a weird place for this logic to live.  Seems like there should be an `init` method or something like that that is called once at startup that handles this, rather than bundling logic into this boolean accessor method.

That said this migration code is going to be fairly self-contained and mostly only run during startup anyway, so it probably doesn't matter that much.

stbrody (approved):
lookin good!

one/src/lib.rs
```diff
@@ -325,6 +333,7 @@ impl Daemon {
 
     async fn run_int<I1, I2, E1, E2, E3>(
```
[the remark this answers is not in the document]
dav1do: yeah, I'll rename it `run_internal`. It's needed to monomorphize the services into all the traits required by different components.

store/src/migration.rs
dav1do: Yeah, agreed. I changed it to happen at start up and refactored some things into functions.. sorry, I started with a one line change and then couldn't help myself.

Thinking about this made me realize that this should all be behind &mut and that ceramic one should probably expose a `CeramicDataMigration` service or something so the binary is completely unaware of all store crate things. I started making those changes and backed them out because that's something for another PR.

dav1do:
So I realized the migrations to extract metadata isn't as urgent as I thought (I fixed IOD to not have bugs without it, and it might be more efficient tbh - will have the cleaned up PR today, still need to split it/make it easier to review). I do think we want to do it before GA, but I don't think I have the new table structures schema quite right, so it might make sense to pause this for a couple days. This doesn't really do much, and it doesn't hurt to merge this and store version info, but we may want to track `started_at` or other data (so we can see if you downgrade in history). And we may want to support running migrations in the background (slowly read events and extract metadata but maybe it's better to just rip through everything for now), which would mean the flag might not be quite right.

Anyway, @stbrody, curious to hear your thoughts/discuss briefly before you spend any more time on this (and I really appreciate how much you've already spent).

PR #385, opened 2024-06-17
fix: TimeEvents from single-write anchor batches don't have a separate root block

stbrody:
The downside of this PR is that now it is possible to express with the builder API an event that is invalid and will fail to build at runtime - by specifying a prev that is different than the last edge in the last leaf node in the witness path.

I guess it might be possible to avoid this, by having the penultimate builder state have two methods, of which only one can be called.  One that lets you specify the root node of a witness tree, the other that lets you specify the prev CID in the case that there is no root b/c the event was anchored in a single-element batch.  I guess I didn't go that way because it felt like it would be confusing to readers why you could either specify a root or a prev but not both without extensive comments, and because it would require two separate terminal states for the builder.

Just typing it out makes me feel like it wouldn't be that bad actually.  Now I kind of want to go and make that change...

EDIT: New approach here: https://github.com/ceramicnetwork/rust-ceramic/pull/386

### DIFF
--- a/event/src/unvalidated/builder.rs
+++ b/event/src/unvalidated/builder.rs
@@ -242,6 +242,26 @@ pub struct TimeBuilderWithId {
 impl TimeBuilderState for TimeBuilderWithId {}
 
 impl TimeBuilder<TimeBuilderWithId> {
+    /// Specify the CID of the event being anchored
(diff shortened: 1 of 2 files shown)


stbrody:
closed in favor of https://github.com/ceramicnetwork/rust-ceramic/pull/386

PR #387, opened 2024-06-18
fix: IOD long streams could remain undelivered

dav1do:
Address AES-83 and replaces part of #375 and incorporates a lot of the feedback from that one. Depends on #390.

This changes the initial "deliver all undelivered events" process to happen when creating a `CeramicEventService`, which means it could take a while, and will block (or prevent if it fails) the server from starting rather than happen in the background.

There are new/better tests. Some where deleted/replaced that were specific to the IOD implementation details and the higher API is now tested exclusively. The general premise is to send recon events to the task, and it will track them in memory if they need history. Delivered events will be be stored if we have the stream in memory, and will trigger a review of the stream if they unblock something we have. If we don't have the stream, we drop them (but we need to learn about them jic we needed them). 

When reviewing streams, we order as many events as we can and then insert them. We use mutable state, but things should be cancel and retry safe, as we don't drop things until we've successfully committed to the database. The previous bugs were due to some state management mistakes about when to keep things around, when to trigger a review, when to look things up (so pretty much everything 😞). This implementation should be simpler, handles these situations correctly, does less work and uses an enum rather than flags to make things harder to mess up but is pretty much a rewrite of `ordering_task.rs`.

I'm not happy with the structs needed to wrap different varieties of the event (unvalidated, store with order key or without, etc). Cleaning that up will make event validation easier. This should fix the test in https://github.com/ceramicnetwork/js-ceramic/pull/3205.

### DIFF
--- a/service/src/event/order_events.rs
+++ b/service/src/event/order_events.rs
@@ -1,0 +1,100 @@
+use std::collections::HashSet;
+
+use ceramic_store::{CeramicOneEvent, EventInsertable, SqlitePool};
+use cid::Cid;
+
+use crate::Result;
+
+use super::service::EventHeader;
+
+pub(crate) struct OrderEvents {
+    pub(crate) deliverable: Vec<(EventInsertable, EventHeader)>,
+    pub(crate) missing_history: Vec<(EventInsertable, EventHeader)>,
+}
+
+impl OrderEvents {
+    /// Groups the events into lists of those with a delivered prev and those without. This can be used to return an error if the event is required to have history.
+    /// The events will be marked as deliverable so that they can be passed directly to the store to be persisted.
+    /// Will look up the prev from the database if needed to check if it's deliverable (could possibly change this for recon and allow the ordering task to handle it?)
+    pub async fn try_new(
+        pool: &SqlitePool,
+        mut candidate_events: Vec<(EventInsertable, EventHeader)>,
+    ) -> Result<Self> {
+        let new_cids: HashSet<Cid> =
+            HashSet::from_iter(candidate_events.iter().map(|(e, _)| e.cid()));
+        let mut deliverable = Vec::with_capacity(candidate_events.len());
+        candidate_events.retain(|(e, h)| {
+            if e.deliverable() {
+                deliverable.push((e.clone(), h.clone()));
+                false
+            } else {
+                true
+            }
+        });
+        if candidate_events.is_empty() {
+            return Ok(OrderEvents {
+                deliverable,
+                missing_history: Vec::new(),
+            });
+        }
+
+        let mut prevs_in_memory = Vec::with_capacity(candidate_events.len());
+        let mut missing_history = Vec::with_capacity(candidate_events.len());
+
+        while let Some((mut event, header)) = candidate_events.pop() {
+            match header.prev() {
+                None => {
+                    unreachable!("Init events should have been filtered out since they're always deliverable");
+                }
+                Some(prev) => {
+                    if new_cids.contains(&prev) {
+                        prevs_in_memory.push((event, header));
+                        continue;
+                    } else {
+                        let (_exists, prev_deliverable) =
+                            CeramicOneEvent::deliverable_by_cid(pool, &prev).await?;
+                        if prev_deliverable {
+                            event.body.set_deliverable(true);
+                            deliverable.push((event, header));
+                        } else {
+                            // technically, we may have the "rosetta stone" event in memory that could unlock this chain, if we loaded everything and recursed,
+                            // but the immediate prev is not in this set and has not been delivered to the client yet, so they shouldn't have known how to
+                            // construct this event so we'll consider this missing history. This can be used to return an error if the event is required to have history.
+                            missing_history.push((event, header));
+                        }
+                    }
+                }
+            }
+        }
+
+        // We add the events to the deliverable list until nothing changes
+        while let Some((mut event, header)) = prevs_in_memory.pop() {
+            let mut made_changes = false;
+            match header.prev() {
+                None => {
+                    unreachable!("Init events should have been filtered out of the in memory set");
+                }
+                Some(prev) => {
+                    // a hashset would be better loopkup but we're not going to have that many events so hashing
+                    // for a handful of lookups and then convert back to a vec probably isn't worth it.
+                    if deliverable.iter().any(|(e, _)| e.cid() == prev) {
+                        event.body.set_deliverable(true);
+                        deliverable.push((event, header));
+                        made_changes = true;
+                    } else {
+                        prevs_in_memory.push((event, header));
+                    }
+                }
+            }
+            if !made_changes {
+                missing_history.extend(prevs_in_memory);
+                break;
+            }
+        }
+
+        Ok(OrderEvents {
+            deliverable,
+            missing_history,
+        })
+    }
+}
--- a/service/src/event/order_events.rs
+++ b/service/src/event/order_events.rs
@@ -50,8 +50,8 @@ impl OrderEvents {
                     unreachable!("Init events should have been filtered out since they're always deliverable");
                 }
                 Some(prev) => {
-                    if let Some(in_mem) = new_cids.get(&prev) {
-                        if *in_mem {
+                    if let Some(in_mem_is_deliverable) = new_cids.get(&prev) {
+                        if *in_mem_is_deliverable {
                             event.body.set_deliverable(true);
                             *new_cids.get_mut(&event.cid()).expect("CID must exist") = true;
                             deliverable.push((event, header));
--- a/service/src/event/order_events.rs
+++ b/service/src/event/order_events.rs
@@ -1,7 +1,7 @@
 use std::collections::{HashMap, VecDeque};
 
+use ceramic_core::Cid;
 use ceramic_store::{CeramicOneEvent, EventInsertable, SqlitePool};
-use cid::Cid;
 
 use crate::Result;
 
@@ -15,7 +15,15 @@ pub(crate) struct OrderEvents {
 impl OrderEvents {
     /// Groups the events into lists of those with a delivered prev and those without. This can be used to return an error if the event is required to have history.
     /// The events will be marked as deliverable so that they can be passed directly to the store to be persisted.
-    /// Will look up the prev from the database if needed to check if it's deliverable (could possibly change this for recon and allow the ordering task to handle it?)
+    ///
+    /// The job of this function is different than the `ordering_task` module. That module recurses indefinitely attempting to build history. This
+    /// will only traverse a single prev outside of the initial set of events (that is, to the database). This is important because we don't want to
+    /// allow API users to write an event they haven't yet seen the prev for, and for recon we can allow the other task to sort it out.
+    ///
+    /// The `missing_history` set means that the prev is not deliverable or in the `candidate_events` vec with a deliverable prev. For example,
+    /// with new events [C, D] (in any order), if D.prev = C, C.prev = B, then C and D are deliverable if B is in the database as deliverable.
+    /// Given the same situation, where B is not yet deliverable, but B.prev = A and A is deliverable (shouldn't happen, but as an example), we
+    /// *could* mark B deliverable and then C and D, but we DO NOT want to do this here to prevent API users from writing events that they haven't seen.
     pub async fn try_new(
         pool: &SqlitePool,
         mut candidate_events: Vec<(EventInsertable, EventMetadata)>,
@@ -23,7 +31,7 @@ impl OrderEvents {
         let mut new_cids: HashMap<Cid, bool> = HashMap::from_iter(
             candidate_events
                 .iter()
-                .map(|(e, _)| (e.cid(), e.body.deliverable())),
+                .map(|(e, _)| (e.cid(), e.deliverable())),
         );
         let mut deliverable = Vec::with_capacity(candidate_events.len());
         candidate_events.retain(|(e, h)| {
@@ -78,8 +86,10 @@ impl OrderEvents {
         // we may loop through multiple times putting things back in the queue, but it should be a short list
         // and it will shrink every time we move something to the deliverable set, so it should be acceptable.
         // We can't quite get rid of this loop because we may have discovered our prev's prev from the database in the previous pass.
+        let max_iterations = undelivered_prevs_in_memory.len();
+        let mut iteration = 0;
         while let Some((mut event, header)) = undelivered_prevs_in_memory.pop_front() {
-            let mut made_changes = false;
+            iteration += 1;
             match header.prev() {
                 None => {
                     unreachable!("Init events should have been filtered out of the in memory set");
@@ -89,13 +99,14 @@ impl OrderEvents {
                         *new_cids.get_mut(&event.cid()).expect("CID must exist") = true;
                         event.body.set_deliverable(true);
                         deliverable.push((event, header));
-                        made_changes = true;
+                        // reset the iteration count since we made changes. once it doesn't change for a loop through the queue we're done
+                        iteration = 0;
                     } else {
                         undelivered_prevs_in_memory.push_back((event, header));
                     }
                 }
             }
-            if !made_changes {
+            if iteration >= max_iterations {
                 missing_history.extend(undelivered_prevs_in_memory);
                 break;
             }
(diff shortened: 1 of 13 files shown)


service/src/event/ordering_task.rs
```diff
@@ -175,188 +122,310 @@ impl OrderingTask {
...
+                        trace!(
+                            %prev,
+                            %cid,
+                            "Found event whose prev is already in memory but NOT deliverable."
+                        );
+                        // nothing to do until it arrives on the channel
+                    }
+                }
+
+                continue;
```
stbrody, line 331: could we drop the continue and use an else block instead?  I just think it's more clear to read that way

service/src/event/ordering_task.rs
```diff
@@ -175,188 +122,310 @@ impl OrderingTask {
...
+            }
+
+            let prev_event = StreamEvent::load_by_cid(pool, prev).await?;
+            if let Some(known_prev) = prev_event {
+                match &known_prev {
+                    StreamEvent::InitEvent(_) | StreamEvent::KnownDeliverable(_) => {
+                        deliverable_queue.push_back(cid);
+                    }
+                    StreamEvent::Undelivered(undelivered_ev) => {
+                        // we'll try to follow this back to something deliverable
```
stbrody, line 341: where does that happen?

service/src/event/ordering_task.rs
```diff
@@ -175,188 +122,310 @@ impl OrderingTask {
...
+                }
+                self.add_event(known_prev);
+            } else {
+                trace!("Found event that depends on another event we haven't discovered yet");
+            }
+        }
+
+        let mut newly_ready = deliverable_queue.clone();
+        while let Some(cid) = newly_ready.pop_front() {
+            if let Some(now_ready_ev) = self.remove_by_prev_cid(&cid) {
```
stbrody, line 353: I'm confused about how the `prev_map` and `cip_map` are supposed to be maintained.  Based on this line, I would think that those maps are only supposed to store events are are undelivered, because here we are removing from them once they become deliverable.  But in `add_event` above, we add to both of those maps events that are already deliverable.  So do those maps store only undelivered events, or do they store all the events in the stream, or something else?

service/src/event/ordering_task.rs
```diff
@@ -175,188 +122,310 @@ impl OrderingTask {
...
+                if let Some(ev) = self.cid_map.get(&now_ready_ev) {
+                    match ev {
+                        StreamEvent::InitEvent(_) | StreamEvent::KnownDeliverable(_) => {
+                            warn!(?ev, "should not have found a deliverable event when we expected only undelivered events!");
```
stbrody, line 357: same comment as above on if this should be a fatal error since it indicates programmer error.

service/src/event/ordering_task.rs
```diff
@@ -365,112 +434,96 @@ impl OrderingState {
...
+            if undelivered.is_empty() {
                 break;
             } else {
                 // We can start processing and we'll follow the stream history if we have it. In that case, we either arrive
                 // at the beginning and mark them all delivered, or we find a gap and stop processing and leave them in memory.
                 // In this case, we won't discover them until we start running recon with a peer, so maybe we should drop them
                 // or otherwise mark them ignored somehow.
-                self.process_events(pool, None).await?;
-                if new < DELIVERABLE_EVENTS_BATCH_SIZE {
+                let found_all = undelivered.len() < batch_size as usize;
```
stbrody, line 455: why isn't this `undelivered.len() == 0`?  What happens if you have a handful of leftover undelivered events but not enough to fill a batch?

service/src/event/ordering_task.rs
```diff
@@ -365,112 +434,96 @@ impl OrderingState {
...
-        Ok((new, found))
-    }
 
-    fn add_incoming_batch(&mut self, events: Vec<DeliverableEvent>) -> HashSet<InitCid> {
-        let mut updated_streams = HashSet::with_capacity(events.len());
-        for event in events {
-            if let Some(updated_stream) = self.track_pending(event) {
-                updated_streams.insert(updated_stream);
+        if !to_store_asap.is_empty() {
+            info!("storing init events that were somehow missed previously");
```
stbrody, line 502: again, this is part of why I feel like these programming bugs should be fatal, so we don't then have this random code sprinkled around to handle a case we think should never actually happen.

service/src/event/service.rs
```diff
@@ -71,35 +67,60 @@ impl CeramicEventService {
...
         let event_ipld: unvalidated::RawEvent<Ipld> =
             serde_ipld_dagcbor::from_slice(&ev_block.bytes).map_err(|e| {
                 Error::new_invalid_arg(
                     anyhow::anyhow!(e).context("event block is not valid event format"),
                 )
             })?;
 
-        let maybe_init_prev = match event_ipld {
-            unvalidated::RawEvent::Time(t) => Some((t.id(), t.prev())),
+        let (deliverable, header) = match event_ipld {
```
stbrody, line 104: can we re-use `Event::decode_car` (https://github.com/ceramicnetwork/rust-ceramic/blob/5d40086ecb30fe0fb7e1a490ead72c958012c5cd/event/src/unvalidated/event.rs#L75) to simplify the logic here and reduce code duplication?

stbrody (changes requested):
it's still pretty tough to follow the control flow here, though I do think this is better than the previous version.

Left a bunch of comments. Admittedly, a lot of my comments are kind of nit-picky.  It's just that with code this complicated any little bit to improve readability and reduce cognitive overhead of someone trying to read and understand this code is extra valuable I think.

service/src/event/ordering_task.rs
dav1do: we put everything that's "interesting" in them. they are nearly identical sets, except there is no entry for the init event in the prev_map as it has no prev. when we discover events from the db, they are added to the sets. when we hear about events on the channel, they are always added if they're undelivered, and if they're delivered, we ignore them unless we already have the stream in memory.

we then make a copy of all the undelivered events (`cid, prev`) as a queue and iterate through finding the prev event if we can:
- we check in memory using `prev_map` and then fall back to the database.
- if we find a new prev event, we put it in our maps. 
- if it was undelivered, we add it to the queue we're iterating to try to find it's prev. 
- if we find our prev and it's deliverable, we put ourself in the deliverable queue. 

when we call `completed_processing` after all the updates have been committed to the db, we iterate through our cid_map and change everything we just delivered from undelivered to deliverable. if we have nothing left that's undelivered, we're done and dropped from memory, else we stick around and try to reuse the information we previously collected when we discover new events rather than going to disk again.

service/src/event/ordering_task.rs
dav1do: I'm starting to walk this back though. I do think it's worth marking all init events as fatal, because that should never happen... i'm trying to figure out if there's any possibility of changes in the database and our sets getting out of date but all the mutations should be cancel safe, so I don't think so, but crashing the task would be a huge bummer, even though i'm generally a fan of "asserts in production".

service/src/event/ordering_task.rs
dav1do: So I wrote that ☝️ and then looked over it and realized that's not what I did anymore and it was a bit of a mess with the weird removal. So now I do keep the two maps in sync except for the init event not being in the prev map and don't remove anything until the all event for the stream are delivered and we drop it. if we get new deliverable events, we won't load up the old history unless we need it. the `prev_map` is used to find something we've unlocked that is a deliverable candidate. I used more mutable state to modify the `cid_map` and track the queue of newly discovered deliverable events and modified some control flow to hopefully make things clearer/better.

service/src/event/ordering_task.rs
dav1do: a handful of events that don't fill a batch is this `found_all` (now `found_everything`) and we will exit the loop after processing them. I modified it to only have one break at the end rather than two under conditionals.

service/src/event/service.rs
```diff
@@ -46,8 +50,14 @@ impl CeramicEventService {
     pub async fn new(pool: SqlitePool) -> Result<Self> {
         CeramicOneEvent::init_delivered_order(&pool).await?;
 
-        let delivery_task =
-            OrderingTask::run(pool.clone(), PENDING_EVENTS_CHANNEL_DEPTH, true).await;
+        let _updated = OrderingTask::process_all_undelivered_events(
```
dav1do, line 53: this happening inline at startup is new

service/src/event/ordering_task.rs
```diff
@@ -175,302 +97,470 @@ impl OrderingTask {
...
+                self.prev_map.insert(meta.prev, meta.cid);
+                if !self.should_process {
+                    // we depend on something in memory
+                    if let Some(in_memory_prev) = self.cid_map.get(&meta.prev) {
+                        match in_memory_prev {
+                            StreamEvent::InitEvent(_) | StreamEvent::KnownDeliverable(_) => {
+                                self.should_process = true;
+                            }
+                            StreamEvent::Undelivered(_) => {
+                                // nothing to do until it's prev arrives
```
stbrody, line 249:
```suggestion
                                // nothing to do until its prev arrives
```

service/src/event/ordering_task.rs
```diff
@@ -175,302 +97,452 @@ impl OrderingTask {
...
+    fn discovered_deliverable_prev(&mut self, cid: EventCid) {
+        self.new_deliverable.push_back(cid);
+        let event = self
+            .cid_map
+            .get_mut(&cid)
+            .expect("Must have event in cid_map");
+        match event {
+            StreamEvent::InitEvent(cid)
+            | StreamEvent::KnownDeliverable(StreamEventMetadata { cid, .. }) => {
+                unreachable!(
```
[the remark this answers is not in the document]
stbrody: if this is a case that should never happen short of a bug in our code, then I prefer to assert.

That said I don't feel super strongly, maybe @nathanielc can serve as tie-breaker.

service/src/event/ordering_task.rs
```diff
@@ -175,302 +97,470 @@ impl OrderingTask {
...
-                .as_ref()
-                .map_or(false, |to_do| !to_do.contains(cid))
-            {
-                continue;
-            }
-            let deliverable = Self::discover_deliverable_events(pool, stream_events).await?;
-            if !deliverable.is_empty() {
-                self.ready_events.extend(deliverable)
-            }
+    /// Add a stream to the list of streams to process.
```
stbrody, line 410:
```suggestion
    /// Add streams to the list of streams to process.
```

stbrody: looks like the events passed here might not all belong to the same stream right?

service/src/event/service.rs
```diff
@@ -1,27 +1,31 @@
...
 
 use crate::{Error, Result};
 
+/// How many events to select at once to see if they've become deliverable when we have downtime
+/// Used at startup and occassionally in case we ever dropped something
+/// We keep the number small for now as we may need to traverse many prevs for each one of these and load them into memory.
+const DELIVERABLE_EVENTS_BATCH_SIZE: u32 = 1000;
+
+/// How many batches of undelivered events are we willing to process on start up?
+/// To avoid an infinite loop. It's going to take a long time to process `DELIVERABLE_EVENTS_BATCH_SIZE * MAX_ITERATIONS` events
```
stbrody, line 24: this means we can process up to 1 billion undelivered events at startup.

Should we add an upfront check to fail fast if the count of undelivered events is more than this when the node starts up?

service/src/event/ordering_task.rs
```diff
@@ -175,302 +97,470 @@ impl OrderingTask {
...
 
-    fn add_incoming_batch(&mut self, events: Vec<DeliverableEvent>) -> HashSet<InitCid> {
-        let mut updated_streams = HashSet::with_capacity(events.len());
-        for event in events {
-            if let Some(updated_stream) = self.track_pending(event) {
-                updated_streams.insert(updated_stream);
+            event_cnt += 1;
+            self.add_stream_event(*stream_cid, loaded);
+        }
+        // while undelivered init events should be unreachable, we can fix the state if it happens so we won't panic in release mode
```
stbrody, line 538: this seems reasonable to me. I think its okay to do a little more at startup to try to recover from an unexpected condition, but continue to bail if something impossible happens at runtime.

service/src/event/ordering_task.rs
```diff
@@ -175,302 +97,470 @@ impl OrderingTask {
...
-            self.prev_map.remove(&cid.meta.prev);
-            Some(cid)
-        } else {
-            None
+    /// Called when we've persisted the deliverable events to the database and can clean up our state.
+    /// Returns `true` if we're finished processing and can be dropped from memory.
+    /// Returns `false` if we have more work to do and should be retained for future processing
+    fn processing_completed(&mut self) -> bool {
+        self.should_process = false;
+        self.new_deliverable.clear();
```
stbrody, line 274: should we remove entries from `cid_map` and `prev_map` based on what was in new_deliverable?  No need to keep maintaining in-memory state for events that were fully delivered.

It's theoretically possible a stream could keep having new events marked deliverable without ever having *all* of its events deliverable all at once so the state never gets dropped and the amount of data being stored related to old already delivered events just grows and grows, but I think we can do something here to prune what we're keeping in memory and avoid that.

stbrody (approved):
LGTM.  A few small cleanup suggestions, but no need for another round of review from me.

I do still think it would be nice to get @nathanielc to take a look at this also though so we have some more eyes on it and more people familiar with how this works.

service/src/event/ordering_task.rs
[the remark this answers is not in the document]
dav1do: I decided to go with the "crash if it's related to our internal state being mismanaged" and "crash in debug mode" if it's unexpected but somewhat out of our control (e.g.. we somehow find undelivered init events when starting up). This _should_ never happen, but it's really easy for us to fix when we find it so crashing and making someone do it manually seems kind of mean.. but is more likely to make sure it never happens. I'd be more comfortable increasing this restriction after things have been deployed for a bit and we have a better sense of the current state of the database.

service/src/event/order_events.rs
```diff
@@ -0,0 +1,100 @@
...
+                        deliverable.push((event, header));
+                        made_changes = true;
+                    } else {
+                        prevs_in_memory.push((event, header));
+                    }
+                }
+            }
+            if !made_changes {
+                missing_history.extend(prevs_in_memory);
+                break;
```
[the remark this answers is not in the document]
dav1do: I mean unstable rustc feature and fails to compile

```
error[E0658]: `let` expressions in this position are unstable
  --> service/src/event/order_events.rs:92:16
   |
92 |             && let Some((mut event, header)) = undelivered_prevs_in_memory.pop_front()
   |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: see issue #53667 <https://github.com/rust-lang/rust/issues/53667> for more information

For more information about this error, try `rustc --explain E0658`.
error: could not compile `ceramic-service` (lib) due to 1 previous error
```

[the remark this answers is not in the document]
stbrody: ah I see, bummer

service/src/event/service.rs
dav1do: It'd might be good to log a warning, I was considering whether adding some time estimation log based on how many undelivered events exist would be useful but decided against it. Sqlite does a full table scan to get a count result, so hesitated to add the overhead rather than just make an attempt and inform the user with some messages as we go.

I don't think it's worth failing, because we could process 1 billion and they could restart and process 1 billion more... and if we actually get to this situation things are in a really bad state since we'd need to have like a TB of unordered events and we're probably being spammed somehow.

Anyway, I'm open to making a change or adding a ticket to come back to it, generally, I think making a ticket about startup processing (migrations, ordering, build redtree, etc) might be worthwhile and we can make sure we're offloading what is reasonable, blocking what we must, and being informative about what we're doing.

service/src/event/ordering_task.rs
```diff
@@ -1,159 +1,81 @@
...
+type StreamCid = Cid;
+type EventCid = Cid;
+type PrevCid = Cid;
 
 #[derive(Debug)]
 pub struct DeliverableTask {
     pub(crate) _handle: tokio::task::JoinHandle<()>,
-    pub(crate) tx: tokio::sync::mpsc::Sender<DeliverableEvent>,
-    pub(crate) tx_new: tokio::sync::mpsc::Sender<DeliveredEvent>,
+    /// Currently only receives events discovered over recon that are out of order and need to be marked ready (deliverable)
```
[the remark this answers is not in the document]
dav1do: bah, fixed. I used to have two channels and now it's just one. simplified the doc as I don't think it added anything useful

service/src/event/ordering_task.rs
dav1do: yeah, that's correct. modified to read

```
    /// Update the list of streams to process with new events.
    /// Relies on `add_stream_event` to handle updating the internal state.
```

service/src/event/ordering_task.rs
dav1do: I don't want to make this change now but added this comment.

```
        // It's worth considering whether we should prune the `cid_map` and `prev_map` with everything from the `new_deliverable` set to avoid
        // memory leaks due to never finding events. This requires setting `self.should_process` to true for all undelivered events (or making a db query),
        // as we will no longer have everything we've discovered in memory and need to go to the db to figure out what to do. As streams are generally
        // short, I'm opting to not make this change now. We could consider something like an expiration time as well, but generally we don't want
        // to be dropping things from memory that we might need to process in the future unless we have to.
```

service/src/event/service.rs
```diff
@@ -197,225 +176,94 @@ impl CeramicEventService {
...
+                    let discovered = DiscoveredEvent {
+                        cid: ev.cid(),
+                        known_deliverable: ev.deliverable(),
+                        metadata: metadata.to_owned(),
+                    };
+                    trace!(?discovered, "sending delivered to ordering task");
+                    if let Err(e) = self.delivery_task.tx_inserted.try_send(discovered) {
+                        match e {
+                            tokio::sync::mpsc::error::TrySendError::Full(e) => {
+                                warn!(event=?e, limit=%PENDING_EVENTS_CHANNEL_DEPTH, "Delivery task full. Dropping event and will not be able to mark deliverable new stream event arrives or process is restarted");
```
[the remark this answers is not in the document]
dav1do: yeah, it might be worth just using send.. I increased the channel depth to 1 million (304 bytes each * 1M events ~= 3MB of data) for now which would mean we're way behind on processing...

[ticket AES-176](https://linear.app/3boxlabs/issue/AES-176/apply-backpressure-when-discovering-events-faster-than-order-task-can): 

```
Context: Currently, when we accept a new event and it's out of order, we send it to another task to process using a tokio mpsc channel and the try_send function. This fails immediately if the channel is full. We should consider whether blocking until it's available with send or for a time period using send_timeout is preferable. This would consume a task and may slow down writing new events while that's happening. 

We could also notify the task that we were unable to tell it of events at the time and store them in memory to retry when it's less congested, or tell it to seek them out in the database like it does at start up. 

We currently allow 1 million events in the channel buffer (3 MB of RAM), so we'd have to be way behind to encounter this.```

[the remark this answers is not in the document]
Comment (dav1do) [reply service/src/event/service.rs:245]: actually, I just switched to send because slowing down is probably fine if we're that behind on pulling things out of the buffer. I didn't want to block writes before but this is probably a better default for now. I flipped the ticket to consider whether we should use try_send or send_timeout instead.

File: service/src/event/service.rs (hunk 12 above)
Comment (dav1do) [reply service/src/event/service.rs:24]: created AES-177

File: service/src/event/ordering_task.rs (hunk 14 above)
Comment (stbrody) [reply service/src/event/ordering_task.rs:278]: I don't understand that comment, that seems way more complicated than I was imagining.  Why do we need to go back to the database?  Can't we just loop once over `new_deliverable` and if any of those events are keys in `cid_map` or values in `prev_map`, just remove them from the map?

Comment (stbrody) [reply service/src/event/ordering_task.rs:278]: Also the memory leak isn't due to "never finding events".  Even if you always find all events eventually you can wind up with a memory leak if by the time you've filled one hole a new hole has always shown up so that the full `StreamEvents` struct never gets dropped.

For example imagine a stream where you get delivered events in the following pattern:
1, 3, 5, 2, 7, 4, 9, 6, 11, etc.  You never have a moment where you have all the events in the stream without any holes, but you also do get every event eventually and there is always a point in the stream after which you have received everything and there are no holes earlier than that point, which should allow you to clean up any state associated with those earlier events that are before the point where holes remain.

Comment (dav1do) [reply service/src/event/ordering_task.rs:278]: Yes, we can do that and it works fine as long as we make sure to be aggressive about looking up history for the messages we're discovering. The problem is that messages can sit in the channel so there's a period after they are checked for being deliverable initially and when they get reviewed, which can cause things to change. Currently, by keeping the history for partially processed streams, we're able to see what's happened while we've been in the queue. 

But, this isn't totally true. Even though I [just changed this](https://github.com/ceramicnetwork/rust-ceramic/pull/387/commits/e4f5a6e8d008afce02ebb609244a07dd6d700ae5) to try to avoid extra work, it might be better to set for every undelivered event, since I'm imagining a forking scenario when the queue is deep that results in our history being updated and dropped, then some new fragment we don't depend on is in memory. In that case, relying on the recent changes in memory won't get us reviewed, so it seems better to do extra work and make sure we don't miss anything for now.

Anyway, with the change to always make sure to process every undelivered event, pruning works fine by looping through and removing things, but it will result in more database lookups since we'll be dropping things we're about to need, especially when the queue is deeper.

Comment (dav1do) [reply service/src/event/ordering_task.rs]: We talked, was so much easier 😅.. made the change in [this commit](https://github.com/ceramicnetwork/rust-ceramic/pull/387/commits/1c0ca5fa35121af4f8dd773dbf5f91763ca3a4ec).

File: service/src/event/ordering_task.rs [hunk 18]
```diff
@@ -175,302 +98,506 @@ impl OrderingTask {
...
+                self.update_should_process_for_new_delivered(&meta.cid);
+                meta.cid
+            }
+            StreamEvent::Undelivered(meta) => {
+                self.prev_map.insert(meta.prev, meta.cid);
+                if !self.should_process {
+                    // as we could sit in the channel for a while and streams can fork, we can't rely
+                    // on our in memory state to know what happened between enqueue and dequeque
+                    // and need to try to process. We could add an LRU cache in the future to
+                    // avoid looking up that were processed while were were enqueued.
```
stbrody, line 248:
```suggestion
                    // avoid looking up events that were processed while were were enqueued.
```

service/src/event/service.rs
```diff
@@ -197,225 +177,88 @@ impl CeramicEventService {
...
-        to_check: Vec<(DeliverableMetadata, EventInsertable)>,
-    ) -> Result<()> {
-        if to_check.is_empty() {
-            return Ok(());
+        // api writes shouldn't have any missed pieces that need ordering so we don't send those and return early
+        if history_required {
+            return Ok(InsertResult {
+                store_result: res,
+                missing_history,
+            });
```
nathanielc, line 227: nit, can you remove this early return and guard the discovered logic with `!history_required`?

service/src/event/service.rs
```diff
@@ -197,225 +177,88 @@ impl CeramicEventService {
...
+            .map(|(e, _)| e.clone())
+            .collect::<Vec<_>>();
 
-    fn mark_event_deliverable_now(&mut self, mut ev: EventInsertable, init_cid: Cid) {
-        ev.set_deliverable(true);
-        self.notify_task_new
-            .push(DeliveredEvent::new(ev.body.cid, init_cid));
-        self.insert_now.push(ev);
-    }
+        let res = CeramicOneEvent::insert_many(&self.pool, &to_insert[..]).await?;
```
nathanielc, line 220: In reading the code it _seems_ like if we change this method to accept an iterator instead of a slice we do not have to do as much cloning of the data. For example we could keep the to_insert vector as a tuple of the insertable and the metadata and use an iter to hide the metadata. This would avoid having the map the insertable back to its metadata below.

service/src/event/service.rs
```diff
@@ -101,53 +111,34 @@ impl CeramicEventService {
...
+            return Err(Error::new_app(anyhow::anyhow!(
+                "EventId CID ({}) does not match the body CID ({})",
+                event_cid,
+                cid
+            )));
+        }
+
+        let metadata = EventMetadata::from(parsed_event);
+        let mut body = EventInsertableBody::try_from_carfile(cid, carfile).await?;
+        body.set_deliverable(matches!(metadata, EventMetadata::Init { .. }));
```
nathanielc, line 139: Why is this logic here instead of part of the event ordering bits? While true any init event is deliverable it seems like this logic should live with the other logic that knows that an init event doesn't have a prev.

service/src/event/ordering_task.rs
```diff
@@ -175,302 +98,506 @@ impl OrderingTask {
...
-    fn remove_by_prev_cid(&mut self, cid: &Cid) -> Option<EventCid> {
-        if let Some(cid) = self.prev_map.remove(cid) {
-            self.cid_map.remove(&cid);
-            Some(cid)
+    /// Called when we've persisted the deliverable events to the database and can clean up our state.
+    /// Returns `true` if we're finished processing and can be dropped from memory.
+    /// Returns `false` if we have more work to do and should be retained for future processing
+    fn processing_completed(&mut self) -> bool {
+        // if we're done, we don't need to bother cleaning up since we get dropped
+        if !self
```
nathanielc, line 271: nit, instead of `!...any(Undelivered)` use `..all(!Undelivered)`.

nathanielc (commented):
LGTM. Its merged but I do have a few places where code readability could be improved. Not urgent but thought I'd share.

service/src/event/service.rs
stbrody: I'm pretty sure I asked for the exact opposite in an earlier review 😅.

Can I ask why you don't like the early return?  Personally I'm a big fan of early returns, I like removing the extra indentation layer and I think it helps reduce cognitive load by letting me see that this case is over and dealt with and I don't need to think about it any more, vs being in a big if block where there's some overhead to remembering what block I'm in and what the condition in play is right now.

nathanielc: I generally dislike early returns as they make it harder to determine the different code branches as they can be easy to miss. In this case both returns, return the same value so they were really the same code branch just duplicated.   

Not a strong preference here either way.

stbrody: fascinating!  I've never heard an argument against early returns before.  I often push for early returns when I do code reviews because I've always felt they improve code readability.  I always thought that opinion was non-controversial 😅

Good to know it's not as clear-cut as I thought.  I still like them but don't feel strongly either way.  Maybe I'll stop pushing as hard for others to use them going forward

dav1do: I also don't like early returns unless they're right away. At least in rust I prefer to rely on the fact that it's an expression and returns something. 

That said, I think I can extract the big if block so it's easy to understand (and hopefully satisfies everyone 😁).

service/src/event/service.rs
```diff
@@ -197,225 +177,88 @@ impl CeramicEventService {
...
-        insertable: EventInsertable,
-        meta: DeliverableMetadata,
-    ) {
-        self.background_task_deliverable
-            .insert(insertable.body.cid, meta);
-        self.insert_now.push(insertable);
-    }
+        let to_insert = to_insert_with_metadata
+            .iter()
+            .map(|(e, _)| e.clone())
```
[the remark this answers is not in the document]
dav1do: removed by refactoring

service/src/event/service.rs
dav1do: yeah, I thought that too but hesitated to add another refactor. I did it in the new branch and it removes some allocations (the clone above as well as collects)

service/src/event/service.rs
dav1do: I moved it to the Ordering task. Originally I thought it might be less likely to ever be missed if it just happened right away but I think it makes sense in the new location.

service/src/event/service.rs
```diff
@@ -197,225 +177,88 @@ impl CeramicEventService {
...
-    }
+        let to_insert_with_metadata = if history_required {
+            ordered.deliverable
+        } else {
+            ordered
+                .deliverable
+                .into_iter()
+                .chain(ordered.missing_history)
+                .collect()
+        };
```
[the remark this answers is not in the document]
dav1do: yeah, I wanted to avoid the database twice, but I adjusted things with the change to iterator that avoids needing to allocate

PR #391, opened 2024-06-18
chore: replace tracing_test with test-log

dav1do:
Running with tokio-console on could cause subscribers to register and conflict on the port so now we don't init tracing manually and instead rely on `test-log`. `tracing_test` is only to assert log statements, not get debug tests like we want so it was replaced.

Requires running with `--nocapture` i.e. `RUST_LOG=info cargo test -- --nocapture`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1045,10 +1045,10 @@ dependencies = [
  "serde",
  "serde_ipld_dagcbor",
  "swagger",
+ "test-log",
  "tikv-jemalloc-ctl",
  "tokio",
  "tracing",
- "tracing-test",
 ]
 
 [[package]]
@@ -1165,10 +1165,10 @@ dependencies = [
  "serde_ipld_dagjson",
  "serde_json",
  "swagger",
+ "test-log",
  "thiserror",
  "tokio",
  "tracing",
- "tracing-test",
 ]
 
 [[package]]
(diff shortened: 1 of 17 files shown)


nathanielc (approved):
LGTM, literally made the same refactor in #392 however this one catch a few bits I missed.

Adding a useful comment from #392 here so its easier to find

With this change all tests now use the test-log::test method. Here is a quick summary of how this method, and now therefore all tests, behave:

* By default no tracing logs are output
* Use env var RUST_LOG, i.e. RUST_LOG=debug, to enable logging in tests. Note only failing tests will print their logs, as this is the default cargo test behavior.
* Use cargo test -- --show-output to print logs from passing tests. This is because by default cargo test suppresses all logs from passing tests.
* Use cargo test -- --nocapture to print logs from passing or failing tests as they are printed without any buffering.

stbrody:
> With this change all tests now use the test-log::test method. Here is a quick summary of how this method, and now therefore all tests, behave:
> 
> * By default no tracing logs are output
> * Use env var RUST_LOG, i.e. RUST_LOG=debug, to enable logging in tests. Note only failing tests will print their logs, as this is the default cargo test behavior.
> * Use cargo test -- --show-output to print logs from passing tests. This is because by default cargo test suppresses all logs from passing tests.
> * Use cargo test -- --nocapture to print logs from passing or failing tests as they are printed without any buffering.

this comment is super useful.  Can we capture it somewhere for future reference?  In the ceramic-one help output, or at least in the README?

nathanielc:
@stbrody https://github.com/ceramicnetwork/rust-ceramic/pull/394

PR #392, opened 2024-06-18
refactor: use test-log::test for all testing log needs

nathanielc:
Prior to this change we had three different ways of enabling logs/tracing inside of tests. Using tracing-test, using test-log, and a manual setup using init_local_tracing. All of these methods had the same goal, to be able to debug tests using tracing::debug etc macros. All accomplished the goal slightly differently leading to confusion when working with across multiple tests.

With this change all tests now use the test-log::test method. Here is a quick summary of how this method, and now therefore all tests, behave:

* By default no tracing logs are output
* Use env var `RUST_LOG`, i.e. `RUST_LOG=debug`, to enable logging in tests. Note only failing tests will print their logs, as this is the default `cargo test` behavior.
* Use `cargo test -- --show-output` to print logs from passing tests. This is because by default `cargo test` suppresses all logs from passing tests.
* Use `cargo test -- --nocapture` to print logs from passing or failing tests as they are printed without any buffering.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1045,10 +1045,10 @@ dependencies = [
  "serde",
  "serde_ipld_dagcbor",
  "swagger",
+ "test-log",
  "tikv-jemalloc-ctl",
  "tokio",
  "tracing",
- "tracing-test",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


nathanielc:
Closing in favor of #391

PR #393, opened 2024-06-19
feat: access control stubs

dbcfd:
add api stubs for supporting access control behind a feature flag

### DIFF
--- a/api/ceramic.yaml
+++ b/api/ceramic.yaml
@@ -25,7 +25,7 @@ paths:
           required: true
           schema:
             type: string
-        - name: Did
+        - name: X-Did
           in: header
           description: DID used to sign the bearer token
           required: true
--- a/api/ceramic.yaml
+++ b/api/ceramic.yaml
@@ -15,6 +15,60 @@ servers:
   - url: /ceramic
 
 paths:
+  /auth/token/refresh:
+    post:
+      summary: Create a refresh token
+      parameters:
+        - name: Authorization
+          in: header
+          description: Bearer token
+          required: true
+          schema:
+            type: string
+        - name: Did
+          in: header
+          description: DID used to sign the bearer token
+          required: true
+          schema:
+            type: string
+      requestBody:
+        $ref: "#/components/requestBodies/RefreshTokenRequest"
+      responses:
+        "204":
+          description: success
+        "500":
+          description: Internal server error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorResponse"
+  /auth/token/access:
+    post:
+      summary: Create an access token
+      parameters:
+        - name: Authorization
+          in: header
+          description: Bearer token
+          required: true
+          schema:
+            type: string
+        - name: Did
+          in: header
+          description: DID used to sign the bearer token
+          required: true
+          schema:
+            type: string
+      requestBody:
+        $ref: "#/components/requestBodies/AccessTokenRequest"
+      responses:
+        "204":
+          description: success
+        "500":
+          description: Internal server error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorResponse"
   /liveness:
     get:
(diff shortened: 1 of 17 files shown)


nathanielc:
@dbcfd Do you have any design docs explaining these API endpoints?

api/ceramic.yaml
```diff
@@ -495,3 +575,17 @@ components:
         streamId:
           type: string
           description: Multibase encoded stream ID.
+    RefreshTokenRequest:
+      title: A request for a refresh token
+      type: object
+      properties:
```
dav1do, line 581: should the `requestId` property be required on this and AccessTokenRequest?

api/ceramic.yaml
```diff
@@ -15,6 +15,60 @@ servers:
...
+    post:
+      summary: Create a refresh token
+      parameters:
+        - name: Authorization
+          in: header
+          description: Bearer token
+          required: true
+          schema:
+            type: string
+        - name: Did
```
dav1do, line 28: should this header be `X-Did` or similar throught? (or X-Did... something that's not so easily conflated with the word `did` 🤷). have a custom header called `Did` seems unusual

api/src/server.rs
```diff
@@ -312,6 +320,27 @@ where
             model,
             insert_task,
             marker: PhantomData,
+            feature_flags: HashSet::default(),
+        }
+    }
+
+    pub fn enable_feature(&mut self, feature: &str) {
```
dav1do, line 327: longer term I image feature flags living inside the service crate or similar so they can be shared by recon and other parts of the system but it's implementation details and fine for now

dav1do (commented):
I don't know if I see the point in adding routes that just panic 😂... but if we have a follow up to add the details I don't really have concerns with POC refresh/access token routes and accepting optional headers to validate.. minor suggestions to use the custom header `X-` prefix and maybe making properties required on the output

api/ceramic.yaml
dbcfd: At this point, I don't have enough information to really put anything in this structure, so would like to leave it empty, but openapi doesn't allow that. So I added requestId as an optional parameter, for tracing, even though that might be better as a header.

I feel like I might need list of resources and actions for a refresh token, but wasn't confident enough to add them in this PR.

api/ceramic.yaml
dbcfd: X-Did is probably better, will change.

api/src/server.rs
dbcfd: Yeah, there's better crates for feature flags, we may want to do a quick survey and find one that meets our needs before this or other features are done.

one/src/lib.rs
```diff
@@ -149,6 +149,14 @@ struct DaemonOpts {
...
     )]
     idle_conns_timeout_ms: u64,
+
+    /// Enable authentication feature
+    #[arg(
+        long,
+        default_value_t = false,
+        env = "CERAMIC_ONE_ENABLE_AUTHENTICATION"
+    )]
+    enable_authentication: bool,
```
dav1do, line 159: didn't think about it first pass, but wonder if we'd want to make features flags and enum we could pass multiple times or as a comma separated string... A bool seems find for now, though we should probably mark these things unstable before GA.

dav1do (approved):
Okay from my perspective, the only concern I guess is removing flags from the CLI but as long as we make more progress on this before GA it's not an issue, plus I doubt anyone will use the flags we don't suggest as there are so many.

nathanielc (commented):
Can you help me understand why we want these stubs in place before implementing them? Are you looking for feedback on the design of the API?

I agree with @dav1do we should not change the CLI flags yet until we know what/how they will be used.

We are also working to publish the generated clients for this API. As such having these partially defined endpoints means confusion for end users about the failing endpoints and future breaking changes when we do change them to be complete.

PR #394, opened 2024-06-19
docs: add cargo testing docs

nathanielc:


README.md
```diff
@@ -59,6 +59,29 @@ Using the makefile is not necessary during your development cycle, feel free to
...
+The above `make` targets test changes as a whole.
+In order to test specific changes use the Rust `cargo` tooling directly.
+
+To run all tests via `cargo` use:
+
+    cargo test
+
+To run tests for a single crate in the workspace use:
+
+    cargo test -p ceramic-event
```
stbrody, line 73: nitpick, but might not be obvious to someone that `ceramic-event` is an example.  It should probably say `<crate name>` instead, possibly with a second example showing a more concrete example with an actual create name like `ceramic-event`

stbrody (approved):

PR #395, opened 2024-06-21
fix: address edge cases in from-ipfs time events

nathanielc:
There were two edge cases not covered with the time event migration code:

1. The time event previous may be either a signed or unsigned event.
2. The witness nodes may contain null entries.

Additionally fixes a bug with initializing tracing for the migration CLI cmd and removes resulting dead code.

### DIFF
--- a/event/src/unvalidated/payload/init.rs
+++ b/event/src/unvalidated/payload/init.rs
@@ -4,7 +4,7 @@ use iroh_car::{CarHeader, CarWriter};
 use serde::{Deserialize, Serialize};
 
 /// Payload of an init event
-#[derive(Serialize, Deserialize)]
+#[derive(Debug, Serialize, Deserialize)]
 pub struct Payload<D> {
     header: Header,
     data: Option<D>,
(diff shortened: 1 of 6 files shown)


event/src/unvalidated/payload/init.rs
```diff
@@ -28,6 +28,11 @@ impl<D> Payload<D> {
 }
 
 impl<D: serde::Serialize> Payload<D> {
+    /// Compute CID of encoded event.
+    pub async fn encoded_cid(&self) -> Result<Cid, anyhow::Error> {
+        let event = serde_ipld_dagcbor::to_vec(self)?;
```
stbrody, line 33: I'm always confused by the ipld library regarding this - is this taking the Payload struct and trying to encode it as an ipld array/vector?  Cause seems like that would produce a different CID than if you were encoding it as an IPLD object.

If that isn't what this is doing... then why is the function called to_vec?

stbrody (approved):

event/src/unvalidated/payload/init.rs
nathanielc: Its `to_vec` as in to a Rust vector of bytes as opposed to the function `to_writer` which encodes the data into something that implements the Writer trait. 

It has nothing to do with the input type as that is an opaque type that implements the Serialize trait.  

https://docs.rs/serde_ipld_dagcbor/latest/serde_ipld_dagcbor/fn.to_vec.html

Hope that helps clarify.

stbrody: ah yes that does help.  I do kind of wish it was called to_bytes or something like that, but oh well

PR #398, opened 2024-06-26
feat: add query parameter to return event data on feed endpoint

dav1do:
There is now an enum parameter called `includeData` that can be `none` or `full`. I originally had a boolean, so it's not the best name and open to suggestions. I also changed the data field to return null instead of an empty string to avoid crashing multibase decoding (need to verify js-ceramic is okay with this but it doesn't use the value currently).

Example usage: `curl -v --location 'http://127.0.0.1:5001/ceramic/feed/events?includeData=full&limit=1` and `curl -v --location 'http://127.0.0.1:5001/ceramic/feed/events?limit=1`

```json
{
  "events": [
    {
      "id": "bafyreia2ihx26sxlukdpnjzkpkyi4flmyxrwq4ritcbwxlxxyf2qltiyhq",
      "data": "mOqJlcm9vdHOB2CpYJQABcRIgGkHvr0rroob2pyp6sI4VbMXjaHIomINrrvfBdQXNGDxndmVyc2lvbgHLAQFxEiAaQe+vSuuihvanKnqwjhVsxeNociiYg2uu98F1Bc0YPKJkZGF0YaJhYQFhYgJmaGVhZGVypGNzZXBlbW9kZWxlbW9kZWxYJc4BAhIg0yo2H8npd2cKQSdTnW/+biSFW79zdPyVWE4sJu0icYNmdW5pcXVlTJsqdYUc+GBaeYDfgmtjb250cm9sbGVyc4F4OGRpZDprZXk6ejZNa29GVXBwY0tFVllUUzhvVmlkcmphOTRVb0pUYXROaG5oeEpSS0Y3TllQU2NT"
    }
  ],
  "resumeToken": "2"
}

{
  "events": [
    {
      "id": "bafyreia2ihx26sxlukdpnjzkpkyi4flmyxrwq4ritcbwxlxxyf2qltiyhq"
    }
  ],
  "resumeToken": "2"
}


```

I made some refactors simultaneously. First I modified the store to return an unvalidated event, which requires parsing the bytes to an event and then back to bytes, so a bit unnecessary but it (again) verifies it's valid and makes getting the CID easy. Could change the queries instead to read from disk but cbor should be fast to parse. I also moved the related queries for delivered/undelivered together and made sure the resume/limit params are in the same order. The API trait response for feed is now a struct and there were a handful of updates to tests to test the new option.

### DIFF
--- a/api/ceramic.yaml
+++ b/api/ceramic.yaml
@@ -405,7 +405,7 @@ paths:
             Whether to include the event data (carfile) in the response.
             In the future, only the payload or other options may be supported:
               * `none` - Empty, only the event ID is returned
-              * `full` - The entire envelope carfile (including the envelope and payload)
+              * `full` - The entire event carfile (including the envelope and payload)
       responses:
         "200":
           description: success
--- a/api/ceramic.yaml
+++ b/api/ceramic.yaml
@@ -391,10 +391,21 @@ paths:
           required: false
         - name: limit
           in: query
-          description: the maximum number of events to return, default is 10000.
+          description: The maximum number of events to return, default is 100. The max with data is 10000.
           required: false
           schema:
             type: integer
+        - name: includeData
+          in: query
+          required: false
+          schema:
+            type: string
+            enum: [none, full]
+          description: >
+            Whether to include the event data (carfile) in the response.
+            In the future, only the payload or other options may be supported:
+              * `none` - Empty, only the event ID is returned
+              * `full` - The entire envelope carfile (including the envelope and payload)
       responses:
         "200":
           description: success
--- a/api/ceramic.yaml
+++ b/api/ceramic.yaml
@@ -505,7 +505,6 @@ components:
       type: object
       required:
         - id
-        - data
       properties:
         id:
           type: string
--- a/api-server/README.md
+++ b/api-server/README.md
@@ -15,7 +15,7 @@ To see how to make this your own, look here:
 [README]((https://openapi-generator.tech))
 
 - API version: 0.30.0
-- Build date: 2024-07-29T11:20:12.699785-06:00[America/Denver]
+- Build date: 2024-07-30T10:27:12.329754-06:00[America/Denver]
 
 
 
--- a/api-server/README.md
+++ b/api-server/README.md
@@ -15,7 +15,7 @@ To see how to make this your own, look here:
 [README]((https://openapi-generator.tech))
 
 - API version: 0.30.0
-- Build date: 2024-07-30T10:27:24.718269-06:00[America/Denver]
+- Build date: 2024-07-30T18:32:38.681994-06:00[America/Denver]
 
 
 
--- a/api-server/README.md
+++ b/api-server/README.md
@@ -15,7 +15,7 @@ To see how to make this your own, look here:
 [README]((https://openapi-generator.tech))
 
 - API version: 0.30.0
-- Build date: 2024-07-30T10:27:12.329754-06:00[America/Denver]
+- Build date: 2024-07-30T10:27:24.718269-06:00[America/Denver]
 
 
 
--- a/api-server/api/openapi.yaml
+++ b/api-server/api/openapi.yaml
@@ -421,13 +421,28 @@ paths:
         schema:
           type: string
         style: form
-      - description: "the maximum number of events to return, default is 10000."
+      - description: "The maximum number of events to return, default is 100. The\
+          \ max with data is 10000."
         explode: true
         in: query
         name: limit
         required: false
         schema:
           type: integer
+        style: form
+      - description: |
+          Whether to include the event data (carfile) in the response. In the future, only the payload or other options may be supported:
+            * `none` - Empty, only the event ID is returned
+            * `full` - The entire envelope carfile (including the envelope and payload)
+        explode: true
+        in: query
+        name: includeData
+        required: false
+        schema:
+          enum:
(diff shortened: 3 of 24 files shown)


PaulLeCam:
Could we have the data only be the payload rather than the full envelope please? Assuming C1 does the signature validation, I think having the payload directly would be simpler and lower the amount of data that needs to be sent.

oed:
Ideally we would have the ability to do both?

PaulLeCam:
> Ideally we would have the ability to do both?

Yes I thought that was what was suggested yesterday.
Assuming the envelope is the DAG-JWS as handled by Ceramic, it shouldn't be a problem to handle it for the JS SDK, but I expect for other implementations it would be simpler to not require to do the signature verification.

dav1do:
> > Ideally we would have the ability to do both?
> 
> Yes I thought that was what was suggested yesterday. Assuming the envelope is the DAG-JWS as handled by Ceramic, it shouldn't be a problem to handle it for the JS SDK, but I expect for other implementations it would be simpler to not require to do the signature verification.

Added a new `payload` enum value that encodes the signed event payload as a carfile without the envelope or capabilities.

api/ceramic.yaml
```diff
@@ -293,10 +293,22 @@ paths:
...
+          in: query
+          required: false
+          schema:
+            type: string
+            enum: [none, payload, full]
+          description: >
+            Whether to include the event data (carfile) in the response.
+            In the future, only the payload or other options may be supported:
+              * `none` - Empty, only the event ID is returned
+              * `payload` - Only the signed event payload data as a carfile. Unsigned and time events are unchanged.
```
stbrody, line 310: probably what we actually want for time events here is literally just the timestamp and prev and nothing else.  In the same way that the agg library will assume ceramic-one is doing signature validation, it can also assume that ceramic-one has verified the time event, and so the only actual information that the time event conveys up to the consumer is the new timestamp.

dav1do: hmm, I was trying to stick to the protocol types, is there something that exists that conveys that? we have the event and the proof, but it seems like you're suggesting part of both in something new or am I misunderstanding? I'm definitely open to returning something else if it makes sense.

```rust
pub struct RawTimeEvent {
    id: Cid,
    prev: Cid,
    proof: Cid,
    path: String,
}

pub struct Proof {
    chain_id: String,
    root: Cid,
    tx_hash: Cid,
    tx_type: String,
}
```

stbrody: yeah it would be something new.  Problem is that right now ceramic-one doesn't even have the timestamp anywhere since it isn't stored in the event data itself, it comes from looking up the tx_hash against the actual blockchain.  So we won't be able to return the timestamp until ceramic-one is doing event validation for time events.  But maybe we could still put the structure in place and just return a placeholder timestamp for now?

dav1do: I'm willing to do that but would appreciate if @PaulLeCam would chime in. 

Maybe it's worth chatting about event validation? These payloads don't have to stick directly to the protocol for consumers and we can provide a higher level construct that is useful. For now, the signed event payload is just implied as validated, and this feels like it would be a `validated::TimeEvent`. Since we have the `full` option, anyone is free to do it all themselves if they don't like what we provide via the `payload` parameter.

PaulLeCam: As Spencer wrote I think it would be useful eventually to have the timestamp provided for TimeEvents, one thing it needs to keep having though is the `id` to identify the stream it belongs to.

I agree though that it would be good to define what event validation will provide and possibly start defining the higher-level structures that ceramic-one could expose.

stbrody:
I suspect we're going to want to re-work this based on the conversations we've been having about the "conclusion" API?

@dav1do @nathanielc - does it still make sense to try to get this PR in now and then iterate, or should we close this and put up a new one once the design of the new API is more fully fleshed-out?

dav1do:
> I suspect we're going to want to re-work this based on the conversations we've been having about the "conclusion" API?
> 
> @dav1do @nathanielc - does it still make sense to try to get this PR in now and then iterate, or should we close this and put up a new one once the design of the new API is more fully fleshed-out?

I'm open.. it probably doesn't make sense to include the `payload` option now though. If we do it's just going to be the same for time/init events... I think adding the data option is still useful so I can rebase if others agree.

nathanielc:
> I'm open.. it probably doesn't make sense to include the payload option now though. If we do it's just going to be the same for time/init events... I think adding the data option is still useful so I can rebase if others agree.

Agreed this API can/should remain a way to get at raw event data and not much else.

api/ceramic.yaml
dav1do: Going to defer this conversation to the conclusion API. For now I removed the payload option and we only return the complete event envelope if requested.

api/ceramic.yaml
```diff
@@ -391,10 +391,21 @@ paths:
...
+          in: query
+          required: false
+          schema:
+            type: string
+            enum: [none, full]
+          description: >
+            Whether to include the event data (carfile) in the response.
+            In the future, only the payload or other options may be supported:
+              * `none` - Empty, only the event ID is returned
+              * `full` - The entire envelope carfile (including the envelope and payload)
```
stbrody, line 408:
```suggestion
              * `full` - The entire event carfile (including the envelope and payload)
```

stbrody (approved):
lgtm

PR #399, opened 2024-06-26
refactor: use an Iterator for insert_many and other clean up

dav1do:
This addresses some feedback from #387. 

The `recon::Store` trait takes an explicit lifetime instead of elided because that could cause confusing issues (and in the changes to test with loom that got delayed we needed it #312). Mostly small clean up, other than changing the store insert_many function to take an `Iterator<Item = &'a EventInsertable>` and refactored some of the code in service to reduce allocations. 

~~I accidentally started from #398 so I'm currently targeting that but can rebase to main.~~

### DIFF
--- a/service/src/event/service.rs
+++ b/service/src/event/service.rs
@@ -111,8 +111,7 @@ impl CeramicEventService {
         Ok(())
     }
 
-    /// Currently only verifies that the event parses into a valid ceramic event, determining whether it's
-    /// immediately deliverable because it's an init event or it needs review.
+    /// Currently only verifies that the event parses into a valid ceramic event.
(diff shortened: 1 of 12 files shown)


dbcfd (approved):

service/src/event/service.rs
```diff
@@ -197,68 +197,72 @@ impl CeramicEventService {
...
-            return Ok(InsertResult {
-                store_result: res,
-                missing_history,
-            });
-        }
+        Ok(InsertResult {
+            store_result,
+            missing_history,
+        })
+    }
```
nathanielc, line 227: I really like how this logic has simplified. It very clear how the two branches differ in behavior with and without history being required.

nathanielc (approved):

PR #402, opened 2024-06-27
docs: adds UPGRADE.md docs

nathanielc:
Fixes AES-98

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -98,21 +98,15 @@ The first few logs line will look similar to this:
 
 ### Configure `js-ceramic` for `ceramic-one`
 
-First change the `ipfs` section in the `daemon.config.json` to use `remote` mode and point it at the running `ceramic-one` daemon.
-By default the `ceramic-one` daemon listens on port 5001, if you configured it differently make sure to match that configuration for `js-ceramic` as well.
+Change the `ipfs` section in the `daemon.config.json` to use `remote` mode and point it at the running `ceramic-one` daemon.
+By default the `ceramic-one` daemon listens on port 5101, if you configured it differently make sure to match that configuration for `js-ceramic` as well.
 
-```josn
+```json
 "ipfs": {
     "mode": "remote",
-    "host": "http://localhost:5001"
+    "host": "http://localhost:5101"
   },
 ```
-
-Second, set an environment variable for the `js-ceramic` process so it knows its talking to `ceramic-one` and not Kubo.
-
-    CERAMIC_RECON_MODE=true
-
-Both of these configuration options may not be required in the future, however for now they are.
 
 
 ### Start `js-ceramic` Node and Validate
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,0 +1,114 @@
+# Upgrade to Ceramic One from Kubo
+
+This document describes how to upgrade an existing Ceramic node using Kubo to run with `ceramic-one` while preserving existing data.
+
+## Overview
+
+Follow these steps for basic upgrade of a Ceramic node with some downtime.
+
+1. Shutdown existing `js-ceramic` node.
+2. Migrate data from IPFS blocks into Sqlite database file `ceramic-one` can consume.
+3. Start `ceramic-one` daemon.
+4. Configure `js-ceramic` to run against `ceramic-one`.
+5. Start `js-ceramic` node and validate.
+
+If you require a zero downtime upgrade see the section below that expands the basic steps for a near zero downtime upgrade.
+
+### Shutdown Existing Node
+
+First shutdown both the `js-ceramic` and `kubo` processes using whatever mechanism your infrastructure dictates.
+
+### Migrate IPFS Blocks to Ceramic Events
+
+Kubo stores data as IPFS blocks on disk.
+If you are not using the disk backed Kubo storage first migrate data to an on disk store following Kubo's instructions.
+
+These IPFS blocks are the raw pieces of Ceramic Events.
+In order to preserve this data we need to migrate them into Ceramic Events stored in as a Sqlite database file.
+
+The `ceramic-one` command contains a subcommand for migrating data from IPFS blocks.
+Install the `ceramic-one` binary following the instructions in README.md.
+
+See the complete usage information for the migration command:
+
+    ceramic-one migrations from-ipfs --help
+
+The command requires an input directory where the IPFS blocks are found and where the output storage directory.
+
+The following command will migrate data assuming the default directories for both Kubo and `ceramic-one`.
+Adjust the command arguments as needed by your infrastructure.
+
+    ceramic-one migrations from-ipfs -i ~/.ipfs/blocks -o ~/.ceramic-one --network mainnet
+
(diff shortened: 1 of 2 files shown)


UPGRADE.md
```diff
@@ -0,0 +1,114 @@
+# Upgrade to Ceramic One from Kubo
+
+This document describes how to upgrade an existing Ceramic node using Kubo to run with `ceramic-one` while preserving existing data.
```
oed, line 3: Can we be more precise here? It's really how to upgrade a js-ceramic node from using Kubo to using ceramic-one.

UPGRADE.md
```diff
@@ -0,0 +1,114 @@
...
+If you require a zero downtime upgrade see the section below that expands the basic steps for a near zero downtime upgrade.
+
+### Shutdown Existing Node
+
+First shutdown both the `js-ceramic` and `kubo` processes using whatever mechanism your infrastructure dictates.
+
+### Migrate IPFS Blocks to Ceramic Events
+
+Kubo stores data as IPFS blocks on disk.
+If you are not using the disk backed Kubo storage first migrate data to an on disk store following Kubo's instructions.
```
JustinaPetr, line 24: Could we potentially add a link here for where to find the instructions? I understand this should be somewhere in Kubo's documentation, but having an easy to find link could help following this process easier.

UPGRADE.md
```diff
@@ -0,0 +1,114 @@
...
+
+Read the full usage of the `ceramic-one` daemon:
+
+    ceramic-one daemon --help
+
+Once you are familiar with configuring the daemon start it with defaults or specify any arguments specific to your infrastructure.
+
+    ceramic-one daemon
+
+You should see logs indicating the daemon is running.
```
JustinaPetr, line 62: Being more specific of what output the devs should see here would be ideal to make it easier to verify that devs are following the upgrade process correctly

UPGRADE.md
```diff
@@ -0,0 +1,114 @@
...
+    ceramic-one migrations from-ipfs --help
+
+The command requires an input directory where the IPFS blocks are found and where the output storage directory.
+
+The following command will migrate data assuming the default directories for both Kubo and `ceramic-one`.
+Adjust the command arguments as needed by your infrastructure.
+
+    ceramic-one migrations from-ipfs -i ~/.ipfs/blocks -o ~/.ceramic-one --network mainnet
+
+This command will migrate all blocks stored in the `~/.ipfs/blocks` directory into the database file `~/.ceramic-one/db.sqlite3` for the `mainnet` network.
```
JustinaPetr, line 43: I wonder if there is a step or an output that devs could use as a checkpoint to verify that this step in the migration process was completed correctly? If there is, ti would be worth including it here.

UPGRADE.md
```diff
@@ -0,0 +1,114 @@
...
+Second, set and environment variable for the `js-ceramic` process so it knows its talking to `ceramic-one` and not Kubo.
+
+    CERAMIC_RECON_MODE=true
+
+Both of these configuration options may not be required in the future, however for now they are.
+
+
+### Start `js-ceramic` Node and Validate
+
+Now start the `js-ceramic` node with its new configuration and environment.
```
JustinaPetr, line 85: For easier usability of the guide it would be worth including the actual command that devs should use here.

UPGRADE.md
```diff
@@ -0,0 +1,114 @@
...
+Validate the process starts as before and that your application can communicate with it as needed.
```
JustinaPetr, line 86: Can we include the pointers of what exactly devs should check here? Anything else that would confirm developers that the migration has completed successfully?

UPGRADE.md
```diff
@@ -0,0 +1,114 @@
...
+```josn
+"ipfs": {
+    "mode": "remote",
+    "host": "http://localhost:5001"
+  },
+```
+
+Second, set and environment variable for the `js-ceramic` process so it knows its talking to `ceramic-one` and not Kubo.
+
+    CERAMIC_RECON_MODE=true
```
stbrody, line 78: this shouldn't be necessary in the GA version right since we plan to make that the default?

UPGRADE.md
nathanielc: Hmm their docs do not seem to exist. I'll reword this

UPGRADE.md
nathanielc: Starting the js-ceramic process will fail if events haven't been migrated. I'll call that out here

UPGRADE.md
nathanielc: It depends on their infrastructure so I left it unspecified.

UPGRADE.md
nathanielc: Correct I'll remove this once that change is live.

UPGRADE.md
```diff
@@ -0,0 +1,148 @@
...
+Now start the `js-ceramic` node with its new configuration and environment.
+Validate the process starts as before and that your application can communicate with it as needed.
+
+To validate that everything worked you can run any tests you may have against the node or perform a random sample query to ensure all data has migrated.
+This will depend on the specifics of your application.
+
+## Recovery
+
+At this point you should have a working upgraded `js-ceramic` node but sometimes things go wrong.
+The Ceramic node does some basic checks on startup to ensure its various data store are in sync, if these check fail the process with exit with an error message about missing data.
```
stbrody, line 123:
```suggestion
The Ceramic node does some basic checks on startup to ensure its various data stores are in sync, if these checks fail the process will exit with an error message about missing data.
```

UPGRADE.md
```diff
@@ -0,0 +1,148 @@
...
+
+1. Stop the `ceramic-one` and `js-ceramic` processes.
+2. Revert the configuration and environment changes to `js-ceramic`.
+3. Start `js-ceramic` process.
+
+These steps work because the existing data is not modified during the migration, instead it is only read and copied.
+
+## Live Upgrade Steps
+
+If you cannot tolerate a window of downtime of o your Ceramic node you can perform a two pass migration process to minimize the amount of time your node is offline.
```
stbrody, line 134:
```suggestion
If you cannot tolerate a window of downtime of your Ceramic node you can perform a two pass migration process to minimize the amount of time your node is offline.
```

UPGRADE.md
```diff
@@ -0,0 +1,148 @@
...
+The basic strategy here is to run two nodes behind a proxy in an active/passive mode.
+
+These steps are very similar to the normal upgrade process so please review those steps first.
+The high level sequence is:
+
+1. Migrate the data from the active old node.
+2. Start new node as the passive with the migrated data.
+3. Validate new passive node is working
+4. Switch active role to new node and passive role to old node.
```
stbrody, line 143: this means there's a period of time where the node is live but missing data.  I'm not sure that really counts as a no-downtime migration.  Missing data could cause application-level logic issues.  At the very least I think we need to call this out loudly as a serious caveat for this migration approach.

stbrody (changes requested):
most comments are small nitpicks, the only one that actually matters is the last one at the bottom about the "zero downtime" migration not actually really being safe or zero downtime.

stbrody (approved):

PR #403, opened 2024-06-27
fix: use event service rather than store to enforce deliverability/or…

dav1do:
The migration collects the blocks and builds a carfile that it hands off to the event service to order/insert/etc. This definitely does more work (we parse the cbor more and do more sorting) but I haven't measured to see how much of an impact it has.

### DIFF
--- a/service/src/event/migration.rs
+++ b/service/src/event/migration.rs
@@ -3,21 +3,21 @@ use std::collections::{BTreeMap, BTreeSet};
 use anyhow::{anyhow, bail, Context as _, Result};
 use ceramic_core::{EventId, Network};
 use ceramic_event::unvalidated::{self, signed::cacao::Capability};
-use ceramic_store::{
-    BlockHash, CeramicOneEvent, EventBlockRaw, EventInsertable, EventInsertableBody, SqlitePool,
-};
 use cid::Cid;
 use futures::{Stream, StreamExt, TryStreamExt};
 use ipld_core::ipld::Ipld;
 use thiserror::Error;
 use tracing::{debug, error, info, instrument, Level};
 
+use crate::CeramicEventService;
+
 use super::service::{Block, BoxedBlock};
 
-pub struct Migrator {
+pub struct Migrator<'a> {
+    service: &'a CeramicEventService,
     network: Network,
     blocks: BTreeMap<Cid, Box<dyn Block>>,
-    batch: Vec<EventInsertable>,
+    batch: Vec<(EventId, Vec<u8>)>,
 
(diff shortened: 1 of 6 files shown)


service/src/event/migration.rs
```diff
@@ -133,14 +135,21 @@ impl Migrator {
...
         Ok(())
     }
-    async fn write_batch(&mut self, sql_pool: &SqlitePool) -> Result<()> {
-        CeramicOneEvent::insert_many(sql_pool, self.batch.iter()).await?;
+    async fn write_batch(&mut self) -> Result<()> {
+        let items = self
+            .batch
+            .iter()
+            .map(|(id, body)| recon::ReconItem::new(id, body))
+            .collect::<Vec<_>>();
```
dav1do, line 149: would be nice to take an iterator here

service/src/event/migration.rs
```diff
@@ -295,33 +303,42 @@ impl EventBuilder {
...
+            .iter()
+            .fold(0, |sum, (_, row, _)| sum + row.len());
+        // Reconstruct the car file
+        let mut car = Vec::with_capacity(size + 100 * self.blocks.len());
+        let roots = self
+            .blocks
+            .iter()
+            .flat_map(|(cid, _, root)| if *root { Some(*cid) } else { None })
+            .collect::<Vec<_>>();
+        let mut writer = iroh_car::CarWriter::new(iroh_car::CarHeader::V1(roots.into()), &mut car);
```
nathanielc, line 337: This isn't correct, not all blocks should be roots. Why not leverage the unvalidated::Event::encode_car here? We could make this EventBuilder a light wrapper on the unvalidated::Builder? That way we can keep the car file encoding in one place where it is well tested.

nathanielc (changes requested):
I agree this is a better API than using the EventInsertable directly. Maybe we should just go all the way to the unvalidated::Event type?

service/src/event/migration.rs
dav1do: Yeah, I like that idea. I kind of went qnd and will change it. This should only include roots (the third field in the tuple is whether it was a root and does encode as the tests expect) but it'd be a lot better to not include here at all.

service/src/event/migration.rs
```diff
@@ -269,10 +263,13 @@ impl<'a> Migrator<'a> {
                 error!(%curr, "missing block");
                 break;
             }
+            // Add edge data to event
+            proof_edges.extend(edge);
```
nathanielc, line 267: with this being after the above break the last iteration doesn't add the edge. Maybe that is the issue?

PR #407, opened 2024-06-29
chore: homebrew support

oed:
This PR adds support for easier installation of Ceramic on Macos through homebrew. When a new release is made, the [homebrew-tap](https://github.com/ceramicnetwork/homebrew-tap) repo is updated to point to the new version.

Not sure what the best way to test the CI workflow without breaking anything. Ideally this isn't merged until we've validated it works.


smrz2001 (approved):
🚀

PR #409, opened 2024-07-01
feat: track version history in database

dav1do:
Very basic right now. Stores the current version in the database at daemon startup. This can be used in the future to prevent downgrades that go across migrations. Related to #382 but split off as we don't need to migrate any data right now.


stbrody:
@dav1do is this still something we want to move forward with getting in place?

dav1do:
> @dav1do is this still something we want to move forward with getting in place?

I rebased it as I think it's kind of nice to have, but I don't feel strongly that it's important. The ticket became less necessary as we didn't end up needing to migrate c1 data for GA. We may want this (plus something more like the changes in #382) if we need to make changes to data/schemas for event validation or the conclusion API.

nathanielc (approved):

PR #411, opened 2024-07-01
feat: change default ports used for swarm and rpc endpoint

dav1do:
As we no longer provide a compatible IPFS server, we are changing the default ports we use from 4001 (swarm) and 5001 (rpc) to 4101 (swarm) and 5101 (rpc). This will avoid any conflicts with people running a vanilla IPFS server on their machine. I also added some better logging around errors binding addresses so you'll get something like the following, instead of something like: `Address already in use (os error 48)`.

```
❯ RUST_BACKTRACE=1 RUST_LOG=info,ceramic_store=debug ./target/debug/ceramic-one daemon --network local --local-network-id 0 --store-dir ./local-data
Connected to sqlite database: ./local-data/db.sqlite3
  2024-07-01T19:58:09.317175Z  INFO ceramic_one: service__name: "ceramic-one", version: "0.25.0", build: "git:6acf4a09-modified", instance_id: "omniscient-volleyball", exe_hash: "uAAUAdebugg"
    at one/src/lib.rs:365

  2024-07-01T19:58:09.319031Z  INFO ceramic_p2p::node: identity loaded: 12D3KooWEQcETREP3t7P5RNCn9dbaWapeThSCFGVia4ryWEC2Nnc
    at p2p/src/node.rs:1158

  2024-07-01T19:58:09.325505Z  INFO libp2p_swarm: local_peer_id: 12D3KooWEQcETREP3t7P5RNCn9dbaWapeThSCFGVia4ryWEC2Nnc
    at /Users/david/.cargo/registry/src/index.crates.io-6f17d22bba15001f/libp2p-swarm-0.44.2/src/lib.rs:367

  2024-07-01T19:58:09.325657Z  INFO ceramic_p2p::node: iroh-p2p peerid: 12D3KooWEQcETREP3t7P5RNCn9dbaWapeThSCFGVia4ryWEC2Nnc
    at p2p/src/node.rs:168

  2024-07-01T19:58:09.327095Z ERROR ceramic_one: Daemon run error: Failed to listen on swarm address: /ip4/0.0.0.0/udp/4101/quic-v1.
    at one/src/lib.rs:315

Error: Failed to listen on swarm address: /ip4/0.0.0.0/udp/4101/quic-v1.
```

PR #420, opened 2024-07-02
feat: add log-format option to migrations command

nathanielc:
Additionally the log init code was moved before the sq db code so that we can use normal tracing for the db init code.


one/src/lib.rs
```diff
@@ -275,7 +289,7 @@ impl DBOpts {
             ceramic_store::SqlitePool::connect(path, ceramic_store::Migrations::Apply).await?;
         let interest_store = Arc::new(CeramicInterestService::new(sql_pool.clone()));
         let event_store = Arc::new(CeramicEventService::new(sql_pool).await?);
-        println!("Connected to sqlite database: {}", path);
+        info!(path, "connected to sqlite db");
```
dav1do, line 292: I deleted this in #411 but I like your change better 😄

dav1do (approved):

PR #422, opened 2024-07-02
chore: release archive with linux binary

oed:
This PR updates the pre-release process to also package binaries for linux. This will be helpful when creating an AUR package for arch linux.

Note: macos archives will change extension from `.tar.gz` to `.bin.tar.gz`, so we will likely need to update the archive names to include `.bin` on the most recent pre-release to not break the workflow.

Also changed the order of the README install instructions to Macos before Linux because I didn't want the Macos instructions to get lost under the *from source* instructions for Linux.

### DIFF
--- a/.github/workflows/publish-prerelease.yml
+++ b/.github/workflows/publish-prerelease.yml
@@ -70,20 +70,17 @@ jobs:
       - name: Run package script
         run: |
           ./ci-scripts/package.sh -a ${{ matrix.arch }} -e ${{ matrix.ext }}
-      - name: Archive artifact
+      - name: Archive Binary
         uses: actions/upload-artifact@v3
-        if: ${{ matrix.ext == 'bin' }}
         with:
           name: ceramic-one_${{ matrix.target }}
           path: ceramic-one_${{ matrix.target }}.bin.tar.gz
-      - name: Archive artifact
+      - name: Archive Debian Package
         uses: actions/upload-artifact@v3
         if: ${{ matrix.ext == 'deb' }}
         with:
           name: ceramic-one_${{ matrix.target }}
-          path: |
-            ceramic-one_${{ matrix.target }}.bin.tar.gz
-            ceramic-one_${{ matrix.target }}.tar.gz
+          path: ceramic-one_${{ matrix.target }}.tar.gz
 
   release:
     needs: [build-binaries]
--- a/.github/workflows/publish-prerelease.yml
+++ b/.github/workflows/publish-prerelease.yml
@@ -72,9 +72,17 @@ jobs:
           ./ci-scripts/package.sh -a ${{ matrix.arch }} -e ${{ matrix.ext }}
       - name: Archive artifact
         uses: actions/upload-artifact@v3
+        if: ${{ matrix.ext == 'bin' }}
+        with:
+          name: ceramic-one_${{ matrix.target }}
+          path: ceramic-one_${{ matrix.target }}.bin.tar.gz
+      - name: Archive artifact
+        uses: actions/upload-artifact@v3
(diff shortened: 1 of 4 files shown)


ci-scripts/package.sh
```diff
@@ -82,10 +82,10 @@ echo "Building artifacts for "$TARGET
...
 cargo build --release --locked --target $TARGET
 
-if [ "$EXT" = "bin" ]; then
-    echo "Compressing package for $TARGET"
-    tar -cvzf ceramic-one_$TARGET.tar.gz -C $BIN_DIR ceramic-one
-else
+echo "Compressing binary package for $TARGET"
+tar -cvzf ceramic-one_$TARGET.bin.tar.gz -C $BIN_DIR ceramic-one
+
+if [ "$EXT" != "bin" ]; then
```
smrz2001, line 88: Since we're using a different extension for the zipped binary, should we do the FPM build regardless of `EXT`?

smrz2001: And then we can upload both `*.bin.tar.gz` and `*.tar.gz` for all archs.

.github/workflows/publish-prerelease.yml
```diff
@@ -72,9 +72,17 @@ jobs:
           ./ci-scripts/package.sh -a ${{ matrix.arch }} -e ${{ matrix.ext }}
       - name: Archive artifact
         uses: actions/upload-artifact@v3
+        if: ${{ matrix.ext == 'bin' }}
+        with:
+          name: ceramic-one_${{ matrix.target }}
+          path: ceramic-one_${{ matrix.target }}.bin.tar.gz
+      - name: Archive artifact
+        uses: actions/upload-artifact@v3
+        if: ${{ matrix.ext == 'deb' }}
```
smrz2001, line 81: Can possible take these lines out and just upload both files for all archs?
```
            ceramic-one_${{ matrix.target }}.bin.tar.gz
            ceramic-one_${{ matrix.target }}.tar.gz
```

oed: How?

smrz2001: I'll try it out and let you know.

ci-scripts/package.sh
smrz2001: We don't really, I was just trying to simplify the flow to work the same for all cases.

.github/workflows/publish-prerelease.yml
```diff
@@ -72,9 +72,17 @@ jobs:
...
         with:
           name: ceramic-one_${{ matrix.target }}
           path: |
+            ceramic-one_${{ matrix.target }}.bin.tar.gz
             ceramic-one_${{ matrix.target }}.tar.gz
```
smrz2001, line 86: How about this?
```suggestion
      - name: Archive Binary
        uses: actions/upload-artifact@v3
        with:
          name: ceramic-one_${{ matrix.target }}
          path: ceramic-one_${{ matrix.target }}.bin.tar.gz
      - name: Archive Debian Package
        uses: actions/upload-artifact@v3
        if: ${{ matrix.ext == 'deb' }}
        with:
          name: ceramic-one_${{ matrix.target }}
          path: ceramic-one_${{ matrix.target }}.tar.gz
```

smrz2001 (approved):
LGTM! Just one minor suggestion to simplify the flow.

oed:
Thanks Mohsin. Can I just click merge now, or how does the merge queue work?

smrz2001:
> Thanks Mohsin. Can I just click merge now, or how does the merge queue work?
> 

Yes, can queue for merge. It will go through the merge queue checks then get merged to main. Then I'll create a new prerelease.

PR #424, opened 2024-07-03
fix: optimize migrations logic for less memory

nathanielc:
The migrations logic now does not need to read all blocks CIDs into memory.

This has a slight draw back from the previous code in that if a block was mislabeled on disk we would have still found it. Now we will not be able to find that block, however this is not expected to happen.

### DIFF
--- a/one/src/migrations.rs
+++ b/one/src/migrations.rs
@@ -181,5 +181,5 @@ async fn block_from_path(block_path: PathBuf) -> Result<Option<(Cid, Vec<u8>)>> 
     } else {
         Cid::new_v1(DAG_CBOR, hash)
     };
-    Ok(Some((cid, tokio::fs::read(&block_path).await?)))
+    Ok(Some((cid, blob)))
 }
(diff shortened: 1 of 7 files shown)


one/src/migrations.rs
```diff
@@ -84,34 +84,68 @@ async fn from_ipfs(opts: FromIpfsOpts) -> Result<()> {
...
+        // The leading "B" for the b32 sha256 multihash is left off
+        // ~/.ipfs/blocks/QV/CIQOHMGEIKMPYHAUTL57JSEZN64SIJ5OIHSGJG4TJSSJLGI3PBJLQVI.data // cspell:disable-line
+        info!(path = %self.input_ipfs_path.display(), "opening IPFS repo");
+
+        let mut dirs = Vec::new();
+        dirs.push(self.input_ipfs_path.clone());
+
+        try_stream! {
+            while !dirs.is_empty() {
+                let mut entries = tokio::fs::read_dir(dirs.pop().unwrap()).await?;
```
dav1do, line 112: nit: I was worried this might crash on an empty dir before I realized the while condition.. my preference is probably `while let Some(dir) = dirs.pop()` but same same.

dav1do (approved):

PR #426, opened 2024-07-08
feat: make private key dir configurable

smrz2001:
Tested locally to be working via env var and CLI flag.

### DIFF
--- a/one/src/lib.rs
+++ b/one/src/lib.rs
@@ -143,6 +143,9 @@ struct DaemonOpts {
 
     #[command(flatten)]
     db_opts: DBOpts,
+
+    #[command(flatten)]
+    key_opts: KeyOpts,
 }
 
 #[derive(Args, Debug)]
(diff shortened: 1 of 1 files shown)


AaronGoldman (approved):

one/src/lib.rs
```diff
@@ -239,6 +242,27 @@ impl Network {
     }
 }
 
+#[derive(Args, Debug)]
+struct KeyOpts {
+    /// Path to private key directory
```
stbrody, line 247: can we clarify somehow *what* private key we are talking about here?

I'm assuming this is the key that is used by libp2p and is what the PeerID derives from?

Eventually we may also additional private keys such as:
- a node DID private key for the node to sign messages and identify itself on the network (ideally this would re-use the same key as libp2p is using for the peerid I think)
- an ethereum wallet private key
- an adm chain wallet private key

So we should be specific here about what we're talking about

nathanielc (approved):

PR #431, opened 2024-07-11
feat: change default dir to current directory

nathanielc:
BREAKING CHANGE: Both the store and p2p key dirs now default to the current working directory instead of the home/.ceramic-one directory.

Generally this makes it less error prone when running multiple daemon on a single host as the naturally isolate from each other using their working directory. Running multiple daemons locally is a common task in order to experiment with the p2p nature of ceramic-one.

This is a breaking change. If you previously relied on the home directory default you must now explicitly pass in the home directory like so:

    ceramic-one daemon -s ~/.ceramic-one -p ~/.ceramic-one

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1250,7 +1250,6 @@ dependencies = [
  "expect-test",
  "futures",
  "git-version",
- "home",
  "hyper",
  "iroh-bitswap",
  "iroh-rpc-client",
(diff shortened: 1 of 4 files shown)


dav1do (approved):
LGTM. I wouldn't mind including (breaking) or something like that in the PR title just to make it obvious when we're doing release notes but otherwise like the simplification and explicitness.

PR #432, opened 2024-07-11
feat: add debug to all event types

nathanielc:
Adds Debug implementation for Event type and ensures CIDs are printed as strings instead of deconstruction of the CID struct

### DIFF
--- a/event/src/bytes.rs
+++ b/event/src/bytes.rs
@@ -1,13 +1,23 @@
 //! Thin wrapper on a Vec<u8> that correctly serializes and deserializes as bytes as opposed to a
 //! sequence of u8 integers.
+use std::fmt::Debug;
+
 use serde::{
     de::{self, Visitor},
     Deserialize, Serialize,
 };
 
 /// Sequence of byte values.
-#[derive(Clone, PartialEq, Eq, Default, Debug)]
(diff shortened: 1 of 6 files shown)


smrz2001 (approved):
🥳

event/src/bytes.rs
```diff
@@ -1,14 +1,24 @@
...
 
 /// Sequence of byte values.
-#[derive(Clone, PartialEq, Eq, Default, Debug)]
+#[derive(Clone, PartialEq, Eq, Default)]
 pub struct Bytes(Vec<u8>);
 
+impl Debug for Bytes {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_tuple("Bytes")
+            .field(&multibase::encode(multibase::Base::Base64Url, &self.0))
```
AaronGoldman, line 17: Why did we pick b64 for this?
I find when debugging hex is easier to look at what the bytes are doing as bytes don't span chars. That said the b64 is shorter for use in expect tests.

nathanielc: It was arbitrary however I wanted it to have the multibase prefix instead of raw hex since many tools expect multibase (cpk being one). I'll change to base 16 lower. Thanks for the call out.

PR #441, opened 2024-07-16
feat: expose current ceramic network via api

dav1do:
This adds a new endpoint to support retrieving the current network the server is connected to. I changing the name formatting to remove `/ceramic/` and return the values that match more closely to what we previously used in js-ceramic. It is possibly a subtly breaking change if people were relying on this string representation to match the pubsub topics, but I can't find any uses in the repo, http client or keramik so went ahead and changed it. Open to implementing `Display` or another function instead if others deem that worthwhile.

I also added `inmemory` as an alias of `in-memory` for the network cli flag so either can be passed in. It's not advertised in help/completion but it will work if passed in.

```bash
# GET /ceramic/network
❯ curl http://127.0.0.1:5101/ceramic/config/network
{"name":"local-0"}
```

### DIFF
--- a/api/ceramic.yaml
+++ b/api/ceramic.yaml
@@ -254,7 +254,7 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorResponse"
-  "/network":
+  "/config/network":
     options:
       summary: cors
       responses:
--- a/api/ceramic.yaml
+++ b/api/ceramic.yaml
@@ -254,6 +254,21 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorResponse"
+  "/network":
+    options:
+      summary: cors
+      responses:
+        "200":
+          description: cors
+    get:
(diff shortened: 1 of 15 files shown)


api/ceramic.yaml
```diff
@@ -254,6 +254,21 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorResponse"
+  "/network":
```
stbrody, line 257: Should we maybe nest this under some other namespace?  I'm still trying to think ahead to where we might have dozens of different things we want to be able to ask a node about its configuration.

`/config/network`?
`/nodeInfo/network`?

something like that?

stbrody (approved):
LGTM mod question about namespacing the endpoint

api/ceramic.yaml
dav1do: Yeah, I moved it under /config. I considered /node as that is similar to what js-ceramic used for things like chain info but went with config for now. Will merge tomorrow if no objections.

PR #442, opened 2024-07-16
fix: workflow update deb repo

3benbox:
This works, but there is only 1 version allowed at a time.

So when we publish a new version of the debian package it will be the only one available.

Not sure how I feel about that


oed:
Why do we have a separate repo to host the .deb file in the first place? Why can't the .deb file just be downloaded from the github release?

3benbox:
> Why do we have a separate repo to host the .deb file in the first place? Why can't the .deb file just be downloaded from the github release?

The same reason we have a separate homebrew formula repo, to make installation easier with existing host tooling.

PR #446, opened 2024-07-24
feat: Log bind address at startup

stbrody:
With this change at startup there's a message like
```
  2024-07-24T21:40:00.705970Z  INFO ceramic_one: starting api server at address 127.0.0.1:5101
    at one/src/lib.rs:575
```

It does make me wonder though if we should separate `bind_address` into separate args for the port to use and the network interface to bind to.  It's kind of weird to print `127.0.0.1:5101` when presumably we also accept connections that don't just come from localhost.  Really it's the port that's going to be the most interesting part for most people

### DIFF
--- a/one/src/lib.rs
+++ b/one/src/lib.rs
@@ -572,7 +572,7 @@ impl Daemon {
         let signals_handle = tokio::spawn(handle_signals(signals, tx));
 
         // The server task blocks until we are ready to start shutdown
-        debug!("starting api server");
+        info!("starting api server at address {}", opts.bind_address);
         hyper::server::Server::try_bind(&opts.bind_address.parse()?)
             .map_err(|e| anyhow!("Failed to bind address: {}. {}", opts.bind_address, e))?
             .serve(service)


dav1do (approved):
For some reason I thought we already did this but it's only the p2p address. I think accepting the socket address as one parameter is fine personally, but it might be nice to actually make the CLI flag a `SocketAddr` instead of a string. It doesn't really change much, other than getting an error earlier e.g.

```
❯ ./target/debug/ceramic-one daemon --network local --local-network-id 0 --bind-address 127.0.0:5101
  2024-07-25T17:43:18.790913Z  INFO ceramic_one: service__name: "ceramic-one", version: "0.29.0", build: "git:570d299f", instance_id: "beautiful-able", exe_hash: "uAAUAdebugg"
    at one/src/lib.rs:343

...

  2024-07-25T17:43:18.974446Z  INFO ceramic_api::server: Shutting down insert task.
    at api/src/server.rs:347

  2024-07-25T17:43:18.974709Z ERROR ceramic_one: Error running command: invalid socket address syntax
    at one/src/main.rs:15

Error: invalid socket address syntax
```
vs
```
❯ ./target/debug/ceramic-one daemon --network local --local-network-id 0 --bind-address 127.0.0:5101
error: invalid value '127.0.0:5101' for '--bind-address <BIND_ADDRESS>': invalid socket address syntax
```

stbrody:
yeah that does look like a nice improvement

PR #447, opened 2024-07-26
fix: use default (google) dns servers if system resolv.conf is invalid

dav1do:
If the system dns conf is missing nameservers, dns queries will never be resolved. Currently, hickory-dns doesn't error in this case so we handle it manually until the fix is released. This happens on osx if wifi is disabled when you start c1 and is never able to resolve itself without this change, even after connectivity returns. However, if the interface is active when the process starts and then wifi is disabled, it will resolve itself and we will use whatever was defined in resolv.conf.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1292,6 +1292,7 @@ dependencies = [
  "criterion2",
  "futures",
  "futures-util",
+ "hickory-resolver",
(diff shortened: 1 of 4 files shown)


smrz2001 (approved):
🚀

PR #448, opened 2024-07-27
feat: feature and experimental feature flags

dbcfd:


one/src/lib.rs
```diff
@@ -167,6 +169,24 @@ struct DaemonOpts {
...
     cors_allow_origins: Vec<String>,
+
+    /// Enable experimental feature flags
+    #[arg(
+        long,
+        value_parser = parse_flags::<ExperimentalFeatureFlags>,
+        env = "CERAMIC_ONE_EXPERIMENTAL_FEATURE_FLAGS",
+        value_names = ["Authentication", "None"]
+    )]
+    experiemental_feature_flags: Vec<ExperimentalFeatureFlags>,
```
dav1do, line 180: typo: `experiemental` => experimental

one/src/lib.rs
```diff
@@ -167,6 +169,24 @@ struct DaemonOpts {
...
+
+    /// Enable feature flags
+    #[arg(
+        long,
+        value_parser = parse_flags::<FeatureFlags>,
+        env = "CERAMIC_ONE_FEATURE_FLAGS",
+        value_names = ["None"]
+    )]
+    feature_flags: Vec<FeatureFlags>,
```
dav1do, line 189: Do we need this one right now? I don't really like having `none` as an option for either of these. It seems like it should just be a vec and if it's empty it means none. having 3 features and none "enabled" is confusing.

one/src/lib.rs
```diff
@@ -167,6 +169,24 @@ struct DaemonOpts {
             env = "CERAMIC_ONE_CORS_ALLOW_ORIGINS"
         )]
     cors_allow_origins: Vec<String>,
+
+    /// Enable experimental feature flags
+    #[arg(
+        long,
+        value_parser = parse_flags::<ExperimentalFeatureFlags>,
```
dav1do, line 176: Why the custom parser? It seems easier to just derive `ValueEnum` on the two types and do something like this, which gives us nicer help text/matching.

```rust
    #[arg(
        long,
        use_value_delimiter = true,
        value_delimiter = ',',
        env = "CERAMIC_ONE_EXPERIMENTAL_FEATURE_FLAGS"
    )]```
    
plus I don't think this quite works how it's expected as it's requiring all the values to be passed

dbcfd: Yeah, I was trying to get it to list the enum values, but that ended up not working how I wanted. Switching to value_delimiter

dav1do (approved):
I don't like the ExperimentalFeatureFlags supporting `none,authentication` as it doesn't really make sense.. but we can change it in the private data follow ups. I don't think anyone other than us will rely on these flags anytime soon.

Also, I'm a bit mixed on the utility of feature flags like this. I'm having a hard time coming up with examples we'd want to use that are just bools and suspect some features might require inputs (e.g. c1 anchoring requires a CAS url).. so not sure if we just want the new CLI flag to represent feature=enabled, and/or have a FF in addition. Anyway, being able to find the FFs in one place is maybe more convenient than a bunch of cli flags if we can include some description but I'm not sure how much we'll be able to leverage it in practice.

stbrody (commented):
agreed that we likely want more than just boolean flags here.  The way MongoDB handled feature flags that I think worked reasonably well was that there was a CLI flag `--setParameter <key>=<value>`.  The flag could be set multiple times, each time it was given a key-value pair.  That way you got the benefit of being able to control multiple flags, each flag could have its own name and value type that could be different for each, and each flag was still grouped under one system so it was easy to list all available flags, but they were also taken one level below the top-level CLI flags so there was a kind of hierarchy between which configuration options were important/commonly-used enough to be top-level cli flags vs being nested under setParameter.  There were also multiple ways to set the options that were controlled by setParameter: cli flag, config file, or API command at runtime.  See https://www.mongodb.com/docs/manual/reference/parameters/ for the docs on this.

For us to start I don't think we need to go that far, we definitely don't need to wire this into the runtime API or config file.  But I do like the idea of a flag that can be set multiple times and each time provides a key-value pair.  That feels the most flexible to me.

PR #450, opened 2024-07-29
chore: rust 1.80 lints

dav1do:
Modified the server generator scripts to ignore some new warnings and removed some unused code. I considered upgrading the generator to version 7.7.0 as it removes the duplicate allow_unused statements, but it didn't compile when I regenerated the server. 

It looks like axum support is in beta and hyper 1.0 support was merged but I'm not sure the ETA on arrival. Maybe a breaking release in September, which would be nice to have.


smrz2001 (approved):
🚀

PR #454, opened 2024-07-31
fix: create release prs via bot

smrz2001:
...so that they're not all created using my token.

[This](https://github.com/ceramicnetwork/rust-ceramic/pull/453) is an example of what they'll look like now.


dav1do (approved):
:shipit:

PR #456, opened 2024-08-01
feat: add peer id filter to interests

nathanielc:
Add simple ability to filter interests by a peer. Useful in debugging the interests for a local node.

### DIFF
--- a/api/src/server.rs
+++ b/api/src/server.rs
@@ -498,7 +498,10 @@ where
         ))
     }
 
-    pub async fn get_interests(&self) -> Result<ExperimentalInterestsGetResponse, ErrorResponse> {
+    pub async fn get_interests(
+        &self,
+        peer_id: Option<String>,
+    ) -> Result<ExperimentalInterestsGetResponse, ErrorResponse> {
         let interests = self
             .interest
             .range(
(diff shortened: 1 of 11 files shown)


AaronGoldman (approved):

api/src/server.rs
```diff
@@ -841,9 +863,10 @@ where
 
     async fn experimental_interests_get(
         &self,
+        peer_id: Option<String>,
         _context: &C,
     ) -> Result<ExperimentalInterestsGetResponse, ApiError> {
-        self.get_interests()
+        self.get_interests(peer_id)
```
gvelez17, line 869: hm just curious how this relates to the definition of interests with dids, and if the peer id is the forwarding peer or the originating peer, if we know.  Do we record the originating peer anywhere, i don't think so?

anyway this is not critical, its just my only thought here, maybe clarify the meaning of peer_id in a comment

nathanielc: Interests only store the originating peer. The forwarding peer is not stored, so this is always the originating peer. Or in other words the peer id of an interest is the peer that is interested in the range contained within.  

And for now we use the libp2p peer id and not the did of the peer to identify it.

PR #468, opened 2024-08-07
feat: recon event insert batching

dav1do:
The main feature added is some basic insert batching in recon. Originally, we thought we might use a channel and manage the result of batches outside of the recon protocol, but it would require substantial changes to make it work without reading its writes currently. If we add RedTree or another in memory state to capture the known keys before persisting them, that would work. In the end, this only really ends up batching writes if we're receiving a lot of values without sending ranges. Previously we always inserted every key immediately, so this should still be an improvement but it will require benchmarking to say for certain. The messages may arrive in a slightly different order, but the non-deterministic word order test passes fine, as does a new test using disjoint and demonstrating the same messages being sent but in a slightly different order.

This is mostly refactoring to simplify the insert/insert_many operations. It removes the lifetime and uses an `Arc` over the value to make it cheap to clone if needed. We send things to recon on a channel, we have to use an owned value (and it generally simplifies thing to remove lifetimes as they're basically code pollution). It's possible use a higher ranked bound instead of the function lifetime would be better, but the Arc shouldn't be too bad and we still pass in slices to the store generally. The `ceramic_api` uses a new type that wraps the tuple, and we just pass a `Vec` to make it simple to map into one thing in the service since we don't need it, but that could be adjusted.

### DIFF
--- a/api/src/lib.rs
+++ b/api/src/lib.rs
@@ -4,7 +4,8 @@ mod server;
 pub use resume_token::ResumeToken;
 
 pub use server::{
-    EventDataResult, EventInsertResult, EventStore, IncludeEventData, InterestStore, Server,
+    ApiItem, EventDataResult, EventInsertResult, EventStore, IncludeEventData, InterestStore,
+    Server,
 };
 
 #[cfg(test)]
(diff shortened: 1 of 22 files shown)


dav1do:
I used the wrong ticket number 🤦 opening a new PR with a new branch name.

PR #470, opened 2024-08-07
feat: recon event insert batching

dav1do:
The main feature added is some basic insert batching in recon. Originally, we thought we might use a channel and manage the result of batches outside of the recon protocol, but it would require substantial changes to make it work without reading its writes currently. If we add RedTree or another in memory state to capture the known keys before persisting them, that would work. In the end, this only really ends up batching writes if we're receiving a lot of values without sending ranges. Previously we always inserted every key immediately, so this should still be an improvement but it will require benchmarking to say for certain. The messages may arrive in a slightly different order, but the non-deterministic word order test passes fine, as does a new test using disjoint and demonstrating the same messages being sent but in a slightly different order.

This is mostly refactoring to simplify the insert/insert_many operations. It removes the lifetime and uses an `Arc` over the value to make it cheap to clone if needed. We send things to recon on a channel, we have to use an owned value (and it generally simplifies things to remove lifetimes as they can feel a bit like code pollution until proven necessary). If we need to bring it back, it's possible that a higher ranked lifetime bound would be better, but the Arc shouldn't be too bad and we still pass in slices to the store generally. The `ceramic_api` uses a new type that wraps the tuple, and we just pass a `Vec` to make it simple to map into one thing in the service since we the API no longer needs the data, but that could be adjusted.

### DIFF
--- a/recon/src/client.rs
+++ b/recon/src/client.rs
@@ -242,7 +242,7 @@ where
                     Request::Insert { key, value, ret } => {
                         let val = self
                             .recon
-                            .insert(ReconItem::new(&key, &value))
+                            .insert(ReconItem::new(key, value))
                             .await
                             .map_err(Error::from);
                         send(ret, val);
--- a/recon/src/client.rs
+++ b/recon/src/client.rs
@@ -12,7 +12,10 @@ use crate::{
 
 /// Client to a [`Recon`] [`Server`].
 #[derive(Debug, Clone)]
-pub struct Client<K, H> {
+pub struct Client<K, H>
+where
+    K: Key,
+{
     sender: Sender<Request<K, H>>,
     metrics: Metrics,
 }
@@ -23,11 +26,9 @@ where
     H: AssociativeHash,
 {
     /// Sends an insert request to the server and awaits the response.
-    pub async fn insert(&self, key: K, value: Vec<u8>) -> Result<()> {
+    pub async fn insert(&self, items: Vec<ReconItem<K>>) -> Result<()> {
         let (ret, rx) = oneshot::channel();
-        self.sender
-            .send(Request::Insert { key, value, ret })
-            .await?;
+        self.sender.send(Request::Insert { items, ret }).await?;
         rx.await?
     }
 
@@ -138,10 +139,12 @@ where
     }
 }
 
-enum Request<K, H> {
+enum Request<K, H>
+where
+    K: Key,
+{
     Insert {
-        key: K,
-        value: Vec<u8>,
+        items: Vec<ReconItem<K>>,
         ret: oneshot::Sender<Result<()>>,
     },
     Len {
@@ -239,12 +242,8 @@ where
             let request = self.requests.recv().await;
             if let Some(request) = request {
                 match request {
-                    Request::Insert { key, value, ret } => {
-                        let val = self
-                            .recon
-                            .insert(ReconItem::new(key, value))
-                            .await
-                            .map_err(Error::from);
+                    Request::Insert { items, ret } => {
+                        let val = self.recon.insert(items).await.map_err(Error::from);
                         send(ret, val);
                     }
                     Request::Len { ret } => {
--- a/recon/src/client.rs
+++ b/recon/src/client.rs
@@ -242,7 +242,7 @@ where
                     Request::Insert { key, value, ret } => {
                         let val = self
                             .recon
-                            .insert(&ReconItem::new(&key, &value))
+                            .insert(ReconItem::new(&key, &value))
                             .await
                             .map_err(Error::from);
                         send(ret, val);
--- a/recon/src/client.rs
+++ b/recon/src/client.rs
@@ -23,7 +23,7 @@ where
     H: AssociativeHash,
 {
     /// Sends an insert request to the server and awaits the response.
-    pub async fn insert(&self, key: K, value: Vec<u8>) -> Result<bool> {
+    pub async fn insert(&self, key: K, value: Vec<u8>) -> Result<()> {
         let (ret, rx) = oneshot::channel();
         self.sender
             .send(Request::Insert { key, value, ret })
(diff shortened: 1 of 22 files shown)


recon/src/recon.rs
```diff
@@ -279,19 +280,12 @@ where
...
-    /// Insert many keys into the key space. Includes an optional value for each key.
-    /// Returns an array with a boolean for each key indicating if the key was new.
-    /// The order is the same as the order of the keys. True means new, false means not new.
-    pub async fn insert_many(&self, items: &[ReconItem<'_, K>]) -> Result<Vec<bool>> {
-        let result = self.store.insert_many(items).await?;
-
-        Ok(result.keys)
+    /// Insert key into the key space.
+    /// Returns Ok if the result was accepted. It may be validated and stored
+    /// out of band, meaning it may not immediately return in range queries.
```
stbrody, line 285: is this still true about validation happening out of band?  Didn't you say on our call earlier that this is actually going to be happening inline for now?

recon/src/recon.rs
```diff
@@ -279,19 +280,12 @@ where
...
     }
 
-    /// Insert key into the key space. Includes an optional value.
-    /// Returns a boolean (true) indicating if the key was new.
-    pub async fn insert(&self, item: &ReconItem<'_, K>) -> Result<bool> {
-        self.store.insert(item).await
-    }
-
-    /// Insert many keys into the key space. Includes an optional value for each key.
-    /// Returns an array with a boolean for each key indicating if the key was new.
```
stbrody, line 289: we lost this per-key return value, do we not need that for anything?

service/src/event/service.rs
```diff
@@ -139,20 +139,16 @@ impl CeramicEventService {
...
-            return Ok(InsertResult::default());
-        }
-
-        let mut to_insert = Vec::with_capacity(items.len());
+        let mut to_insert = Vec::new();
 
         for event in items {
             let insertable =
-                Self::validate_discovered_event(event.key.to_owned(), event.value).await?;
+                Self::validate_discovered_event(event.key.to_owned(), &event.value).await?;
```
stbrody, line 151: is it correct that this function is where you see the main validation logic going?

stbrody (approved):
most comments are really just questions to make sure I'm understanding things right.  There are a few small suggestions, but all optional.

recon/src/client.rs
```diff
@@ -138,11 +139,13 @@ where
...
+enum Request<K, H>
+where
+    K: Key,
+{
     Insert {
-        key: K,
-        value: Vec<u8>,
-        ret: oneshot::Sender<Result<bool>>,
+        items: Vec<ReconItem<K>>,
+        ret: oneshot::Sender<Result<()>>,
```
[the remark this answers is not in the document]
dav1do: ah, sorry. it's not but this is addressed in #473 which currently targets this one.

recon/src/libp2p/protocol.rs
```diff
@@ -10,6 +10,11 @@ use crate::{
     protocol::{self, Recon},
 };
 
+// The max number of writes we'll batch up before flushing anything to disk.
+// As we descend the tree and find smaller ranges, this won't apply as we have to flush
+// before recomputing a range, but it will be used when we're processing large ranges we don't yet have.
+const INSERT_BATCH_SIZE: usize = 100;
```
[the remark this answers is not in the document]
dav1do: probably not.. it's just 100x bigger than before now and I hesitated to go 1000x :)... we can benchmark/adjust for sure

recon/src/protocol.rs
```diff
@@ -429,6 +449,7 @@ where
         to_writer: &mut ToWriterSender<InitiatorMessage<R::Key, R::Hash>>,
         remote_range: RangeHash<R::Key, R::Hash>,
     ) -> Result<()> {
+        self.common.persist_all().await?;
```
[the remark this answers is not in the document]
dav1do: yeah, that seems reasonable. we have to persist before we do the range otherwise we don't get the same result for all the values we just received. I can clean up the comments/handling a bit for sure.

recon/src/recon.rs
```diff
@@ -279,19 +280,12 @@ where
...
-    /// The order is the same as the order of the keys. True means new, false means not new.
-    pub async fn insert_many(&self, items: &[ReconItem<'_, K>]) -> Result<Vec<bool>> {
-        let result = self.store.insert_many(items).await?;
-
-        Ok(result.keys)
+    /// Insert key into the key space.
+    /// Returns Ok if the result was accepted. It may be validated and stored
+    /// out of band, meaning it may not immediately return in range queries.
+    pub async fn insert(&self, items: Vec<ReconItem<K>>) -> Result<()> {
+        let _res = self.store.insert_many(&items).await?;
```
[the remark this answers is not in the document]
dav1do: yep, it is. the result is passed back to the recon conversation in #473 and actually used

recon/src/recon.rs
dav1do: it's not, but that comment is also changed in #473 😅

recon/src/recon.rs
dav1do: we never used it for anything before. the changes in #473 can still return it if we want it, but we really only care about the range query result after persisting things

service/src/event/service.rs
dav1do: yeah, something like this. I was thinking we might want to change it to be a `self.validate_events(&[ReconItem<EventID>])` or something.. it could split things into unsigned, init that are valid, data, and time (anything unparsed is returned as invaild). if init is missing for data, it goes back as "pending", time events can be given to another task to process, and then we can `try_join!` (or something) on the set of init/data events and time events and return the result for every key (the #473 changes give a better store interface to actually do this).

recon/src/protocol.rs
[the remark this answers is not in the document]
dav1do: Will address in #473.

PR #471, opened 2024-08-07
feat: change default dir back to shared home location

dav1do:
BREAKING CHANGE: We've decided to change the default storage location back to a shared location (`$HOME/.ceramic-one` or `./.ceramic-one` if home cannot be determined). While I believe the reasoning in #431 was sound, we've decided to optimize for the unboxing experience where things "just work" no matter where you run ceramic-one, rather than experienced users or operators who will likely want to provide a directory and run multiple nodes.

Rather than simply revert that PR, I made a slight change to show the default value in the command line help, which was previously not the case when we used the shared location.

```
  -s, --store-dir <STORE_DIR>
          Path to storage directory. The default is ~/.ceramic-one [env: CERAMIC_ONE_STORE_DIR=] [default: /Users/david/.ceramic-one]
  -p, --p2p-key-dir <P2P_KEY_DIR>
          Path to libp2p private key directory [env: CERAMIC_ONE_P2P_KEY_DIR=] [default: /Users/david/.ceramic-one]
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1250,6 +1250,7 @@ dependencies = [
  "expect-test",
  "futures",
  "git-version",
+ "home",
  "hyper",
  "iroh-bitswap",
  "iroh-rpc-client",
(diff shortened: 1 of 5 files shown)


smrz2001 (approved):
🚀

PR #473, opened 2024-08-09
feat: modify recon store to support event validation concepts

dav1do:
The recon store returns a new struct that includes information about the store's willingness to process the data. It includes invalid items (e.g. something that made no sense) and the count of "pending" items (i.e. something that can't be interpreted without another event. For now, recon simply logs the count of pending/invalid items in prometheus and we close the conversation if a peer is sending us invalid items. There are also some changes to the service layer to support returning this information to the API and recon.

We don't track the pending items yet, we intend to have the event service keep a map of sorts from `{neededCid: [events]}` and then it can process them when discovered. This may cause odd behavior if recon isn't getting ranges for discovered keys, so we may need to push this information back up eventually. I initially added this tracking to recon but we pretty much just send them back every time unless we communicate more information about what is needed to recon. Since we don't really care to actively seek this out yet, having the store just manage this pending set will be simpler, and if recon not getting in sync is causing problems we'll have to fix that, but it can be tested with the next phase of event validation where we actually run into these.

Previously we just threw an error when an event can't be parsed, which has been modified to return invalid and errors will be reserved for real problems like an issue talking to the storage system. There is now a struct with some recon config injected (batch size, remote peer). I don't really like this, but didn't want to update the recon or role traits to require this. We can do that (or something else) if it's preferred, for now this was simple.

This targets #470, if it's easier I can merge them into one but there are already a decent number of changes in both. This does rework the API changed in the previous one, so consolidating might make sense.


recon/src/recon/pending_cache.rs
```diff
@@ -0,0 +1,153 @@
+use std::collections::HashMap;
+
+use crate::{Key, PendingItem, ReconItem};
+
+#[derive(Debug, Default)]
+/// This struct manages tracking items that we attempted to deliver but could not be processed
+/// because we need to discover additional information.
+pub struct PendingCache<K: Key> {
+    by_needed_key: HashMap<Vec<u8>, ReconItem<K>>,
```
dav1do, line 9: This needs to point to a vec of items as we could have many events depending on one init event. However, we may nuke this concept and just keep it really simple initially, simply returning the items and putting them in the next batch.

recon/src/recon.rs
```diff
@@ -429,23 +428,97 @@ where
     }
 }
 
+/// Represents reasons the store is unwilling to persist keys and values
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum InvalidItem<K: Key> {
```
dav1do, line 433: After discussion with @nathanielc, we probably don't care about this and can simply return an (optional?) Invalid struct that simply contains counts of each type.

recon/src/recon.rs
```diff
@@ -429,23 +428,97 @@ where
...
+            invalid: Vec::new(),
+            pending: Vec::new(),
+        }
+    }
+
+    /// yah
+    pub fn item_count(&self) -> usize {
+        self.new_cnt + self.invalid.len() + self.pending.len()
+    }
+    /// yayayay
```
dav1do, line 501: oops, clean these doc comments up (or remove methods).. I added them when I was testing returning randomly selected items to mark pending and accidentally committed.

recon/src/protocol.rs
```diff
@@ -44,15 +48,58 @@ use crate::{
...
+    /// The max number of writes we'll batch up before flushing anything to disk.
+    /// As we descend the tree and find smaller ranges, this won't apply as we have to flush
+    /// before recomputing a range, but it will be used when we're processing large ranges we don't yet have.
+    pub insert_batch_size: usize,
+    /// The maximum number of items that can not be stored in memory before we begin to drop them.  
+    /// This is due to the event being dependent on another event we need to discover before it can be interpreted.
+    pub max_pending_items: usize,
+    #[allow(dead_code)]
+    /// The ID of the peer we're syncing with.
+    peer_id: PeerId,
```
dav1do, line 74: Added based on the discussion that this was needed for self anchoring work @smrz2001 @AaronGoldman

recon/src/protocol.rs
```diff
@@ -735,10 +783,35 @@ where
...
+            }
+            tracing::warn!(
+                "Recon discovered data it will never allow. Hanging up on peer {}",
+                self.config.peer_id
+            );
+            bail!("Received unknown data from peer: {}", self.config.peer_id);
         }
+
+        if !batch.pending.is_empty() {
+            let (added, remaining_capacity) = self.pending.track_pending(&mut batch.pending);
```
dav1do, line 805: Details depend on how we change the pending cache, but we should probably track every item we re-add as well

recon/src/recon.rs
```diff
@@ -429,23 +428,97 @@ where
...
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+/// Represents items the store may be able to persist in the future
+pub struct PendingItem<K>
+where
+    K: Key,
+{
+    /// The key this item needs in order to be processed.
+    pub required_key: K,
```
dav1do, line 453: We may not want to compute/return this and simply return the items and let recon retry them in the next batch (related to comments on `pending_cache.rs`)

p2p/src/node.rs
```diff
@@ -1256,7 +1256,7 @@ mod tests {
         type Key = K;
         type Hash = Sha256a;
 
-        async fn insert(&self, _items: Vec<ReconItem<Self::Key>>) -> ReconResult<()> {
+        async fn insert(&self, _items: Vec<ReconItem<Self::Key>>) -> ReconResult<InsertBatch<K>> {
```
stbrody, line 1259: why `InsertBatch<K>` instead of `InsertBatch<Self::Key>`?  They're aliases right?  Just seems weird to use `Key` in the arguments but `K` in the return type, seems better to be consistent?

recon/src/protocol.rs
```diff
@@ -44,15 +48,58 @@ use crate::{
 // Even a small limit will quickly mean that both peers have work to do.
 const PENDING_RANGES_LIMIT: usize = 20;
 
+// The max number of writes we'll batch up before flushing anything to disk.
+// As we descend the tree and find smaller ranges, this won't apply as we have to flush
+// before recomputing a range, but it will be used when we're processing large ranges we don't yet have.
+const INSERT_BATCH_SIZE: usize = 100;
```
stbrody, line 54: so I notice this moved from `recon/src/libp2p/protocol.rs` to `recon/src/protocol.rs`.  How should we be thinking about these two `protocol.rs` files and what belongs in one vs the other?

recon/src/protocol.rs
```diff
@@ -44,15 +48,58 @@ use crate::{
...
 type RM<K, H> = ReconMessage<ResponderMessage<K, H>>;
 
+#[derive(Clone, Debug)]
+/// Parameters for the protocol that can be used to track the conversation or adjust internal behavior
+pub struct ProtocolConfig {
+    /// The max number of writes we'll batch up before flushing anything to disk.
+    /// As we descend the tree and find smaller ranges, this won't apply as we have to flush
+    /// before recomputing a range, but it will be used when we're processing large ranges we don't yet have.
+    pub insert_batch_size: usize,
+    /// The maximum number of items that can not be stored in memory before we begin to drop them.
```
stbrody, line 69: that can **not** be stored in memory?  typo?

recon/src/protocol.rs
stbrody: huh, I'm surprised this is relevant to self anchoring?  Without seeing how this is going to be used though it's hard to reason about whether this actually needed or not.

recon/src/protocol.rs
```diff
@@ -735,10 +783,35 @@ where
...
+            );
+            bail!("Received unknown data from peer: {}", self.config.peer_id);
         }
+
+        if !batch.pending.is_empty() {
+            let (added, remaining_capacity) = self.pending.track_pending(&mut batch.pending);
+            if let Ok(cnt) = added.try_into() {
+                self.recon.metrics().record(&PendingEvents(cnt))
+            }
+            if remaining_capacity < self.config.max_pending_items / 10 {
```
stbrody, line 809: what happens when capacity hits zero?

recon/src/recon.rs
stbrody: Would that approach risk the conversation getting stuck in the worst case?  Sending the same events over and over that are all blocked waiting on some other event in another part of the range that it never gets to?

recon/src/recon.rs
```diff
@@ -429,23 +428,97 @@ where
...
+}
+
 /// The result of an insert operation.
-#[derive(Clone, Debug, PartialEq, Eq, Hash, Default)]
-pub struct InsertResult {
-    /// A true/false list indicating whether or not the key was new.
-    /// It is in the same order as the input list of keys.
-    pub keys: Vec<bool>,
+#[derive(Clone, Debug, PartialEq, Eq, Default)]
+pub struct InsertBatch<K>
```
stbrody, line 468: nitpick, when I see this type name I assume it contains the list of entries to be inserted.  I actually think the original name of `InsertResult` is clearer to me.  Don't feel that strongly though

recon/src/recon/pending_cache.rs
```diff
@@ -0,0 +1,153 @@
...
+        if allowed > 0 {
+            while let Some(val) = items.pop() {
+                if !self.item_keys.contains(val.item.key.as_bytes()) {
+                    self.item_keys.insert(val.item.key.as_bytes().to_vec());
+                    self.by_needed_key
+                        .insert(val.required_key.as_bytes().to_vec(), val.item);
+                    allowed = allowed.saturating_sub(1);
+                    tracked += 1;
+                }
+                if allowed == 0 || items.is_empty() {
```
stbrody, line 62: can we move these checks into the while loop condition?

recon/src/recon/pending_cache.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    /// Update our cache to include the new items up to the allowed capacity.
+    /// Returns a tuple of (items tracked, capacity remaining)
+    pub fn track_pending(&mut self, items: &mut Vec<PendingItem<K>>) -> (usize, usize) {
+        let mut tracked = 0;
+        let mut allowed = self.capacity();
+        if allowed > 0 {
+            while let Some(val) = items.pop() {
+                if !self.item_keys.contains(val.item.key.as_bytes()) {
```
stbrody, line 55: nitpick: doesn't insert return whether the key was present already or not, so that we can avoid this contains check and thus avoid an extra hash lookup?

service/src/event/store.rs
```diff
@@ -20,22 +20,12 @@ impl recon::Store for CeramicEventService {
     async fn insert_many(
         &self,
         items: &[ReconItem<Self::Key>],
-    ) -> ReconResult<recon::InsertResult> {
+    ) -> ReconResult<recon::InsertBatch<EventId>> {
         let res = self
             .insert_events(items, DeliverableRequirement::Asap)
             .await?;
-        let mut keys = vec![false; items.len()];
-        // we need to put things back in the right order that the recon trait expects, even though we don't really care about the result
```
stbrody, line 28: why don't we need to do this anymore?

recon/src/protocol.rs
dav1do: I believe the idea is to track which peer generated the event (at least in our eyes).. if it was us, we need to anchor it

recon/src/recon.rs
dav1do: in a well behaving scenario, we shouldn't recompute ranges.. when I tested one side being bad, we kept descending and got to the end of a conversation and were a bit confused, with one side hanging up and the other trying to send some more data but finding no one.. and then we started over. we need add better tests for it (including if both sides are just talking "correct" nonsense) because it's possible something worse happens like both sides not getting what they want indefinitely, which is not supposed to happen.

p2p/src/node.rs
dav1do: oversight :).. I just typed what I was typing in other places and will change it

recon/src/protocol.rs
dav1do: recon is "lower".. so I pushed it down to avoid having to pipe it through both levels. the recon protocol is the set reconciliation part, while the libp2p one includes the network streams and is an entry point from the libp2p connection. I think it would be nice to review things and see what we can simplify and if we still want/need all the layers.

recon/src/protocol.rs
dav1do: yep, definitely a typo... storing these in memory is the whole point 😅

recon/src/protocol.rs
dav1do: since I default max_pending_items to a minimum of ten, it will start logging a lot because 0 < 1.. it's debug mode but probably better to not log thousands of messages just because we ran out.. i'm not sure if we want/need this message at all tbh and can just rely on something in prometheus

recon/src/recon.rs
dav1do: I'm fine leaving it `InsertResult`.. I changed it because Result<InsertResult> is a bit odd (often having Result in the name implies it's a type alias of Result<T,E> for the crate). I originally intended to include a batch ID and maybe a channel but didn't go that route due to complexity managing the state over the connection.

recon/src/recon/pending_cache.rs
dav1do: yeah, probably a good call. I was trying to avoid overwriting something "different" discovered for whatever reason but that would be an entirely other red flag. plus this needs to accept a vec so using the entry api is probably what will happen if we keep it

recon/src/recon/pending_cache.rs
dav1do: not with the while let (unstable rust feature) but I could flip the loops so this is the outer one

service/src/event/store.rs
dav1do: we always just ignored it in the past and it's a bit annoying to recompute since everything can get reordered before being inserted

recon/src/recon/pending_cache.rs
dav1do: changed to a for loop using drain

recon/src/recon.rs
stbrody: Hmm yeah, the convention around the word Result is fair... `InsertOutcome`?  `InsertStatus`?  `InsertDetails`?

service/src/event/service.rs
```diff
@@ -230,29 +231,75 @@ impl CeramicEventService {
...
 }
 
-#[derive(Debug, PartialEq, Eq, Default)]
-pub struct InsertResult {
-    pub(crate) store_result: ceramic_store::InsertResult,
-    pub(crate) missing_history: Vec<EventId>,
+#[derive(Debug)]
+pub struct IncomingEvent;
+#[derive(Debug)]
+pub struct InitialIncomingEvent<'a> {
```
stbrody, line 237: why have this type at all if it's just a thin wrapper around ReconItem?  Why not just have `IncomingEvent::parsed` take a`ReconItem` as input?

For that matter, why have this `IncomingEvent` type at all?  Why not just a standalone function that takes `ReconItem` and parses it?

stbrody (commented):
looks good, main question is if we can simplify the types and flow for the event parsing part.

recon/src/recon.rs
```diff
@@ -429,23 +427,77 @@ where
...
+        Self {
+            new_cnt,
+            invalid: Vec::new(),
+            pending_cnt: 0,
+        }
+    }
+
+    /// Get the total count of items included whether added, pending or invalid
+    pub fn item_count(&self) -> usize {
+        self.new_cnt + self.invalid.len()
```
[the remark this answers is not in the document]
dav1do: good catch. I was going to delete pending entirely and then added it back for metrics

recon/src/recon.rs
dav1do: I think any of those would be good... but it is _kind of_ a result since it contains info about "invalid" things.. so I just left it for now. We can rename it if we feel it's important in a future PR.

recon/src/recon/pending_cache.rs
dav1do: This has been deleted for now and we can add something to the service when we create these items.

service/src/event/service.rs
dav1do: fair enough. I was trying to quickly add a state machine/builder pattern but it's not worth it yet. just deleted it and put it back to how it was

service/src/event/service.rs
```diff
@@ -113,30 +114,27 @@ impl CeramicEventService {
 
     /// Currently only verifies that the event parses into a valid ceramic event.
     /// In the future, we will need to do more event validation (verify all EventID pieces, hashes, signatures, etc).
```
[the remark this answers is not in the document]
dav1do: It did still call into the parsing code, but I just put that code back in here for now so it should be valid.

recon/src/protocol.rs
```diff
@@ -734,11 +766,39 @@ where
...
+        }
+
         let evs: Vec<_> = self.event_q.drain(..).collect();
-        if !evs.is_empty() {
-            self.recon.insert(evs).await?;
+
+        let batch = self.recon.insert(evs).await.context("persisting all")?;
+        if !batch.invalid.is_empty() {
+            if let Ok(cnt) = batch.invalid.len().try_into() {
+                self.recon.metrics().record(&InvalidEvents(cnt))
```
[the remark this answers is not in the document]
dav1do: I was on the fence about it.. I added a commit with labels if we like that better than simply a count (similar to how we label the message metrics)

stbrody (approved):
LGTM

PR #475, opened 2024-08-14
fix: add better TileDocument detection

nathanielc:
TileDocument init events are not required to have a data field, in which case they must have a header field. TileDocument data events are opposite, they are required to have a data field and the header field is optional.

### DIFF
--- a/service/src/event/migration.rs
+++ b/service/src/event/migration.rs
@@ -175,8 +175,8 @@ impl<'a, S: BlockStore> Migrator<'a, S> {
         let payload: unvalidated::Payload<Ipld> = serde_ipld_dagcbor::from_slice(&payload_data)
             .context("decoding payload")
             .map_err(|err| {
-                if self.is_tile_doc(&payload_data) {
-                    anyhow!("found Tile Document, skipping")
+                if self.is_tile_doc_data(&payload_data) {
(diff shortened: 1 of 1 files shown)


service/src/event/migration.rs
```diff
@@ -219,16 +219,29 @@ impl<'a, S: BlockStore> Migrator<'a, S> {
...
         #[derive(Debug, Deserialize)]
-        struct TileDocPayload {
+        struct TileDocInitPayload {
+            header: Ipld,
+        }
+        serde_ipld_dagcbor::from_slice::<TileDocInitPayload>(data).is_ok()
+    }
+    fn is_tile_doc_data(&self, data: &[u8]) -> bool {
+        // Attempt to decode the payload as a loose TileDocument.
+        // If we succeed produce a meaningful error message.
```
dav1do, line 235: does the comment `// If we succeed produce a meaningful error message` still apply here and above?

dav1do (approved):
One nit/question about a comment. Looks like you have to remove the HashMap import to satisfy clippy and then g2g.

service/src/event/migration.rs
nathanielc: Removed comment, thanks

PR #476, opened 2024-08-14
feat: anchor service (part 2)

smrz2001:
This is the first PR for self-anchoring and includes:
- Anchor service machinery
- Merkle tree building
- Time Event building
- Traits for:
  - Anchor client (e.g. Event Service)
  - Transaction managers (e.g. remote CAS)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ resolver = "2"
 members = [
     "api",
     "api-server",
+    "anchor-service",
     "car",
     "core",
     "event",
@@ -53,6 +54,7 @@ bs58 = "0.4"
 bytecheck = "0.6.7"
 bytes = "1.1"
 bytesize = "1.1"
+ceramic-anchor-service = { path = "./anchor-service" }
 ceramic-api = { path = "./api" }
 ceramic-api-server = { path = "./api-server" }
 ceramic-car = { path = "./car" }
@@ -101,6 +103,7 @@ http-serde = "1.1"
 humansize = "2"
 hyper = { version = "0.14", features = ["full"] }
 ignore = "0.4.18"
+indexmap = "2.3.0"
 indicatif = "0.17.1"
 integer-encoding = "3.0"
 ipld-core = "0.4"
@@ -110,7 +113,7 @@ iroh-p2p = { version = "0.2.0", path = "./beetle/iroh-p2p" }
 iroh-rpc-client = { path = "./beetle/iroh-rpc-client" }
 iroh-rpc-types = { path = "./beetle/iroh-rpc-types" }
 iroh-util = { path = "./beetle/iroh-util" }
-itertools = "0.12.0"
+itertools = "0.13.0"
 k256 = "0.13"
 keyed_priority_queue = "0.4.1"
 lazy_static = "1.4"
@@ -226,6 +229,9 @@ authors = [
     "Danny Browning <dbrowning@3box.io>",
     "Nathaniel Cook <nvcook42@gmail.com>",
     "Aaron D Goldman <goldmanaaron@gmail.com>",
+    "Mohsin Zaidi <@smrz2001>",
+    "David Estes <@dav1do>",
+    "Spencer T Brody <@stbrody>"
 ]
 license = "Apache-2.0/MIT"
 repository = "https://github.com/3box/rust-ceramic"
--- a/anchor-service/Cargo.toml
+++ b/anchor-service/Cargo.toml
@@ -1,0 +1,23 @@
+[package]
+name = "ceramic-anchor-service"
+version.workspace = true
+edition.workspace = true
+authors.workspace = true
+license.workspace = true
+repository.workspace = true
+publish = false
+
+[dependencies]
+anyhow.workspace = true
+async-trait.workspace = true
+ceramic-core.workspace = true
+ceramic-event.workspace = true
+cid.workspace = true
+expect-test.workspace = true
+indexmap.workspace = true
+serde.workspace = true
+serde_ipld_dagjson.workspace = true
+tokio.workspace = true
+
+[features]
+test-network = []
--- a/anchor-service/src/anchor_batch.rs
+++ b/anchor-service/src/anchor_batch.rs
@@ -1,0 +1,59 @@
+use std::sync::Arc;
+
+use anyhow::Result;
+use async_trait::async_trait;
+
+use crate::{
+    anchor::{AnchorRequest, TimeEventBatch},
+    merkle_tree::{build_merkle_tree, MerkleTree},
+    time_event::build_time_events,
+    transaction_manager::{Receipt, TransactionManager},
+};
+
+/// ceramic_anchor_service::Store is responsible for fetching AnchorRequests and storing TimeEvents.
+#[async_trait]
+pub trait Store: Send + Sync {
+    /// Get a batch of AnchorRequests.
+    async fn get_anchor_requests(&self) -> Result<Vec<AnchorRequest>>;
+    /// Store a batch of TimeEvents.
+    async fn put_time_events(&self, batch: TimeEventBatch) -> Result<()>;
+}
+
+/// An AnchorService is responsible for anchoring batches of AnchorRequests and storing TimeEvents generated based on
+/// the requests and the anchor proof.
+pub struct AnchorService {
+    tx_manager: Arc<dyn TransactionManager>,
+}
+
+impl AnchorService {
+    /// Create a new AnchorService.
+    pub fn new(tx_manager: Arc<dyn TransactionManager>) -> Self {
+        Self { tx_manager }
+    }
+
+    /// Anchor a batch of requests using a Transaction Manager:
+    /// - Build a MerkleTree from the anchor requests
+    /// - Anchor the root of the tree and obtain a proof from the Transaction Manager
+    /// - Build TimeEvents from the anchor requests and the proof
+    ///
+    /// This function will block until the proof is obtained from the Transaction Manager.
+    pub async fn anchor_batch(&self, anchor_requests: &[AnchorRequest]) -> Result<TimeEventBatch> {
+        let MerkleTree {
+            root_cid,
+            nodes,
+            count,
+        } = build_merkle_tree(anchor_requests)?;
+        let Receipt {
+            proof,
+            detached_time_event,
+            mut remote_merkle_nodes,
+        } = self.tx_manager.make_proof(root_cid).await?;
+        let time_events = build_time_events(anchor_requests, &detached_time_event, count)?;
+        remote_merkle_nodes.extend(nodes);
+        Ok(TimeEventBatch {
+            merkle_nodes: remote_merkle_nodes,
+            proof,
+            time_events,
+        })
+    }
+}
--- a/anchor-service/src/lib.rs
+++ b/anchor-service/src/lib.rs
@@ -1,0 +1,11 @@
+//! This crate include all the machinery needed for building Merkle Trees from Anchor Requests and then the Time Events
+//! corresponding to those requests once the root of the tree has been anchored.
+#![warn(missing_docs)]
+mod anchor;
+mod anchor_batch;
+mod merkle_tree;
+mod time_event;
+mod transaction_manager;
+
+pub use anchor_batch::{AnchorService, Store};
+pub use transaction_manager::{DetachedTimeEvent, Receipt, TransactionManager};
--- a/anchor-service/src/merkle_tree.rs
+++ b/anchor-service/src/merkle_tree.rs
@@ -1,0 +1,83 @@
+use crate::anchor::{AnchorRequest, MerkleNode, MerkleNodes};
+use anyhow::{anyhow, Result};
+use ceramic_core::SerializeExt;
+use cid::Cid;
+
+pub struct MerkleTree {
+    pub root_cid: Cid,
+    pub nodes: MerkleNodes,
+    pub count: u64,
+}
+
+/// Make a tree using Merkle mountain range
+/// Ref: https://eprint.iacr.org/2021/038.pdf
+pub fn build_merkle_tree(anchor_requests: &[AnchorRequest]) -> Result<MerkleTree> {
+    // For size zero trees return an error
+    if anchor_requests.is_empty() {
+        return Err(anyhow!("no requests to anchor"));
+    }
+    // The roots of the sub-trees with full power of 2 trees.
+    // They are in the places in the array where the 1s are in the count u64.
+    // e.g. 13 = 0b1110
+    //                root
+    //         /                    \
+    //     /      \            /        \
+    //    / \    /  \       /     \     / \
+    //   /\  /\ / \ / \   / \   /   \  12 13
+    //  0 1 2 3 4 5 6 7   8 9  10   11
+    //
+    // here the peeks are [none, 12..13, 8..11, 0..7]
+    // place values         1's,    2's,   4's,  8's
+    let mut peaks: Vec<Option<Cid>> = vec![None; 64];
+
+    // The nodes in the Merkle Map[node_cid, [left_cid, right_cid]]
+    let mut nodes = MerkleNodes::default();
+
+    // insert all the `anchor_request.prev`s into the peaks.
+    for anchor_request in anchor_requests {
+        let mut new_node_cid: Cid = anchor_request.prev;
+        for peek in peaks.iter_mut() {
+            // walk the place values
+            match peek {
+                None => {
+                    // if the place values peek is empty put the cid there
+                    *peek = Some(new_node_cid);
+                    // we found a place to put it. we are done
+                    break;
+                }
+                Some(old_node_cid) => {
+                    // if the place values peek is occupied add the old cid to the new cid and carry to the next place values peek
+                    let merged_node: MerkleNode;
+                    (new_node_cid, merged_node) = merge_nodes(*old_node_cid, new_node_cid)?;
+                    // remember the generated nodes
+                    nodes.insert(new_node_cid, merged_node);
+                    // clear the place value peek we took old node from
+                    *peek = None;
+                }
+            }
+        }
+    }
+
+    // Roll up the peaks into a root.
+    // Since each tree is larger then all the preceding trees combined
+    // we just walk the non empty peeks putting the last peek on the left.
+    let mut peaks_iter = peaks.into_iter().flatten();
+    let mut right_cid = peaks_iter.next().expect("should never be empty");
+    for left_cid in peaks_iter {
+        let merged_node: MerkleNode;
+        (right_cid, merged_node) = merge_nodes(left_cid, right_cid)?;
+        nodes.insert(right_cid, merged_node);
+    }
+
+    Ok(MerkleTree {
+        root_cid: right_cid,
+        count: anchor_requests.len() as u64,
+        nodes,
+    })
+}
+
+/// Accepts the CIDs of two blocks and returns the CID of the CBOR list that includes both CIDs.
+pub(crate) fn merge_nodes(left: Cid, right: Cid) -> Result<(Cid, MerkleNode)> {
+    let merkle_node = vec![Some(left), Some(right)];
+    Ok((merkle_node.to_cid()?, merkle_node))
+}
--- a/anchor-service/src/time_event.rs
+++ b/anchor-service/src/time_event.rs
@@ -1,0 +1,170 @@
+use anyhow::{anyhow, Result};
+
+use ceramic_event::unvalidated::RawTimeEvent;
+
+use crate::{
+    anchor::{AnchorRequest, TimeEvents},
+    DetachedTimeEvent,
+};
+
+pub fn build_time_events(
+    anchor_requests: &[AnchorRequest],
+    detached_time_event: &DetachedTimeEvent,
+    count: u64,
+) -> Result<TimeEvents> {
+    let events = anchor_requests
+        .iter()
+        .enumerate()
+        .map(|(index, anchor_request)| {
+            let local_path = index_to_path(index.try_into()?, count)?;
+            let remote_path = detached_time_event.path.as_str();
+            Ok(RawTimeEvent::new(
+                anchor_request.id,
+                anchor_request.prev,
+                detached_time_event.proof,
+                format!("{}/{}", remote_path, local_path)
+                    .trim_matches('/')
+                    .to_owned(),
+            ))
+        })
+        .collect::<Result<Vec<_>>>()?;
+
+    Ok(TimeEvents { events })
+}
+
+pub fn index_to_path(index: u64, count: u64) -> Result<String> {
+    // we want to find the path to the index in a tree length.
+    // first find the sub-tree then append the path in the sub-tree
+    //
+    // e.g. 5, 6
+    // 14 = 0b1110
+    // 10 = 0b1010
+    //               [root]
+    //         /                  [\]
+    //     /      \           [/]      \
+    //    / \    /  \       /    [\]    / \
+    //   /\  /\ / \ / \   / \  [/]  \  12 13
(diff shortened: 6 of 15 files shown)


Cargo.toml
```diff
@@ -213,6 +217,8 @@ authors = [
     "Danny Browning <dbrowning@3box.io>",
     "Nathaniel Cook <nvcook42@gmail.com>",
     "Aaron D Goldman <goldmanaaron@gmail.com>",
+    "Mohsin Zaidi <@smrz2001>",
+    "David Estes <@dav1do>"
```
smrz2001, line 221: Please add more names here as suggestions, if interested.

anchor-service/src/time_event.rs
```diff
@@ -0,0 +1,162 @@
...
+                index.try_into()?,
+                count,
+            )
+        })
+        .collect::<Result<Vec<_>>>()?;
+
+    Ok(TimeEvents { events })
+}
+
+pub fn build_time_event(
```
dav1do, line 34: I wonder if it's worth adding an input struct here to make it slightly harder to mismatch all the cids.. it's still a bunch of cids but it was something like one of these two might be less error prone (if a bit tedious)

```rust
    struct BuildTimeEventInput<'a> {
        pub id: &'a Cid,
        pub prev: &'a Cid,
        pub root: &'a Cid,
        pub remote_path: &'a str,
        pub index: u64,
        pub count: u64,
    }

    struct BuildTimeEventInput2<'a> {
        pub id: Cid,
        pub prev: Cid,
        pub root: Cid,
        pub remote_path: &'a str,
        pub index: u64,
        pub count: u64,
    }

    fn build_it(input: BuildTimeEventInput) {
        todo!()
    }
    fn build_it2(input: &BuildTimeEventInput2) {
        todo!()
    }
    build_it(BuildTimeEventInput {
        id: &id,
        prev: &prev,
        root: &proof,
        remote_path: "",
        index,
        count,
    });
    // only need to assign input here if we still want access to the cids after
    let input = BuildTimeEventInput2 {
        id,
        prev,
        root: proof,
        remote_path: "",
        index,
        count,
    };

    build_it2(&input);
 ```

**Review by dav1do — COMMENTED** [review]
I didn't think too hard about the tree building stuff but poked through the tests. A few comments but no major concerns.

File: anchor-service/Cargo.toml [hunk 3]
```diff
@@ -0,0 +1,25 @@
+[package]
```
nathanielc, line 1: Can you add the `/anchor-service/` dir to the CODEOWNERS file so we establish reviewers for the rest of this code.

event/src/lib.rs
```diff
@@ -1,6 +1,9 @@
 //! # Ceramic Event
 //! Implementation of ceramic event protocol, with appropriate compatibilility with js-ceramic
 #![warn(missing_docs)]
+
+/// Anchor-related types
+pub mod anchor;
```
nathanielc, line 6: Why do anchor types live in the `ceramic-event` crate? I would have expected them to live in the `ceramic-anchor-service` crate.

event/src/unvalidated/event.rs
```diff
@@ -295,10 +295,14 @@ impl TimeEvent {
...
-    proof: Cid,
-    path: String,
+    /// The CID of the init event of the stream
+    pub id: Cid,
+    /// The CID of the Data event that is being anchored in the chain
+    pub prev: Cid,
+    /// The CID of the proof block that tells us how to query the chain
+    pub proof: Cid,
+    /// path from the root in the proof block to the prev in the merkle tree
+    pub path: String,
```
nathanielc, line 305: These should not be public. They are intentionally private so that they cannot be used to accidentally construct nonsense events.

anchor-service/src/time_event.rs
```diff
@@ -0,0 +1,162 @@
...
+    id: &Cid,
+    prev: &Cid,
+    proof_cid: &Cid,
+    remote_path: &str,
+    index: u64,
+    count: u64,
+) -> Result<RawTimeEvent> {
+    let local_path = index_to_path(index, count)?;
+    Ok(RawTimeEvent {
```
nathanielc, line 43: Why do we not use the builder API here?

event/src/anchor.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub prev: Cid,
+}
+
+/// Merkle tree node
+pub type MerkleNode = Vec<Cid>;
+
+/// A collection of Merkle tree nodes.
+#[derive(Default)]
+pub struct MerkleNodes {
+    nodes: IndexMap<Cid, MerkleNode>,
```
nathanielc, line 21: Curious on the choice of an IndexMap? Should we be using an IndexMap anytime we have CID keys?

anchor-service/src/anchor_batch.rs
```diff
@@ -0,0 +1,99 @@
...
+    merkle_tree::{build_merkle_tree, MerkleTree},
+    time_event::build_time_events,
+    transaction_manager::{Receipt, TransactionManager},
+};
+
+#[async_trait]
+pub trait AnchorClient: Send + Sync {
+    async fn get_anchor_requests(&self) -> Result<Vec<AnchorRequest>>;
+    async fn put_time_events(&self, batch: TimeEventBatch) -> Result<()>;
+}
```
nathanielc, line 22: nit, can you add documentation to all the methods since this is public.

anchor-service/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+mod anchor_batch;
```
nathanielc, line 1: Can you add `#![warn(missing_docs)]` here so we document all public interfaces from this crate.

anchor-service/src/anchor_batch.rs
```diff
@@ -0,0 +1,99 @@
...
+use ceramic_event::anchor::{AnchorRequest, TimeEventBatch};
+
+use crate::{
+    merkle_tree::{build_merkle_tree, MerkleTree},
+    time_event::build_time_events,
+    transaction_manager::{Receipt, TransactionManager},
+};
+
+#[async_trait]
+pub trait AnchorClient: Send + Sync {
```
nathanielc, line 19: Who is expected to implement this trait? Something from the ceramic-service crate?

If so I think we should slightly tweak the semantics of the trait. Instead of AnchorClient we should call this FeedClient and instead of `get_anchor_requests` we should have `get_events(FeedOpts)` where FeedOpts lets you decide which kind of events to get from the Feed. This way the `ceramic-service` does not have to know or understand anything about anchors. Rather it remains specific to Ceramic events.  

This likely also means that the AnchorRequest type can remain private to the anchor-service crate.


Thoughts?

nathanielc (changes requested):
Looking good, have some questions below.

anchor-service/src/anchor_batch.rs
```diff
@@ -0,0 +1,99 @@
...
+            anchor_client,
+            tx_manager,
+            batch_linger_time,
+        }
+    }
+
+    pub async fn run(&mut self) {
+        let mut interval = interval(self.batch_linger_time);
+        loop {
+            interval.tick().await;
```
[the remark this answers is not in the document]
AaronGoldman: yes we want the  first attempt right away

event/src/anchor.rs
AaronGoldman: This is to keep the merkel tree nodes in order.  A hash map will randomize the order.
Non-deterministic order makes testing hard.
Sorted order you will need to use later tree nodes to use earlier ones. Not nesaraly a problem just pulling more nodes into memory as we validate the time events.

stbrody:
Holding off on reviewing this until the first round of comments from Nathaniel and David are addressed, but would love to review this before it goes in

event/src/anchor.rs
nathanielc: Ok so why not a simple BTreeMap in that case?

nathanielc: Actually I think I misunderstood _order_. We need to keep the nodes in insertion order not just any stable order?

anchor-service/src/merkle_tree.rs
```diff
@@ -0,0 +1,69 @@
...
+    pub root_cid: Cid,
+    pub nodes: MerkleNodes,
+    pub count: u64,
+}
+
+pub fn build_merkle_tree(anchor_requests: &[AnchorRequest]) -> Result<MerkleTree> {
+    if anchor_requests.is_empty() {
+        return Err(anyhow!("no requests to anchor"));
+    }
+    let mut peaks: [Option<Cid>; 64] = [None; 64];
```
stbrody, line 17: what are "peaks" here?  TBH I feel like this whole function could use a lot more comments explaining what it's doing

event/src/anchor.rs
```diff
@@ -0,0 +1,86 @@
...
+/// AnchorRequest for a Data Event on a Stream
+pub struct AnchorRequest {
+    /// The CID of the stream
+    pub id: Cid,
+    /// The CID of the Event to be anchored
+    pub prev: Cid,
+}
+
+/// Merkle tree node
+pub type MerkleNode = Vec<Cid>;
```
stbrody, line 16: what about a node that has only has a child in one direction?  Don't we really need each node to be a pair of Option<Cid> so we can also use None when one branch is empty?

event/src/anchor.rs
stbrody, line 21: comment on what the relationship is between the key and the value please

anchor-service/src/merkle_tree.rs
```diff
@@ -0,0 +1,69 @@
...
+        // [none, (cid1, cid2)]
+        // [cid3, (cid1, cid2)]
+        // [none, none, ((cid1, cid2), (cid3, cid4))]
+        // [cid5, none, ((cid1, cid2), (cid3, cid4))]
+        // [none, (cid5, cid6), ((cid1, cid2), (cid3, cid4))]
+        // [cid7, (cid5, cid6), ((cid1, cid2), (cid3, cid4))]
+        // [none, none, none, (((cid1, cid2), (cid3, cid4)), ((cid5, cid6), (cid7, cid8))]
+        let mut new_node_cid: Cid = anchor_request.prev;
+        let mut place_value = 0;
+        while place_value < peaks.len() {
```
stbrody, line 34: I find this code confusing to understand.

Is there a reason we're using some fancy algorithm here instead of just the simple recursive balanced binary tree building algorithm taught in every college data structures class?  We know all the events up front, we don't need a streaming algorithm, and I think there's value in keeping this really simple and obviously correct for now.  If we want to move to a streaming approach to tree building we can do so in the future, but doing that would require changes to a lot more than just this function, so as long as we're building the whole tree in one go I'd strongly vote to keep the algorithm dead simple for now.

stbrody (changes requested):
did a quick skim over this for some initial thoughts.  A big overarching request is more comments throughout.

anchor-service/src/anchor_batch.rs
```diff
@@ -0,0 +1,99 @@
...
+                        error!("error getting anchor requests: {:?}", e);
+                        continue;
+                    }
+                };
+            if anchor_requests.is_empty() {
+                info!("no requests to anchor");
+                continue;
+            }
+            // Anchor the batch to the CAS. This may block for a long time.
+            match self.anchor_batch(anchor_requests.as_slice()).await {
```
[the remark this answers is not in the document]
AaronGoldman: Spawn a std::thread is higher up in the logic.  their will be an anchor loop that is once an hour calling the event service to get all unanchored events and do the anchoring. within that loop it is fine for everything to be serial.

1. get the anchor requests
2. build the tree
3. send root to CAS / ETH
4. poll for transaciton completion
5. build the proof.
5.1. verify proof. not implemented 
6. build the time events.
7. store the time events.
8. sleep until next hour tick. 

One tikio thread going aroind and around in the circle.

event/src/unvalidated/event.rs
AaronGoldman: RawTimeEvent new is just a pass through and there are getters for every field. The fields are not optional. 
 RawTimeEvent is just a data class I don't see why we would not just have all public fields and no functions on the data class. except for `impl Debug for RawTimeEvent`

event/src/anchor.rs
AaronGoldman: for parsing yes for building the tree we don't have any not full nodes.

stbrody: how is that possible?  What if you have a number of events that isn't a power of 2?  Won't you always have some unbalanced nodes?

AaronGoldman: image
The depth may not be balanced but there are no nulls.
The peeks are the place values so
peek[0] has 1 anchor requested CID or a null.
peek[1] has 2 anchor requested CID or a null.
peek[2] has 4 anchor requested CID or a null.
peek[3] has 8 anchor requested CID or a null.
peek[n] has 2**n anchor requested CID or a null.

then we just role up the side merging the pow2 trees.
If we kept the nulls we would get balanced depth but nulls in the roll up nodes.
Since we don't bother with the nodes that have an empty branch we get no nulls but shorter paths where the half empty nodes would have been,

anchor-service/src/anchor_batch.rs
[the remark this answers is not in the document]
smrz2001: Looks like the first tick always completes immediately (from the `interval` docs):
```
    /// use tokio::time;
    ///
    /// use std::time::Duration;
    ///
    /// #[tokio::main]
    /// async fn main() {
    ///     let mut interval = time::interval(Duration::from_millis(10));
    ///
    ///     interval.tick().await;
    ///     // approximately 0ms have elapsed. The first tick completes immediately.
    ///     interval.tick().await;
    ///     interval.tick().await;
    ///
    ///     // approximately 20ms have elapsed.
    /// }
```

anchor-service/src/time_event.rs
smrz2001: We went ahead and inlined the function.

core/src/block.rs
```diff
@@ -0,0 +1,46 @@
+use cid::Cid;
+use ipld_core::{codec::Codec, ipld::Ipld};
+use multihash_codetable::{Code, MultihashDigest};
+use serde_ipld_dagcbor::codec::DagCborCodec;
+
+/// A DagCbor block with a CIDv1 and the raw bytes.
+#[derive(Clone, PartialEq)]
+pub struct DagCborIpfsBlock {
```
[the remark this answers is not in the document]
smrz2001: Yes, we removed the file.

event/src/anchor.rs
smrz2001: Yeah, the order here is only used for debugging and expect tests. It makes it easier for us to know where there are issues because the remote nodes and local nodes remain in their respective groupings.

anchor-service/src/anchor_batch.rs
smrz2001: This feels worth a design discussion.

event/src/anchor.rs
stbrody: Ah this is still with the MMR approach, I was assuming we switched to the simpler balanced tree building approach

anchor-service/src/merkle_tree.rs
smrz2001: Yes, will add more comments soon!

smrz2001: Added comments from @AaronGoldman that we'd put in a later PR.

anchor-service/src/merkle_tree.rs
stbrody: Just spoke about this with @smrz2001 and he realized that because we need to replace older events with newer events on the same stream before we build the tree, that means the set of events that go into the tree isn't even strictly append-only.  Which means it's not even totally clear how we could *ever* do streaming tree building.  So it seems that using the MMR algorithm is not giving us any benefit now and is fairly likely to *never* give us any benefit.  So I fairly strongly think we should rip out this MMR approach and just put in the extremely simple balanced tree building algorithm instead.

AaronGoldman: You mean something like 
```rs
fn build_merkle_tree_r(anchor_requests: &[AnchorRequest], nodes: &mut MerkleNodes) -> Result<Cid> {
    if anchor_requests.len() == 0 {
        return Err(anyhow!("Empty anchor_requests"));
    }
    if anchor_requests.len() == 1 {
        return Ok(anchor_requests[0].prev);
    }
    let top_power_of_2 = 1 << (63 - anchor_requests.len().leading_zeros());
    let split_point;
    if anchor_requests.len() == top_power_of_2 {
        split_point = top_power_of_2 / 2;
    } else {
        split_point = top_power_of_2;
    }
    let left = build_merkle_tree_r(&anchor_requests[..split_point], nodes)?;
    let right = build_merkle_tree_r(&anchor_requests[split_point..], nodes)?;
    let node = vec![Some(left),Some(right)];
    let node_cid = node.to_cid()?;
    nodes.insert(node_cid, node);
    Ok(node_cid)
}
```
That should build the same tree.

anchor-service/src/time_event.rs
AaronGoldman: We have a proof object CID and the builder API takes the fields. We would need to parse the proof block and plum through the fields from the CAS response just to eventually re assemble in the builder api and calculate the same CID. Assuming it round trips witch it should.

nathanielc (changes requested):

anchor-service/src/anchor_batch.rs
```diff
@@ -0,0 +1,118 @@
...
+    transaction_manager::{Receipt, TransactionManager},
+};
+
+/// An AnchorClient is responsible for fetching AnchorRequests and storing TimeEvents.
+#[async_trait]
+pub trait Store: Send + Sync {
+    /// Get a batch of AnchorRequests.
+    async fn get_anchor_requests(&self) -> Result<Vec<AnchorRequest>>;
+    /// Store a batch of TimeEvents.
+    async fn put_time_events(&self, batch: TimeEventBatch) -> Result<()>;
```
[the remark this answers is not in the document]
AaronGoldman: The `impl Store for CeramicEventService` is in part 5. That is where I would like to make the decision on where we hydrate to time events and who is responsible for storing the high water mark. As part 5 is written today the high water mark is stored in the event service we will probably store that somewhere else.

`get_anchor_request` is renamed  `local_sourced_data_events` in part 3 but the real logical isolation will probably be in part 5 when the decision about where to store the high water mark drives the final trait definition.

https://github.com/ceramicnetwork/rust-ceramic/pull/488/files#diff-148a932282cead8137f562abc4a57762ed173ccebd8d74361766ec0df7948d22R424

`$1`  new_high_water_mark (i64)
```sql
INSERT INTO ceramic_one_anchor_high_water_mark (high_water_mark) VALUES ($1);
```

`$1` did:key for the local node
`$2` the limit
```sql
SELECT order_key, init_cid, cid, rowid
FROM ceramic_one_event
WHERE
  source = $1
  AND ROWID > (SELECT MAX(high_water_mark) FROM ceramic_one_anchor_high_water_mark)
ORDER BY rowid
LIMIT $2;
```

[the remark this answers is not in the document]
nathanielc: >The impl Store for CeramicEventService is in part 5. That is where I would like to make the decision on where we hydrate to time events and who is responsible for storing the high water mark.

Why does the ceramic-service what to store the high water mark for the anchor service? I would expect the anchor service to have its own place to store the highwater mark separate from the ceramic-service. Perhaps the name `anchor_service::Store` is incorrect here in that it implies its the interface for storing state on behalf of the anchor service. I see this trait as a way access to the event service. Perhaps we should call this trait `EventService`? 

And do we need another mechnism for storing data on behalf of the anchor service. For example a new sqlite connection and its own anchor store struct that can store and retrieve the high water mark?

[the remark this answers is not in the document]
nathanielc: To expand a bit on the architecture I am after...

I would like to have a set of services and each service manages its own state (i.e. database). The services can depend on each other via public APIs etc. In this PR we are adding a new service that depends on the event service. This means the anchor service needs it a way to manage its own state. It should not delegate management of its state to the event service. The anchor service has very little state that it needs to store (i.e. only the high water mark). However that still means that there should be some kind of struct private to the anchor service crate that manages storing and retrieving the high water mark. I expect this is a sqlite table, and it can be the same sqlite connection that the event service uses but the logic to store and retrieve that state lives inside the anchor service.

When we inevitably want to migrate how we store the high water mark, we will have to modify any code that access that storage directly. Its best if all of that code lives behind the anchor service API boundary. 

As the PR is currently structured its intended that logic lives inside the ceramic-service crate and I explicitly do not want that architecture.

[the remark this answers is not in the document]
AaronGoldman: * Part 2: build trees from anchor requests and build Time Events from proofs.
* Part 3: use the remote CAS to make proofs from root CIDs. All the auth and api stuff to talk to unmodified CAS.
* Part 4: all the changes to EventService so it can find the events that anchor service wants to anchor.  e.g. source and StreamID for events.
* Part 5: starting the loop, command line args, persisting Time events and High Water Mark

I think we can answer those questions in part 5 this is mostly here as a `todo!(get_anchor_requests, put_time_events)`
and to have something to for the MockStore so I could implement and test trees building and Time event building.

anchor-service/src/time_event.rs
```diff
@@ -0,0 +1,170 @@
...
+        .collect::<Result<Vec<_>>>()?;
+
+    Ok(TimeEvents { events })
+}
+
+pub fn index_to_path(index: u64, count: u64) -> Result<String> {
+    // we want to find the path to the index in a tree length.
+    // first find the sub-tree then append the path in the sub-tree
+    //
+    // e.g. 5, 6
```
stbrody, line 39: I don't quite follow this.  I think the example text below the diagram is saying that the example is with `index` as 10 and `count` as 14?  Can we make that explicit in the comment?  Also, what does the "e.g. 5,6" mean here?

anchor-service/src/merkle_tree.rs
```diff
@@ -0,0 +1,83 @@
...
+
+pub struct MerkleTree {
+    pub root_cid: Cid,
+    pub nodes: MerkleNodes,
+    pub count: u64,
+}
+
+/// Make a tree using Merkle mountain range
+/// Ref: https://eprint.iacr.org/2021/038.pdf
+pub fn build_merkle_tree(anchor_requests: &[AnchorRequest]) -> Result<MerkleTree> {
```
stbrody, line 14: is there any unit test coverage of this?  The MMR logic here is complex enough that I think some targeted unit tests would make me feel a lot better about it.

stbrody (approved):
main comment is to add a unit test for merkle tree building, otherwise lgtm

anchor-service/src/anchor_batch.rs
```diff
@@ -0,0 +1,59 @@
...
+    merkle_tree::{build_merkle_tree, MerkleTree},
+    time_event::build_time_events,
+    transaction_manager::{Receipt, TransactionManager},
+};
+
+/// ceramic_anchor_service::Store is responsible for fetching AnchorRequests and storing TimeEvents.
+#[async_trait]
+pub trait Store: Send + Sync {
+    /// Get a batch of AnchorRequests.
+    async fn get_anchor_requests(&self) -> Result<Vec<AnchorRequest>>;
```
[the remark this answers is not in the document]
AaronGoldman: Both actually. The high water mark will be in a table owned by the anchor service and the get data events that originate from this node less then high water mark will be calling a function on  CeramicEventService.

anchor-service/src/merkle_tree.rs
smrz2001: We were in the middle of merging this PR when we saw this comment :(

Create a ticket under SA Ph1 for this:
https://linear.app/3boxlabs/issue/AES-343/add-more-exhaustive-merkle-tree-tests

PR #477, opened 2024-08-14
feat: trait for generating cbor bytes and cids (part 1)

smrz2001:


dav1do (commented):
IMO using a proc macro for this is overkill (plus it won't work for lifetimes/generics)... a simple derive macro would work and makes it simpler to understand... e.g. this with the error type as a param if desired

```rust
#[derive(serde::Serialize, serde::Deserialize)]
struct Dog {
    s: String,
}

macro_rules! derive_serde_ipld {
    ($name: ident, $err: ty) => {
        impl $name {
            /// serde serialize to dag-json
            pub fn to_json(&self) -> Result<String, Box<$err> {
                 let json_bytes = serde_ipld_dagjson::to_vec(self)?;
                 String::from_utf8(json_bytes).map_err(Into::into)
            }

            /// serde serialize to dag-cbor
            pub fn to_cbor(&self) -> Result<Vec<u8>, $err> {
                serde_ipld_dagcbor::to_vec(self).map_err(Into::into)
            }

            /// CID of serde serialize to dag-cbor
            pub fn to_cid(&self) -> Result<Cid, $err> {
                let cbor_bytes = self.to_cbor()?;
                use multihash_codetable::MultihashDigest;
                let hash = multihash_codetable::Code::Sha2_256.digest(&cbor_bytes);
                Ok(Cid::new_v1(0x71, hash))
            }
        }
    };
}
derive_serde_ipld!(Dog, Box<dyn std::error::Error>);
```

event/src/unvalidated/mod.rs
```diff
@@ -52,4 +60,42 @@ mod tests {
...
 
     pub const UNSIGNED_INIT_NO_SEP_CAR:&str="uOqJlcm9vdHOB2CpYJQABcRIgrY2L_wTWrzng7Mpf2kvh9Q9Uyz-Ei2CF7NfjzRD0illndmVyc2lvbgG5AwFxEiCtjYv_BNavOeDsyl_aS-H1D1TLP4SLYIXs1-PNEPSKWaJkZGF0YaZkZGF0YaNjdXJsYGVsYWJlbGdGYXN0aW5nbmNoaWxkcmVuSGlkZGVu9GR0eXBlbFF1ZXN0aW9uTm9kZWdjcmVhdGVkeBgyMDIzLTAyLTIwVDE1OjE5OjM2LjI3OVpocG9zaXRpb26iYXj7QKWDkiAAAABheftAtqeAAAAAAGlsYXRlcmFsSUR4JDY5YWFmMzdkLTU1OWItNGI5NS1hMTAwLTVlZTk4MThmYzVlZGlwcm9qZWN0SUR4P2tqemw2a2N5bTd3OHk1ZDY1Zjlva240cml5ZGF0OTJoM3M2dnZ6cHd3dTc1NDRpOTJmanlnYzU2N2x6aHJ2Y2ZoZWFkZXKjZW1vZGVsWCjOAQIBhQESIJbss9X3kzfag-eF6OFx0KDIOV6P5DxOi-cWAWnTN5FjZnVuaXF1ZUwq2_OJxYVaSoawmJtrY29udHJvbGxlcnOBeDtkaWQ6cGtoOmVpcDE1NToxOjB4Y2E1ZmY0YjM0NDJmY2FjMjdlMWFmNDQ1N2UwMmViNjI5YzcxMjk4Mw";
+
+    #[derive(Serialize, Deserialize, SerdeIpld)]
```
nathanielc, line 64: The name `SerdeIpld` is confusing as it makes me think we are deriving some traits from serde.

macro-ipld-derive/src/lib.rs
```diff
@@ -0,0 +1,35 @@
...
+
+#[proc_macro_derive(SerdeIpld)]
+pub fn derive_serde_ipld(input: TokenStream) -> TokenStream {
+    let ast = parse_macro_input!(input as DeriveInput);
+
+    let name = &ast.ident;
+
+    let expanded = quote! {
+            // impl MacroTrait for #name // should we make a trait?
+            impl #name {
```
nathanielc, line 13: Yes we should have a trait here that way we have some context unto which to attach the methods. Having a `to_json` method on a struct is not enough context to know that we are dealing with dag-json specifically. 

Whatever we name the trait should be the name of the derive macro instead of `SerdeIpld`. Then its clear we are deriving the behavior define in that trait.

macro-ipld-derive/src/lib.rs
```diff
@@ -0,0 +1,35 @@
...
+                /// serde serialize to dag-json
+                pub fn to_json(&self) -> Result<String, Box<dyn std::error::Error>> {
+                    let json_bytes = serde_ipld_dagjson::to_vec(self)?;
+                    String::from_utf8(json_bytes).map_err(Into::into)
+                }
+
+                /// serde serialize to dag-cbor
+                pub fn to_cbor(&self) -> Result<Vec<u8>, Box<dyn std::error::Error>> {
```
nathanielc, line 21: I'd prefer not to loose the information about which concrete serde error was produced. Can we make these method return serde errors since we are calling serde methods?

nathanielc (changes requested):
This feels like too much complexity to abstract away calling `serde_ipld_*` crate methods. 

I won't push back super hard but this feels like a lot of code to maintain for a little ergonomic benefit.

core/src/ipld_ext.rs
```diff
@@ -0,0 +1,64 @@
...
+        let cbor_bytes = self.to_cbor()?;
+        let hash = multihash_codetable::MultihashDigest::digest(
+            &multihash_codetable::Code::Sha2_256,
+            &cbor_bytes,
+        );
+        Ok((Cid::new_v1(0x71, hash), cbor_bytes))
+    }
+}
+
+impl<T: Serialize> IpldExt for T {}
```
nathanielc, line 35: This is perfect, and demonstrates the value of a good trait. Now anyone who wants this behavior can import this trait and leverage its value.

core/src/ipld_ext.rs
```diff
@@ -0,0 +1,64 @@
+use anyhow::Result;
+use cid::Cid;
+use serde::Serialize;
+
+/// IpldExt is a trait for serialization, deserialization, CID calculation assuming dagcbor.
+pub trait IpldExt: Serialize {
```
nathanielc, line 6: This name is not quite right... The Rust convention is that a `*Ext` trait is extending a trait with the * name. In this case it would be `SerializeExt`. Maybe we should rename it to `SerializeExt` and given its in `ceramic-core` its clear these are common serialization extensions on ceramic data types? I'd be happy with other names too but since there isn't an `Ipld` trait this is a bit confusing in its current name.

nathanielc (commented):
Overall really like how this has simplified the code. Really good use of a trait. With some consensus on the name this should be good to go

core/src/ipld_ext.rs
```diff
@@ -0,0 +1,64 @@
...
+    /// serde serialize to dag-json String.
+    fn to_json(&self) -> Result<String> {
```
nathanielc, line 8: nit, these methods should return `Result<Self::Ok, Self::Error>` since we are explicitly extending the serde::Serialize trait

nathanielc: https://docs.rs/serde/latest/serde/trait.Serialize.html

core/src/ipld_ext.rs
AaronGoldman: I like `SerializeExt` that way we could add whatever other common Serializations we use.

core/src/serialize_ext.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    /// CID of Serialize to dag-cbor
+    fn to_cid(&self) -> Result<Cid, Self::Error>;
+
+    /// Serialize to dag-cbor and return the CID and bytes
+    fn to_dag_cbor_block(&self) -> Result<(Cid, Vec<u8>), Self::Error>;
+}
+
+impl<T: Serialize> SerializeExt for T {
+    type Error = anyhow::Error;
```
nathanielc, line 23: Why are we using anyhow here?

dav1do: It looks like the ipld crates use two different error types and don't just use a serde error, so we need to provide something that matches both. Should we use an error that passes through the variants or is using anyhow okay?

```rust
// ipld_dag_json
#[derive(Debug)]
pub enum EncodeError {
    Message(String),
}

// ipld_dag_cbor
#[derive(Debug)]
pub enum EncodeError<E> {
    /// Custom error message.
    Msg(String),
    /// IO Error.
    Write(E),
}
```

nathanielc: Lets use [thiserror](https://docs.rs/thiserror/latest/thiserror/) and create a variant for each type of error that can be produced. Thiserror make this very easy. With this approach we expose the underlying errors which is what we want in the context of a library trait.

core/src/serialize_ext.rs
```diff
@@ -0,0 +1,69 @@
+use cid::Cid;
+use serde::Serialize;
+use std::collections::TryReserveError;
+
+/// SerializeExt is a trait for serialization, deserialization, CID calculation assuming dag-cbor.
+pub trait SerializeExt: Serialize {
+    /// Error type for SerializeExt
+    // type Error;
```
nathanielc, line 8: nit, remove these comments.

nathanielc (approved):

PR #478, opened 2024-08-14
feat: remote anchor transaction manager (part 3)

smrz2001:
This PR builds on the [previous PR](https://github.com/ceramicnetwork/rust-ceramic/pull/476) to add a remote CAS transaction manager that communicates with the existing 3Box Labs CAS to anchor Merkle tree roots while hiding from CAS that this is not a normal stream.

The stream ID used for the anchor request is the node public key (from the ED25519 private key) encoded as a CID in an [unloadable](https://github.com/ceramicnetwork/js-ceramic/blob/297e192a6d03147f7ec5343c46d3fb7ff6a69678/packages/streamid/src/stream-type.ts#L6) stream ID.


event/src/unvalidated/event.rs
```diff
@@ -345,9 +345,9 @@ pub struct RawTimeEvent {
 impl Debug for RawTimeEvent {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("RawTimeEvent")
-            .field("id", &self.id.to_string())
-            .field("prev", &self.prev.to_string())
-            .field("proof", &self.proof.to_string())
+            .field("id", &format!("{:?}", &self.id))
```
dav1do, line 348: Probably nicer to use the `write!` macro with the formatter for these debug impls e.g. `write!(f, "{:?}, self.id)`

dav1do: actually, nvm.. overlooked the `debug_struct` restraint... there is `format_args!` but it's basically same same

core/src/key.rs
```diff
@@ -0,0 +1,99 @@
...
+            Ok(key.private.to_bytes())
+        })?;
+    Ed25519KeyPair::from_seed_unchecked(seed.as_ref())
+        .map_err(|e| anyhow::anyhow!("failed to create key pair: {}", e))
+}
+
+/// Create an Ed25519 key pair from a secret
+pub fn ed25519_key_pair_from_secret(secret: &str) -> Result<Ed25519KeyPair> {
+    let mut parts = secret.split(':');
+    let secret_with_prefix: [u8; 34] = multibase::decode(parts.next().unwrap())?
```
dav1do, line 26: can we remove the `unwrap()s` from this function and use the result or `expect`  if it is should never panic?

anchor-tx/src/cas_remote.rs
```diff
@@ -0,0 +1,313 @@
...
+            root.to_string(),
+            loop_count,
+            wait_duration_min
+        ))
+    }
+}
+
+impl RemoteCas {
+    /// Create a new RemoteCas instance
+    pub fn new(keypair: Ed25519KeyPair, cas_api_url: String) -> Self {
```
dav1do, line 144: would using a `DidDocument` here instead of a key make sense? it feels like we have a lot of variations on the crypto primitives already - ssi uses ed25519_dalek (and probably ring somewhere) and I used that one in the validation library (continuing this comment/thought on the key.rs file)

anchor-tx/src/cas_remote.rs
```diff
@@ -0,0 +1,313 @@
...
+    for (cid, block) in car_reader
+        .stream()
+        .into_stream()
+        .try_collect::<Vec<(_, _)>>()
+        .await?
+    {
+        if let Ok(block) = serde_ipld_dagcbor::from_slice::<DetachedTimeEvent>(&block) {
+            detached_time_event = Some(block);
+        }
+        if let Ok(block) = serde_ipld_dagcbor::from_slice::<Proof>(&block) {
```
dav1do, line 209: should these last 2 `if let` statements be `else if let`? seems like only 1 will ever pass

core/src/key.rs
```diff
@@ -0,0 +1,99 @@
+use std::{fs, path::PathBuf, str::FromStr};
+
+use anyhow::{anyhow, Result};
+use cid::multihash::Multihash;
+use cid::Cid;
+use ring::signature::{Ed25519KeyPair, KeyPair};
```
dav1do, line 6: I'm wondering which library we should expose here.. we use [ed25519_dalek](https://docs.rs/ed25519-dalek/latest/ed25519_dalek/) in some places as ssi uses it. It has `SigningKey::from_bytes` (for both secret and public key options) but I'm not sure it's really that much nicer.. mostly just curious what we _should_ be using as our public API.

core/src/lib.rs
```diff
@@ -14,6 +15,10 @@ pub use bytes::Bytes;
 pub use event_id::EventId;
 pub use interest::{Interest, PeerId};
 pub use jwk::Jwk;
+pub use key::{
```
dav1do, line 18: do we really need all of these as core primitives? it seems nicer to make APIs around bytes or DIDs or SigningKeys and have the anchor crate read things from strings and directories and then convert them into whatever format we need to construct things

anchor-tx/src/cas_remote.rs
dav1do: just noticed these are changed to `else if let` in part 5

anchor-tx/src/cas_remote.rs
```diff
@@ -0,0 +1,313 @@
...
+    }
+}
+
+async fn auth_jwt(
+    request_url: String,
+    digest: String,
+    signing_key: &Ed25519KeyPair,
+) -> Result<String> {
+    let controller = did_key_from_ed25519_key_pair(signing_key);
+    let header = Header {
```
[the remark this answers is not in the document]
smrz2001: We did start there but it needed much less code to re-implement this than it took to pull in the `ceramic_core` code 😄

event/src/unvalidated/event.rs
```diff
@@ -384,18 +388,22 @@ impl RawTimeEvent {
 #[derive(Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct Proof {
-    chain_id: String,
-    root: Cid,
-    tx_hash: Cid,
-    tx_type: String,
+    /// eip-155 CHAIN_ID see https://chainid.network https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md
+    pub chain_id: String,
```
dav1do, line 392: do these fields actually need to public? I know there's concern about making invalid events without using the builder.

also, this should parse as a `ssi::caip2::ChainID` but probably better to stay a string

core/src/lib.rs
dav1do: talked about putting these behind a `NodeId` struct (or similar) so they have a home on something that is `core`

anchor-tx/src/cas_remote.rs
```diff
@@ -0,0 +1,325 @@
...
+    nonce: String,
+    url: String,
+    digest: String,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+struct AnchorResponse {
+    pub message: String,
+    pub witness_car: Option<String>,
```
stbrody, line 62: nit: should this be a byte vector instead of a string?

anchor-tx/src/cas_remote.rs
```diff
@@ -0,0 +1,325 @@
...
+            alg: "EdDSA".to_string(),
+            typ: "JWT".to_string(),
+        };
+        let jws_header_b64 =
+            b64.encode(serde_json::to_vec(&jws_header).expect("invalid jws header"));
+        let node_stream_id = cid_to_stream_id(cid_from_ed25519_key_pair(&keypair));
+        Self {
+            signing_key: keypair,
+            url: format!("{}/api/v0/anchor", remote_anchor_service_url),
+            poll_interval: anchor_poll_interval,
```
stbrody, line 142: its weird to me to have the interval configurable but the retry count not?  Given that how long you'll poll for is a function of both.  I feel like we should either use a hard-coded constant for both, or make both configurable

anchor-tx/src/cas_remote.rs
```diff
@@ -0,0 +1,325 @@
...
+            poll_retry_count: 12,
```
stbrody, line 143: if we keep this as a constant, can we make it a named constant at the top of the file?

anchor-tx/src/cas_remote.rs
```diff
@@ -0,0 +1,325 @@
...
+        for _ in 0..self.poll_retry_count {
+            let anchor_response = self.create_anchor_request(root).await?;
+            match parse_anchor_response(anchor_response).await {
+                Ok(CasResponseParseResult::Anchored(receipt)) => {
+                    return Ok(*receipt);
+                }
+                Ok(CasResponseParseResult::Unauthorized) => {
+                    return Err(anyhow!("remote CAS request unauthorized"));
+                }
+                Err(e) => {
```
stbrody, line 102: I feel like we might want to treat the request still being PENDING as different than other types of errors.  Like if it's still pending that's expected behavior, so we shouldn't log anything, except possibly at a verbose log level for informational purposes.  But if the request is FAILED or in some other unexpected state, we should probably log an error.

anchor-tx/src/cas_remote.rs
```diff
@@ -0,0 +1,325 @@
...
+    };
+    let witness_car_bytes = b64_standard.decode(witness_car_b64)?;
+    let car_reader = CarReader::new(witness_car_bytes.as_ref()).await?;
+    let mut remote_merkle_nodes = MerkleNodes::default();
+    let mut detached_time_event: Option<DetachedTimeEvent> = None;
+    let mut proof: Option<Proof> = None;
+    for (cid, block) in car_reader
+        .stream()
+        .into_stream()
+        .try_collect::<Vec<(_, _)>>()
```
stbrody, line 209: does this iterate the whole CAR file up front and turn it into a vector of blocks?  Can we write this in a way that we handle the CAR blocks in a streaming, iterative way, so that we don't have to make a fully copy of the CAR file in memory?

anchor-tx/src/cas_remote.rs
```diff
@@ -0,0 +1,325 @@
...
+}
+
+enum CasResponseParseResult {
+    Anchored(Box<Receipt>),
+    Unauthorized,
+}
+
+#[async_trait]
+impl TransactionManager for RemoteCas {
+    async fn make_proof(&self, root: Cid) -> Result<Receipt> {
```
stbrody, line 91: nitpick: I don't love the name "make_proof" for this function.  To me that makes it sound like it's just constructing a proof in memory, rather than that it's actually making a remote call to get the actual work of anchoring to happen.

maybe `anchor_root`?

stbrody (commented):
a few small suggestions, but generally looking good.  None of my comments are blockers, I'm happy whenever David is happy.

core/src/lib.rs
smrz2001: https://linear.app/3boxlabs/issue/AES-342/implement-nodeid-to-encapsulate-node-private-key-utilities

anchor-tx/src/cas_remote.rs
smrz2001: This matches what the CAS responds with, which is a base64-encoded string for the witness CAR file.

smrz2001: We added a comment clarifying this:
```
struct AnchorResponse {
    pub message: String,
    /// This field contains the Base64 Standard encoding of the witness CAR file
    pub witness_car: Option<String>,
}
```

anchor-remote/src/cas_remote.rs
```diff
@@ -0,0 +1,330 @@
...
+    }
+}
+
+async fn parse_anchor_response(anchor_response: String) -> Result<CasResponseParseResult> {
+    // Return if we were unable to parse the anchor response
+    let anchor_response = serde_json::from_str::<CasAnchorResponse>(anchor_response.as_str())?;
+
+    // If the response does not contain a witness CAR file, the anchor request is either still pending or it failed
+    // because the request was unauthorized.
+    let Some(witness_car_b64) = anchor_response.witness_car else {
```
dav1do, line 199: fancy.. I had to look up how long `let/else` has been stable (since 1.65 and I've never used it! 🙊)

dav1do (approved):
:shipit: It looks like Spencer's comments have been addressed in the new version and we can come back to the public key/NodeId questions I had in a follow up.

PR #480, opened 2024-08-15
chore: Flatten EventInsertable type

stbrody:
Removes the EventInsertableBody type and pushes its fields up into the parent EventInsertable

PR #481, opened 2024-08-15
chore: Remove Clone from EventInsertable

stbrody:
Builds on https://github.com/ceramicnetwork/rust-ceramic/pull/480

### DIFF
--- a/service/src/event/order_events.rs
+++ b/service/src/event/order_events.rs
@@ -47,25 +47,27 @@ impl OrderEvents {
                 (e.cid(), e.deliverable())
             }));
         let mut deliverable = Vec::with_capacity(candidate_events.len());
-        candidate_events.retain(|(e, h)| {
+        let mut remaining_candidates = Vec::with_capacity(candidate_events.len());
+
+        for (e, h) in candidate_events {
             if e.deliverable() {
-                deliverable.push((e.clone(), h.clone()));
-                false
+                deliverable.push((e, h))
             } else {
-                true
+                remaining_candidates.push((e, h))
             }
-        });
-        if candidate_events.is_empty() {
+        }
+
+        if remaining_candidates.is_empty() {
             return Ok(OrderEvents {
                 deliverable,
-                missing_history: Vec::new(),
+                missing_history: remaining_candidates,
             });
         }
 
-        let mut undelivered_prevs_in_memory = VecDeque::with_capacity(candidate_events.len());
-        let mut missing_history = Vec::with_capacity(candidate_events.len());
+        let mut undelivered_prevs_in_memory = VecDeque::with_capacity(remaining_candidates.len());
+        let mut missing_history = Vec::with_capacity(remaining_candidates.len());
 
-        while let Some((mut event, header)) = candidate_events.pop() {
+        while let Some((mut event, header)) = remaining_candidates.pop() {
             match header.prev() {
                 None => {
                     unreachable!("Init events should have been filtered out since they're always deliverable");
@@ -181,18 +183,30 @@ mod test {
         (after_1, after_2)
     }
 
+    /// Takes the given events from Recon and turns them into two vectors of insertable events.
     async fn get_insertable_events(
         events: &[ReconItem<EventId>],
-    ) -> Vec<(EventInsertable, EventMetadata)> {
-        let mut insertable = Vec::with_capacity(events.len());
+        first_vec_count: usize,
+    ) -> (
+        Vec<(EventInsertable, EventMetadata)>,
+        Vec<(EventInsertable, EventMetadata)>,
+    ) {
+        let mut insertable = Vec::with_capacity(first_vec_count);
+        let mut remaining = Vec::with_capacity(events.len() - first_vec_count);
+        let mut i = 0;
         for event in events {
             let new = CeramicEventService::parse_discovered_event(event)
                 .await
                 .unwrap();
-            insertable.push(new);
+            if i < first_vec_count {
+                insertable.push(new);
+            } else {
+                remaining.push(new)
+            }
+            i += 1
         }
 
-        insertable
+        (insertable, remaining)
     }
 
     #[test(tokio::test)]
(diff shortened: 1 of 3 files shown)


service/src/event/order_events.rs
```diff
@@ -47,25 +47,31 @@ impl OrderEvents {
...
+                deliverable.push((e, h))
             } else {
-                true
+                remaining_candidates.push((e, h))
             }
-        });
-        if candidate_events.is_empty() {
+        }
+        // Need to reverse the vectors because we popped events off of candidate_events from back to front
+        deliverable.reverse();
```
stbrody, line 61: this feels a bit hacky and probably not the most efficient.  The other idea I had was to replace the `pop().unwrap()` with `remove(0)`, but I was worried that removing from the front of the array repeatedly would be even less efficient than reversing at the end.

It really feels like there should be a better way to move elements out of one vector into two others in an efficient manor...

service/src/event/order_events.rs
```diff
@@ -47,25 +47,31 @@ impl OrderEvents {
                 (e.cid(), e.deliverable())
             }));
         let mut deliverable = Vec::with_capacity(candidate_events.len());
-        candidate_events.retain(|(e, h)| {
+        let mut remaining_candidates = Vec::with_capacity(candidate_events.len());
+
+        while !candidate_events.is_empty() {
+            let (e, h) = candidate_events.pop().unwrap();
```
stbrody, line 53: the unwrap() is safe because of the is_empty check in the while loop condition

store/src/sql/entities/event.rs
```diff
@@ -39,7 +39,7 @@ pub async fn rebuild_car(blocks: Vec<BlockRow>) -> Result<Option<Vec<u8>>> {
     Ok(Some(car))
 }
 
-#[derive(Debug, Clone)]
```
stbrody, line 42: this is the core change.  This is to prep for the next PR, which will replace the `blocks` field of this struct with the parsed `unvalidated::Event<Ipld>` instead, which doesn't support Cloning.

service/src/event/order_events.rs
stbrody: Adding @nathanielc to see if you have any ideas for how to do this more efficiently

service/src/event/order_events.rs
dbcfd: How about `while let Some((e,h)) = candidate_events.pop() {`

service/src/event/order_events.rs
dbcfd: VecDeque and push front.

stbrody: What does this mean?  I have a vector, I want to move the elements from that vector out into two other vectors, deciding which vector to move into based on a property of the values in the input vector.  How can a VecDeque help with that?

dbcfd: Instead of using a vector to store events, use a VecDeque. You push front instead of back. You then return an object with a VecDeque.

stbrody: Hmm but at the end I need to be returning a Vec, so I'd need to move everything out of the VecDeque into a Vec at the end anyway, is that actually going to be any more efficient?

service/src/event/ordering_task.rs
```diff
@@ -628,12 +628,11 @@ mod test {
     }
 
     async fn insert_10_with_9_undelivered(pool: &SqlitePool) {
-        let insertable = get_n_insertable_events(10).await;
-        let mut init = insertable.first().unwrap().to_owned();
+        let mut insertable = get_n_insertable_events(10).await;
+        let mut init = insertable.remove(0);
```
nathanielc, line 632: this change is O(n) when previously it was constant time. Not sure it matters but thought I would point it out.

service/src/event/order_events.rs
nathanielc: Make sure I understand. We have three vectors:

* candidate_events
* deliverable
* remaining_candidates

We want to _move_ the values from candidate_events into one of the other two vectors and we want to do it in order? When we are done candidate_events is empty?

In this case a simple for loop should suffice.

```rust
for event in candidate_events {
   if ... {deliverable.push(event)} else {remaining_candidates.push(event) }
}
```
Note since we are not using a reference to candidate_events in the for loop the values are moved out of the vector into the loop.

Aside there is a one liner for this in nightly rust https://doc.rust-lang.org/std/vec/struct.Vec.html#method.extract_if

service/src/event/order_events.rs
```diff
@@ -181,18 +187,30 @@ mod test {
         (after_1, after_2)
     }
 
+    /// Takes the given events from Recon and turns them into two vectors of insertable events.
```
nathanielc, line 190: How are the two vectors different? Should we make a small struct to give them names?

dav1do (approved):
I'm good with this as is, I noticed two things that could be nicer (both originally my mistakes) so you can take them or leave them.

store/src/sql/entities/event.rs
stbrody: The `unvalidated::Event` type can't be Clone right now because it uses a Box pointer internally.  We could change that to an Rc or do something else to make it cloneable.  But I kind of feel like having Events not be cloneable will force us to be smarter about moving things and not making extraneous copies.  Events could be large and cloning them a bunch seems pretty inefficient unless we really have to.

service/src/event/order_events.rs
stbrody: This looks like a nice cleanup, but I'd like to do that in a separate PR to keep this one more focused.

stbrody: Made a ticket to revisit this: https://linear.app/3boxlabs/issue/AES-309/cleanup-ordereventstry-new

store/src/sql/entities/event.rs
dav1do: Yeah, I agree with that. It can be convenient in tests but using an arc/box/rc is fine. I was also thinking about the "common" things to define on public types e.g. PartialEq, Eq, Hash. IPLD derives Eq because it doesn't really support floats which is mostly where that comes in. We don't have to add them now (or ever). A while back I wanted them on a couple tests and ended up just encoding things to bytes to compare, which is fine.

PR #482, opened 2024-08-15
feat: add noop ceramic-olap binary

nathanielc:
The ceramic-olap binary copies verbatim some setup code from ceramic-one re tracing/logging/metrics. Eventually this may move back into the ceramic-one binary or we can find a way to make it generic and reuse. Given this is only the first copy of the code I don't think we should work to make it generic yet. Not until we either decide this is the long term home for the ceramic-olap binary or have a third binary we want to have a similar setup.

This binary does nothing yet except start tracing/logging/metrics and then wait for a signal to shutdown gracefully.

### DIFF
--- a/olap/src/lib.rs
+++ b/olap/src/lib.rs
@@ -1,0 +1,238 @@
+mod aggregator;
+mod metrics;
+
+use anyhow::{anyhow, Result};
+use ceramic_metrics::config::Config as MetricsConfig;
+use clap::{Args, Parser, Subcommand, ValueEnum};
+use futures::StreamExt as _;
+use multihash::Multihash;
+use multihash_codetable::Code;
+use multihash_derive::Hasher as _;
+use signal_hook::consts::signal::*;
+use signal_hook_tokio::Signals;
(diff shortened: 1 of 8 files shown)


AaronGoldman:
A lot of metrics for a no-op but looks good.

olap/src/lib.rs
```diff
@@ -0,0 +1,238 @@
...
+    Daemon(Box<DaemonOpts>),
+}
+
+#[derive(Args, Debug)]
+struct DaemonOpts {
+    /// Bind address of the metrics endpoint.
+    #[arg(
+        short,
+        long,
+        default_value = "127.0.0.1:9464",
```
dav1do, line 35: I know this is standard, but wondering if we want a different default if we expect c1 and olap to run on one machine?

dav1do (approved):

olap/src/lib.rs
nathanielc: Good point, its easy to change later and there might be a few things like that to change.

PR #483, opened 2024-08-15
chore: Change EventInsertable to store the parsed Event object instead of its raw blocks

stbrody:
builds on https://github.com/ceramicnetwork/rust-ceramic/pull/481

### DIFF
--- a/store/src/sql/entities/event.rs
+++ b/store/src/sql/entities/event.rs
@@ -43,46 +43,7 @@ pub async fn rebuild_car(blocks: Vec<BlockRow>) -> Result<Option<Vec<u8>>> {
 /// The type we use to insert events into the database
 pub struct EventInsertable {
     /// The event order key (e.g. EventID)
-    pub order_key: EventId,
-    /// The data that makes up the event
-    pub body: EventInsertableBody,
-}
-
-impl EventInsertable {
-    /// Try to build the EventInsertable struct from a carfile.
-    pub async fn try_from_carfile(order_key: EventId, body: &[u8]) -> Result<Self> {
-        let cid = order_key.cid().ok_or_else(|| {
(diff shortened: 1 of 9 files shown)


store/src/sql/entities/event.rs
```diff
@@ -92,20 +92,11 @@ impl EventInsertable {
...
-
-        if car_bytes.is_empty() {
-            return Err(Error::new_app(anyhow!(
-                "CAR file is empty: cid={}",
-                event_cid
-            )))?;
-        }
+    /// Underlying bytes that make up the event
+    pub async fn get_raw_blocks(&self) -> Result<Vec<EventBlockRaw>> {
+        let car = self.event.encode_car().await.map_err(Error::new_app)?;
```
stbrody, line 97: I hate this.  Encoding it to a CAR just to re-encode it to the raw blocks is doing two passes over the data for no reason.  I tried to add a method to unvalidated::Event that worked just like `encode_car()` but would instead return a `Vec<EventBlockRaw>` so we could get what we need in a single pass.  The problem is that the `EventBlockRaw` type lives in the `ceramic-store` crate while the `unvalidated::Event` type lives in the `ceramic-event` crate, but the store crate depends on the event crate, so I can't make the event crate depend on the store crate without introducing a circular dependency.

Would love any suggestions on how to clean this up.

stbrody:
This is passing all tests *except* the migration tests, I might need a hand to figure out what's up with the migration tests.

stbrody:
Okay I figured out a part of what's wrong with the migration tests - there are some event CAR files being used in that test that don't round-trip through `unvalidated::Event`.  Pushed a change that demonstrates that.  Now the question is: is this a bug in our serialization/deserialization logic?  Or is the migration test using invalid CAR files?

@nathanielc

PR #484, opened 2024-08-15
feat: add event informant and stream cid (part 4)

smrz2001:
This PR includes the changes for storing the source node DID and stream Init Event CID in the Event Store.

This is needed to be able to filter out events submitted to the C1 node via API so that only our events are anchored.


event-svc/src/event/store.rs
```diff
@@ -44,9 +44,10 @@ impl recon::Store for EventService {
     async fn insert_many(
         &self,
         items: &[ReconItem<Self::Key>],
+        source: String,
     ) -> ReconResult<recon::InsertResult<EventId>> {
         let res = self
-            .insert_events(items, DeliverableRequirement::Asap)
+            .insert_events(items, DeliverableRequirement::Asap(source))
```
AaronGoldman, line 50: Should DeliverableRequirement and source just be two different arguments?

event-svc/src/store/sql/query.rs
```diff
@@ -177,12 +177,12 @@ impl ReconQuery {
             order_key, cid,
             ahash_0, ahash_1, ahash_2, ahash_3,
             ahash_4, ahash_5, ahash_6, ahash_7,
-            delivered
+            delivered, init_cid, source, anchored
```
AaronGoldman, line 180: This is the point of the PR.

event-svc/src/event/service.rs
```diff
@@ -145,15 +149,18 @@ impl EventService {
     /// In the future, we will need to do more event validation (verify all EventID pieces, hashes, signatures, etc).
     pub(crate) async fn parse_discovered_event(
         item: &ReconItem<EventId>,
+        source: Option<String>,
```
stbrody, line 153: I'm wondering if it might make sense to introduce a new type, even if it's just an alias of String, something like `DIDString`, just so it's clearer when reading the code what is being passed around here and various other places that take a DID string?  WDYT @dav1do?

stbrody, line 153: why is source optional?  Aren't all events going to have a source?

event-svc/src/event/service.rs
```diff
@@ -227,7 +235,13 @@ impl EventService {
         items: &[ReconItem<EventId>],
         source: DeliverableRequirement,
     ) -> Result<InsertResult> {
-        let (to_insert, mut invalid) = Self::validate_events(items).await?;
+        let source_node_did = match &source {
+            DeliverableRequirement::Immediate | DeliverableRequirement::Lazy => &self.node_did,
```
stbrody, line 240: this feels weird and maybe like a layering violation?  Why would the requirements around IOD affect how you find the source of an event?  Are we combining behaviors into concepts that should be separate?

event-svc/src/event/service.rs
```diff
@@ -52,15 +54,15 @@ pub trait BlockStore {
...
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum DeliverableRequirement {
     /// Must be ordered immediately and is rejected if not currently deliverable. The appropriate setting
     /// for API writes as we cannot create an event without its history.
     Immediate,
     /// This will be ordered as soon as its dependencies are discovered. Can be written in the meantime
     /// and will consume memory tracking the event until it can be ordered. The appropriate setting for recon
     /// discovered events.
-    Asap,
+    Asap(String),
```
stbrody, line 66: I'm guessing the String here is the source DID string?  It feels pretty weird that Asap would include a source string but Lazy and Immediate don't.  I feel like we're conflating concepts here.

event-svc/src/event/service.rs
stbrody: Maybe we should just add an explicit source arg to this function?  And rename the current "source" arg to "delivery_reqs" or something like that.

event-svc/src/store/sql/entities/event.rs
```diff
@@ -45,20 +45,26 @@ pub async fn rebuild_car(blocks: Vec<BlockRow>) -> Result<Option<Vec<u8>>> {
...
 }
 
 impl EventInsertable {
     /// EventInsertable constructor
     pub fn new(
         order_key: EventId,
+        stream_cid: Cid,
         event_cid: Cid,
         event: unvalidated::Event<Ipld>,
+        source: Option<String>,
```
stbrody, line 67: I see we add stream_cid, source, *and* anchored to the sql table, but we're only adding the first two here.  Does this need to take `anchored` also?

interest-svc/src/interest/store.rs
```diff
@@ -20,6 +20,8 @@ impl recon::Store for InterestService {
     async fn insert_many(
         &self,
         items: &[ReconItem<Self::Key>],
+        // the recon::Store trait is shared between InterestService and EventService but only events track a source.
```
stbrody, line 23: don't interests have a source too?  Could this be unified?

migrations/sqlite/20240809205107_event_source_anchored.up.sql
```diff
@@ -0,0 +1,12 @@
+-- Add up migration script here
+
+-- The CID of the Init Event for the stream
+ALTER TABLE ceramic_one_event ADD COLUMN init_cid BLOB DEFAULT NULL;
```
stbrody, line 4: how will this get backfilled for events that were created before this field was present in the db?

Really ideally this column would be non-nullable and always required, but that makes the upgrade path trickier

one/src/migrations.rs
```diff
@@ -93,8 +93,10 @@ async fn from_ipfs(opts: FromIpfsOpts) -> Result<()> {
     let network = opts.network.to_network(&opts.local_network_id)?;
     let db_opts: DBOpts = (&opts).into();
     let sqlite_pool = db_opts.get_sqlite_pool().await?;
+    // the source did:key is none for migrations as we don't know the source node.
```
stbrody, line 96: hmm, couldn't an argument be made to just consider this local node the source?  This is the first node bringing this data into the Recon network after all...

event-svc/src/event/service.rs
AaronGoldman: or a `[u8; 32]` NodeID class that can to and from `Cid`, `PeerId`, and `DidKey`

```rs
pub struct NodeId {
    public_ed25519_key_bytes: [u8; 32],
    singing_ed25519_keypair: Option<Ed25519KeyPair>,
}

impl NodeId {
    pub fn cid(&self) -> Cid {todo!()}
    pub fn did_key(&self) -> String {todo!()}
    pub fn ed25519_key_pair(&self) -> Ed25519KeyPair {todo!()}
    pub fn peer_id(&self) -> PeerId {todo!()}
    pub fn try_from_cid(cid: Cid)  -> Result<Self> {todo!()}
    pub fn try_from_did_key(did_key: &str) -> Result<Self> {todo!()}
    pub fn try_from_dir(key_dir: PathBuf) -> Result<Self> {todo!()}
    pub fn try_from_secret(secret: &str) -> Result<Self> {todo!()}
    pub fn try_from(peer_id: PeerId) -> Result<Self> {todo!()}
}
```

AaronGoldman: not when we insert from a migration. 
API: self
Generated time event: self
Recon: the `NodeId.try_from_peer_id(peer_id)`
migration: None

one/src/migrations.rs
AaronGoldman: Maybe, maybe not. It was likely created and anchored by someone else before the migration. e.g. the old node you are migrating from.

stbrody: right, but that's almost certainly "logically" the same node.  Ie operated by the same person, serving the same purpose/application, and directly replacing the old node in every way.  What's the harm in considering the migrated event as having this node as the source?

event-svc/src/event/service.rs
AaronGoldman: let use `informant`

recon/src/recon/tests.rs
```diff
@@ -524,10 +524,13 @@ async fn word_lists() {
...
-                r.insert(vec![ReconItem::new(
-                    key.as_bytes().into(),
-                    key.to_uppercase().as_bytes().to_vec(),
-                )])
+                r.insert(
+                    vec![ReconItem::new(
+                        key.as_bytes().into(),
+                        key.to_uppercase().as_bytes().to_vec(),
+                    )],
+                    "".to_owned(),
```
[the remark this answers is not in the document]
smrz2001: Updated to make it an `Option` where it could be unspecified, and passed in a value otherwise.

one/src/migrations.rs
smrz2001: If these events are considered originating locally, they might be considered eligible for self-anchoring.

During migration, it would probably be safe to assume that any events that needed to be anchored were already anchored.

event-svc/src/store/sql/entities/event.rs
smrz2001: We just took out `anchored`. We were planning to use it in part 5 but are not doing so anymore.

event-svc/src/event/service.rs
smrz2001: Yes, we're now passing the source (`informant`) in separately.

migrations/sqlite/20240809205107_event_source_anchored.up.sql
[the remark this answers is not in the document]
smrz2001: We're currently using the `stream_cid` column just while creating the time event.

We're also not doing a table join anymore and have removed `anchored`.

We'll store and use the event service's resume token and will treat it opaquely.

migrations/sqlite/20240809205107_event_source_anchored.up.sql
```diff
@@ -0,0 +1,12 @@
...
+
+-- The DID of the node that sent the event or our own DID if the event came in via API
+ALTER TABLE ceramic_one_event ADD COLUMN source TEXT DEFAULT NULL;
```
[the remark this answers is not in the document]
smrz2001: Aaron and I discussed that this does make the table scan at the beginning of a Recon sync, but it does feel worth having the stream ID in the table and not have to parse a CAR file to get it.

To optimize the scan, we could possibly take the AHash and EventId out into their own table that gets scanned at the start of a sync.

migrations/sqlite/20240809205107_event_source_anchored.up.sql
smrz2001: Right now we're only using this field while generating Time Events, which will only happen for events that also have the `informant` populated. We should never have events with an `informant` and without a `init_cid`. This means that the `init_cid` being missing is ok from the self-anchoring perspective, but it does still need to be populated so that we can support things like looking up events for a stream. 

Created a [ticket](https://linear.app/3boxlabs/project/self-anchoring-pt-1-157039361486/issues) for this.

migrations/sqlite/20240809205107_event_source_anchored.up.sql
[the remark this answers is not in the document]
dav1do: yeah, I kind of feel like we should have "recon_events" which is just an event_id (or a few of them or some columns for flat recon) and the hash.. and the "events" from ceramic's perspective and all the deliverable, stream info should be its own table.. but I'm not sure if it's worth making that change and instead we just make a wide events table for now and then migrate to parquet or whatever optimizations are next

event-svc/src/event/service.rs
AaronGoldman: they are NodeId now

event-svc/src/event/service.rs
AaronGoldman: `insert_events` now just passes informant through to `validate_events`
`validate_events` now just passes informant through to `parse_discovered_event`
`parse_discovered_event`  now just passes informant through to `EventInsertable::new`
`EventInsertable::new` now just passes informant through to `EventInsertable{informant}`

There is just a lot of plumbing to get the `informant` from the [api, recon, migration, time event creator] to the sql bind call. 

We pulled this apart but there might still be a natral struct we are missing that is what we should be passing around for a event insert. Or maby we just have too many layers between [api, recon, migration, time event creator] and the storage.

migrations/sqlite/20240809205107_event_source_anchored.up.sql
```diff
@@ -0,0 +1,9 @@
+-- Add up migration script here
+
+-- The CID of the Init Event for the stream
+ALTER TABLE ceramic_one_event ADD COLUMN init_cid BLOB DEFAULT NULL;
+
+-- The ID of the node that sent the event or our own ID if the event came in via API
+ALTER TABLE ceramic_one_event ADD COLUMN informant TEXT DEFAULT NULL;
```
dav1do, line 7: Should we create indexes on these two columns? At least on informant if we're querying it now, and we'll likely need one on init_cid if we ever want to actually retrieve events by stream.

event-svc/src/store/metrics.rs
```diff
@@ -124,11 +124,12 @@ where
     async fn insert_many(
         &self,
         items: Vec<ceramic_api::ApiItem>,
+        informant: &NodeId,
```
dav1do, line 127: should we just take an owned version since this is a copy type?

event-svc/src/event/store.rs
```diff
@@ -44,9 +44,14 @@ impl recon::Store for EventService {
...
         &self,
         items: &[ReconItem<Self::Key>],
+        informant: &NodeId,
     ) -> ReconResult<recon::InsertResult<EventId>> {
         let res = self
-            .insert_events(items, DeliverableRequirement::Asap)
+            .insert_events(
+                items,
+                DeliverableRequirement::Asap,
+                Some(&informant.clone()),
```
dav1do, line 53: probably don't need clone since this is copy (related to other comment: should we pass owned NodeId as argument?)

event-svc/src/event/service.rs
```diff
@@ -175,11 +177,12 @@ impl EventService {
 
     pub(crate) async fn validate_events(
         items: &[ReconItem<EventId>],
+        informant: Option<&NodeId>,
```
dav1do, line 180: I don't think we want/need the reference on this

core/src/node_id.rs
```diff
@@ -0,0 +1,232 @@
...
+use crate::{StreamId, StreamIdType};
+use anyhow::{anyhow, Context, Ok, Result};
+use cid::multihash::Multihash;
+use cid::Cid;
+use libp2p_identity::PeerId;
+use ring::signature::{Ed25519KeyPair, KeyPair};
+
+/// NodeId is the public_ed25519_key_bytes of the node
+#[derive(Clone, Eq, PartialEq, Default, Copy)]
+pub struct NodeId {
```
dav1do, line 13: I find default a bit weird on this but it does seem useful in the tests. 

Below I suggested we should stop using references and just pass an owned version of this because it's Copy.. but now I'm wondering if this should be copy. We call `did_key()` for every insert now, which means doing the multibase encoding and allocating a string.. should we track more than 32 bytes of public key on this? Does it make sense to keep the a DidDocument (or at least a did:key string) on self? 

I like that we have a NodeId, but it seems a bit odd to have _another_ new construct when we already have DIDs and PeerIds and it'd be nice to consolidate.. I don't feel super strongly but I'm not sure when I should pick which one as they're all just variations of public keys.

dav1do (commented):
A few more questions but overall it's looking good. Happy to get on a call to talk through them if it helps speed it up. And the PR name is out of date now - not sure how to comment on that directly :).

event-svc/src/store/metrics.rs
dav1do: actually, ignore this and see comment on `NodeId`

event/src/unvalidated/event.rs
```diff
@@ -239,6 +239,18 @@ where
             )),
         }
     }
+
+    /// Get the CID of the init event of the stream for this event
+    pub fn stream_cid(&self) -> &Cid {
```
dav1do, line 244: maybe we should delete this? it's (now) a duplicate of the `id()` function that was recently modified

migrations/sqlite/20240809205107_event_source_anchored.up.sql
dav1do: sounds like we're okay skipping this for now. not sure when we'll query the init_cid and for now, we're willing to seek the highwater mark and then scan from there to filter out by informant

dav1do (approved):

event-svc/src/store/sql/entities/event.rs
stbrody: in that case do we still need it in the sql table?

dav1do: I think [it was](https://github.com/ceramicnetwork/rust-ceramic/pull/484/files#diff-98a5d0fd33ad9eeaeafa6bf8b5092b30f73a3d5d2b63f752fba599fbaa90c8d6) removed, unless "it" is not `anchored` 😅.. `init_cid` and `informant` were added

migrations/sqlite/20240809205107_event_source_anchored.up.sql
```diff
@@ -0,0 +1,9 @@
+-- Add up migration script here
```
stbrody, line 1: looks like the file names for these migrations probably should be updated in a follow-up since they don't really reflect what is happening in them anymore

stbrody: e.g. they don't add "anchored" anymore

smrz2001: Yes, we fixed this in part 5.

dav1do: Oh dang, I didn't notice that. Unfortunately, I don't think you can change a migration name after it's been applied somewhere.. pretty sure it's a unique constraint or part of the hash with how sqlx manages them. We can try it but iirc it will error saying something like migration with name x has previously been applied but is missing.

stbrody: presumably so long as we fix this before it goes out in a release it's fine?

dav1do: well, it's in a prerelease now and if anyone has run main locally they'll need to manually adjust their db (or delete the file). plus any test nodes we're running it on... my opinion is that's it a bummer but isn't worth the risk/hassle. I don't really pay attention to migration names though (clearly) and just look at the schemas.

PR #485, opened 2024-08-16
refactor: create sync car crate

nathanielc:
Replaces iroh-car with ceramic-car that has both async and sync APIs. Additionally changes the Event::encode_car and Event::decode_car methods to be synchronous as we nearly always have the bytes in memory already and its cheap to copy them into memory when we do not.


dav1do (approved):
making this sync is nice :shipit:

PR #486, opened 2024-08-16
fix: Fix bug in TimeEvent witness blocks parsing logic

stbrody:


event/src/unvalidated/event.rs
```diff
@@ -65,8 +64,7 @@ where
             let block_bytes = car_blocks
                 .get(cid)
                 .ok_or_else(|| anyhow!("Time Event CAR data missing block for path index"))?;
-            blocks_in_path.push(block);
```
stbrody, line 68: this was the bug.  We pushed the block before we updated it, meaning we pushed the first block twice and dropped the last block

event/src/unvalidated/event.rs
```diff
@@ -50,12 +50,13 @@ where
...
-        let mut block: Ipld = serde_ipld_dagcbor::from_slice(block_bytes)?;
-        blocks_in_path.push(block.clone());
+        blocks_in_path.push(serde_ipld_dagcbor::from_slice(block_bytes)?);
         let parts: Vec<_> = event.path().split('/').collect();
         // Add blocks for all parts but the last as it is the prev.
         for index in parts.iter().take(parts.len() - 1) {
-            let cid = block
+            let cid = blocks_in_path
+                .last()
+                .unwrap()
```
dav1do, line 59: I'd prefer avoid an unwrap unless we're confident it can never happen (and often worth a comment about the guarantees imo). Here's an alternative with `and_then` but there might be a nicer way... 

```rust
let cid = blocks_in_path
    .last()
    .and_then(|i| i.get(*index).transpose())
    .ok_or_else(|| anyhow!("Time Event path indexes missing data"))??;
```

samika98: we can add a check which checks if blocks_in_path is empty

event/src/unvalidated/event.rs
samika98: 😨

event/src/unvalidated/event.rs
stbrody: The unwrap can't fail, added a comment why it's safe

dav1do: actually, this might be fine since we push a block above so it can't be empty.. i just missed it

dav1do: see the comment, thanks

dav1do (approved):

PR #487, opened 2024-08-16
chore: Change EventInsertable to store the parsed Event object instead of its raw blocks

stbrody:


event/src/unvalidated/event.rs
```diff
@@ -9,7 +9,46 @@ use std::{collections::HashMap, fmt::Debug};
 use tokio::io::AsyncRead;
 use tracing::debug;
 
-use super::{cid_from_dag_cbor, init, signed};
+use super::{cid_from_dag_cbor, init, signed, Payload};
+
+/// Helper function for Event::decode_car for gathering all the Ipld blocks used by a time event
```
stbrody, line 14: no functional change here, just moving it out the impl for Event

store/src/sql/entities/event.rs
```diff
@@ -48,22 +49,39 @@ pub struct EventInsertable {
...
+    event: unvalidated::Event<Ipld>,
 }
 
 impl EventInsertable {
     /// EventInsertable constructor
-    pub fn new(order_key: EventId, blocks: Vec<EventBlockRaw>, deliverable: bool) -> Self {
-        let cid = order_key.cid().unwrap();
+    pub fn new(
+        order_key: EventId,
+        event_cid: Cid,
```
dav1do, line 60: It'd be nice to delete the CID from this struct and just get it from event in the future.

event/src/unvalidated/event.rs
dav1do: should we move it to be an impl on `RawTimeEvent`? could be `pub(crate)` if we don't want to make it pub yet

dav1do (approved):
Nice :shipit: a couple of thoughts but nothing blocking this PR.

store/src/sql/entities/event.rs
stbrody: yeah, I was lamenting the fact that there's no easy way to get the event CID from the Event struct itself while writing this.

event/src/unvalidated/event.rs
stbrody: given that it would still need to take the proof and car_blocks as arguments, I actually feel like it makes more sense to leave it as a floating helper function since it really is pretty specific to the context it's being used in in decode_car, at least for now.

PR #488, opened 2024-08-17
feat: self-anchoring (part 5)

smrz2001:
This PR pulls together most of the changes needed for self-anchoring except for the CLI arguments.

It builds on top of the existing PRs ([here](https://github.com/ceramicnetwork/rust-ceramic/pull/478), [here](https://github.com/ceramicnetwork/rust-ceramic/pull/476), [here](https://github.com/ceramicnetwork/rust-ceramic/pull/484), and [here](https://github.com/ceramicnetwork/rust-ceramic/pull/477)) so there are some redundant changes here. They will get cleaned up as those PRs merge.

We'll leave this PR in a draft state until the other PRs are merged.


service/src/event/service.rs
```diff
@@ -288,6 +289,100 @@ impl CeramicEventService {
...
+        )));
+        EventInsertable::new(
+            time_event_order_key,
+            time_event.id(),
+            time_event_cid,
+            event,
+            self.node_did.clone(),
+            true,
+        )
+        .context("Error crating the EventInsertable")
```
dav1do, line 384: nit: `crating` -> `creating`

store/src/sql/query.rs
```diff
@@ -151,6 +151,25 @@ impl EventQuery {
...
+
+    /// Inserts a new high water mark after a batch has been anchored
+    /// Requires 1 parameter1:
+    ///     $1 = new_high_water_mark (i64)
+    pub fn insert_anchoring_high_water_mark() -> &'static str {
+        r#"INSERT INTO ceramic_one_anchor_high_water_mark (high_water_mark) VALUES ($1);"#
+    }
+
+    /// Fetch event CIDs from a specified source that are above the current anchoring high water mark
+    pub fn unanchored_from_high_water_mark() -> &'static str {
```
dav1do, line 163: how do we filter out time events? will source be different? it seems like it might be weird in the interim if js-ceramic is giving us time events and we're putting source == our address and then we anchor them again

service/src/event/service.rs
```diff
@@ -307,6 +402,71 @@ pub enum InvalidItem {
...
+            .filter_map(|r| match r {
+                Ok(e) => Some(e),
+                Err(e) => {
+                    error!("error processing time event: {}", e);
+                    None
+                }
+            })
+            .collect::<Vec<_>>();
+        // TODO: We need to handle high water mark updates when there are errors here
+        CeramicOneEvent::insert_many(&self.pool, events.iter()).await?;
```
dav1do, line 454: I'm concerned about this transaction.. is there any way we can safely break this up or are we just going to take the lock for as long as it takes? We should try to benchmark this because it's what we normally expect per hours happening in one tx.

It seems like we could iter in batches and update the highwater mark after each batch? not sure if somehow erroring in the middle and having half the batch split causes a different problem though

service/src/event/service.rs
```diff
@@ -307,6 +402,71 @@ pub enum InvalidItem {
...
+    async fn local_sourced_data_events(&self) -> anyhow::Result<Vec<AnchorRequest>> {
+        // We must have a node DID to use here. Return an error if we don't.
+        let node_did = self
+            .node_did
+            .clone()
+            .ok_or_else(|| Error::new_app(anyhow!("node DID required to get anchor requests")))?;
+
+        // Fetch event CIDs from the events table using the previous high water mark
+        Ok(
+            CeramicOneEvent::unanchored_events_from_high_water_mark(
```
dav1do, line 416: I'm a bit worried that reading 1M is going to be slow.. but not nearly as worried as I am about writing them back.

service/src/event/service.rs
smrz2001: Yeah, that's been a concern of mine from the start, both for reads and writes.

We should write a batch then update the HWM, and so on, until were through the list. If any batch fails, the HWM doesn't move up. We can error in the middle and redo stuff as long as we don't update the HWM unless the batch was written successfully.

anchor-service/Cargo.toml
```diff
@@ -12,13 +12,19 @@ anyhow.workspace = true
...
 cid.workspace = true
 expect-test.workspace = true
+futures.workspace = true
 indexmap.workspace = true
+ipld-core.workspace = true
 multihash-codetable.workspace = true
+multibase.workspace = true
 serde.workspace = true
-serde_ipld_dagjson.workspace = true
+sqlx.workspace = true
```
dav1do, line 24: do we need this dependency? Ideally we should export everything we need from the `ceramic-sql` crate.

anchor-service/src/anchor.rs
```diff
@@ -76,15 +100,123 @@ pub struct TimeEventBatch {
...
             .field("proof", &self.proof)
-            .field("time_events", &self.time_events)
+            .field("raw_time_events", &self.raw_time_events)
             .finish()
     }
 }
+
+impl TimeEventBatch {
+    /// Try to convert the TimeEventBatch into a list of TimeEventInsertables
+    pub fn try_to_insertables(&self) -> Result<Vec<TimeEventInsertable>> {
```
dav1do, line 118: maybe this can return a stream rather than collect the entire set into a vector in memory? I still need to look where/how it's used but just off the top of my head

anchor-service/src/anchor.rs
```diff
@@ -76,15 +100,123 @@ pub struct TimeEventBatch {
...
+        Ok(events)
+    }
+
+    /// Build a TimeEventInsertable from a RawTimeEvent and AnchorRequest
+    fn build_time_event_insertable(
+        &self,
+        time_event: &RawTimeEvent,
+        anchor_request: &AnchorRequest,
+    ) -> Result<TimeEventInsertable> {
+        let time_event_cid = time_event.to_cid().context(format!(
```
dav1do, line 141: are any of these failures retry-able? I'm wondering if we should we just log a warning and ignore them or drop them from the next batch/attempt?

anchor-service/src/anchor.rs
```diff
@@ -76,15 +100,123 @@ pub struct TimeEventBatch {
...
+                self.build_time_event_insertable(time_event, anchor_request)
+            })
+            .collect::<Result<Vec<_>>>()?;
+        Ok(events)
+    }
+
+    /// Build a TimeEventInsertable from a RawTimeEvent and AnchorRequest
+    fn build_time_event_insertable(
+        &self,
+        time_event: &RawTimeEvent,
```
dav1do, line 138: should this be owned since we're cloning out of it? it seems like a move makes sense and the semantic of consuming raw and returning regular seems reasonable.

anchor-service/src/anchor_batch.rs
```diff
@@ -14,31 +21,121 @@ use crate::{
...
+    }
+
+    async fn process_next_batch(&mut self) -> Result<()> {
+        // Pass the anchor requests through a deduplication step to avoid anchoring multiple Data Events from the same
+        // Stream.
+        let high_water_mark = self
+            .high_water_mark_store
+            .high_water_mark()
+            .await
+            .expect("error getting high water mark from database");
```
dav1do, line 106: this error seems possibly okay... maybe a result rather than crashing the process (or at least matching on fatal/timeout/etc to choose when to be loud)?

anchor-service/src/anchor_batch.rs
```diff
@@ -14,31 +21,121 @@ use crate::{
...
+            )
+            .await
+        {
+            Ok(requests) => IndexMap::<Cid, AnchorRequest>::from_iter(
+                requests.into_iter().map(|request| (request.id, request)),
+            )
+            .into_values()
+            .collect(),
+            Err(e) => {
+                return Err(e);
```
dav1do, line 123: if this fails, what happens? it seems like it will just keep failing forever or am I misunderstanding?

anchor-service/src/anchor_batch.rs
```diff
@@ -14,31 +21,121 @@ use crate::{
...
+            }
+        };
+        if anchor_requests.is_empty() {
+            info!("no requests to anchor");
+            return Ok(());
+        }
+        // Anchor the batch to the CAS. This may block for a long time.
```
dav1do, line 130: I don't think this will actually block the runtime, right? It is expected to take a long time, but it uses async IO so it will allow other uses of the system thread.. if that's not true, we should consider starting our own runtime or using system threads (but we should probably benchmark first)

anchor-service/src/high_water_mark_store.rs
```diff
@@ -0,0 +1,32 @@
+use anyhow::Result;
+use ceramic_sql::sqlite::SqlitePool;
+
+#[derive(Debug)]
+pub struct HighWaterMarkStore {
```
dav1do, line 5: I'm fine with this name since that's all it does right now, but wondering if we could call it Anchor*Store (AnchorRequestStore, AnchorServiceStore, AnchorStore)?

core/src/event_id.rs
```diff
@@ -141,6 +141,24 @@ impl EventId {
             cid,
         })
     }
+
+    /// swap_cid replaces the event CID in the EventId with a new CID.
+    pub fn swap_cid(&self, cid: &Cid) -> Option<Self> {
```
dav1do, line 146: This seems a bit odd.. do we actually need it (e.g. can we just call `EventId::new(..old values.., new_cid)`) or is there a way to use the builder to make this more natural?

If so, should we consume the EventId (self) or does it need to stay around? It seems like this should be taking a previously valid EventId and returning a transformed one rather than keeping both around.

migrations/sqlite/20240809205107_event_init_cid_informant_type.up.sql
```diff
@@ -6,4 +6,7 @@ ALTER TABLE ceramic_one_event ADD COLUMN init_cid BLOB DEFAULT NULL;
 -- The ID of the node that sent the event or our own ID if the event came in via API
 ALTER TABLE ceramic_one_event ADD COLUMN informant TEXT DEFAULT NULL;
 
-SELECT init_cid, informant FROM "ceramic_one_event" WHERE false;
+-- Whether this is a Time Event or not
+ALTER TABLE ceramic_one_event ADD COLUMN is_time_event BOOLEAN DEFAULT FALSE;
```
dav1do, line 10: I'm pretty sure we can't move a migration after it's been applied anywhere. I think we just need to make a new one and leave the old name.

In terms of this column, maybe we should just add event_type integer and create an enum that's 0/1/2 for init/time/data?

migrations/sqlite/20240816213347_anchor_high_water_mark.up.sql
```diff
@@ -0,0 +1,9 @@
+-- Add up migration script here
+
+CREATE TABLE IF NOT EXISTS ceramic_one_anchor_high_water_mark (
+    high_water_mark INTEGER NOT NULL,
+    timestamp TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
dav1do, line 5: maybe put the column name in quotes so since it's a reserved word

recon/src/protocol.rs
```diff
@@ -90,6 +90,7 @@ pub async fn initiate_synchronize<S, R, E>(
     recon: R,
     stream: S,
     config: ProtocolConfig,
+    informant: NodeId,
```
dav1do, line 93: do we need NodeId here and in the config?

dav1do (commented):
Happy to meet to discuss or pair on any of my comments/questions.

service/src/event/service.rs
smrz2001: @dav1do, is it ok if we do that in a follow-up PR? I'd really like to get this PR in before I'm OOO next week. Or... we could wind it up together during the retreat lol though that means pushing it out 2 more weeks...

one/src/daemon.rs
```diff
@@ -386,6 +435,43 @@ pub async fn run(opts: DaemonOpts) -> Result<()> {
...
+                opts.anchor_poll_retry_count,
+            );
+            let mut anchor_service = AnchorService::new(
+                Arc::new(remote_cas),
+                event_svc.clone(),
+                sqlite_pool.clone(),
+                node_id,
+                Duration::from_secs(opts.anchor_interval),
+                opts.anchor_batch_size,
+            );
```
nathanielc, line 461: I really like how simple an straightforward this dependency injection is working. Create a service using its new method and pass it references of the deps.

anchor-service/src/anchor.rs
```diff
@@ -1,16 +1,36 @@
...
+    /// the ResumeToken for the Event being anchored in the local database.
+    pub resume_token: i64,
+}
+
+impl Debug for AnchorRequest {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("AnchorRequest")
+            .field("id", &format!("{:?}", &self.id))
+            .field("prev", &format!("{:?}", &self.prev))
+            .field("order_key", &format!("{:?}", &self.event_id))
```
nathanielc, line 30: why not rename of the event_id field to order_key ?

anchor-service/src/anchor_batch.rs
```diff
@@ -14,31 +21,121 @@ use crate::{
...
+    }
+
+    /// Run the AnchorService:
+    /// - Get anchor requests from the AnchorClient
+    /// - Deduplicate the anchor requests
+    /// - Anchor the batch using a Transaction Manager
+    /// - Store the TimeEvents using the AnchorClient
+    ///
+    /// This function will run indefinitely, or until the process is shutdown.
+    pub async fn run(&mut self, shutdown_signal: impl Future<Output = ()>) {
```
nathanielc, line 73: I do not see any tests of this method. Can we an an integration test that shows this basic flow working?

core/src/network.rs
```diff
@@ -1,12 +1,14 @@
+use serde::{Deserialize, Serialize};
+
 /// Network values from https://cips.ceramic.network/tables/networkIds.csv
 /// Ceramic Pubsub Topic, Timestamp Authority
-#[derive(Debug, Clone)]
+#[derive(Serialize, Deserialize, Debug, Clone)]
```
nathanielc, line 5: curious where we are serializing the network?

event/src/unvalidated/event.rs
```diff
@@ -239,6 +239,23 @@ where
             )),
         }
     }
+
+    /// Get the CID of the init event of the stream for this event
+    pub fn stream_cid(&self) -> &Cid {
```
nathanielc, line 244: This method already exists verbatim above but as the the `id()` method. I like the `stream_cid` name better but we should only have one method.

one/src/daemon.rs
```diff
@@ -178,6 +180,41 @@ pub struct DaemonOpts {
...
+    /// Ceramic One anchor batch size
+    #[arg(
+        long,
+        default_value_t = 1_000_000,
+        hide = true,
+        env = "CERAMIC_ONE_ANCHOR_BATCH_SIZE"
+    )]
+    anchor_batch_size: u64,
+
+    /// Ceramic One anchor polling interval in seconds
```
nathanielc, line 201: can you add a little context on what polling does? Answer the question of why I want a short or longer interval. Also its not clear how this interval is different from the anchor_interval above.

nathanielc (changes requested):
Overall I really like how the service now lives on its own and how it has clear dependencies on the event-svc. A few small changes requested but nothing too large.

anchor-service/src/anchor.rs
AaronGoldman: we still talk about `ceramic_core::EventId` in some places as an "order keys" and in some places as "event ids". We should pick one. ... my preference to use order_key everywhere.

anchor-service/src/anchor_batch.rs
AaronGoldman: It will not block a system thread but the anchoring can't proceed until an anchor land.

event/src/unvalidated/event.rs
AaronGoldman: I will remove id()

recon/src/protocol.rs
AaronGoldman: nope, removed

anchor-service/src/anchor.rs
AaronGoldman: Who would own the TimeEventInsertables in that case.

anchor-service/src/anchor.rs
AaronGoldman: We would have to work around the partial move.

```rs
error[E0382]: borrow of partially moved value: `self`
   --> anchor-service/src/anchor.rs:128:18
    |
127 |             .into_iter()
    |              ----------- `self.raw_time_events.events` partially moved due to this method call
128 |             .map(|(anchor_request, time_event)| {
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ value borrowed here after partial move
129 |                 self.build_time_event_insertable(time_event, anchor_request)
    |                 ---- borrow occurs due to use in closure
    |
```

try_to_insertables calls self.build_time_event_insertable
build_time_event_insertable calls self.find_tree_blocks_along_path

anchor-service/Cargo.toml
dav1do: nah, this is fine. Since we moved all the queries into the service crates so it makes sense.

anchor-service/src/anchor.rs
dav1do: yeah, this is fine. I thought more about it and we're already reading everything from the db in one shot so there's no point in streaming anywhere until we stream that

anchor-service/src/anchor_batch.rs
dav1do: yeah, that's fair. it's a nit. you could modify the comment along those lines but no worries.

anchor-service/src/anchor.rs
dav1do: yeah, I mean we might have to still call clone, and it might not even change the generated code. I mostly meant whether the API semantic should be to take ownership and consume them into something else, but if it doesn't help the internal code it's fine. 

I just generally like API design that doesn't have internal clones and let's the higher level decide whether to move/clone, but sometimes when you only need a partial field it's better to borrow the whole thing and simply clone the smaller pieces out.

anchor-service/src/anchor.rs
dav1do: okay, we can come back to this in the future

anchor-service/src/anchor.rs
nathanielc: Works for me, I say rename the field now. Then in other places can be renamed as we go along. Gotta start somewhere.

anchor-service/src/anchor_batch.rs
AaronGoldman: This Error propagates up to the anchor loop. The Anchor loop is just logging errors and trying again at the next tick. Since this is a call to the local sqlite database if it fails we are in a very bad state. I think it is worth crashing and not log and continue.

core/src/event_id.rs
AaronGoldman: So we need to have the new time event know the recon EventID so we can construct an event insertable. What we have is the recon EventID of the data event being anchored.  To use the builder, we would need information that is not in the Events table. e.g. network ID, separator key, separator value, controller. 
We have two options. Crate a builder that circumvents the protections and lets you just construct with raw values, deconstruct a event id to parts, send the parts around and recombine the parts or jsut send the eventID and have an update CID.

I picked this method as it keeps the structure of the eventID opaque to the anchor service. All it knows it 
```rs
f(data event eventID, time event CID) -> time event EventID
```

migrations/sqlite/20240809205107_event_init_cid_informant_type.up.sql
AaronGoldman: done

nathanielc (approved):

PR #490, opened 2024-08-19
chore: Remove EventMetadata type

stbrody:
Also changes a bunch of places to use references to Cids instead of returning or taking by value.  Cid implements Copy which means there might a lot of silent copying happening under the hood without it being obvious from the code, and since Cids can be larger than a memory address, using references where possible seems prudent.  Have that change split into its own commit if you want to see the changes separately.  Can also split into two PRs if preferred.

### DIFF
--- a/event/src/unvalidated/event.rs
+++ b/event/src/unvalidated/event.rs
@@ -79,11 +79,11 @@ where
     }
 
     /// Returns the prev CID (or None if the event is an init event)
-    pub fn prev(&self) -> Option<Cid> {
+    pub fn prev(&self) -> Option<&Cid> {
         match self {
             Event::Time(t) => Some(t.prev()),
             Event::Signed(event) => match event.payload() {
-                Payload::Data(d) => Some(*d.prev()),
+                Payload::Data(d) => Some(d.prev()),
                 Payload::Init(_) => None,
             },
             Event::Unsigned(_) => None,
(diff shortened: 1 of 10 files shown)


event/src/unvalidated/event.rs
```diff
@@ -90,6 +90,19 @@ where
         }
     }
 
+    /// Returns the 'id' field of the event, which is the Cid of the stream's init event.
+    /// If this event *is* the init event, then it doesn't know its own Cid and returns None.
```
dav1do, line 94: "it doesn't know its own CID" feels wrong xD

dav1do (approved):
Nice, thanks for removing/consolidating the metadata stuff.

Regarding CIDs, I don't know that using references is necessarily better... Nathaniel and I talked about it once because it's annoying that they're 1.5 words, but having to follow the pointer rather than just loading the value directly is a tradeoff.. In reality, rustc/llvm will likely make a similar optimization in both cases so it probably doesn't matter much but who knows. That said, I don't feel strongly enough to _not_ do this, but without benchmarks I don't think we should really worry about it. My philosophy is to stick with copying `Copy` types unless someone sticks it on a big struct that is obviously a mistake.

event/src/unvalidated/event.rs
stbrody: Haha, open to suggestions for alternate wording!

dav1do: I got nothing, just teasing because it's a silly concept

PR #492, opened 2024-08-19
chore: modify cacao, unvalidated module, jwk to add functionality needed for event validation

samika98:
modify cacao rs to support event validation : most of the code is pulled in from danny's private data branch : https://github.com/ceramicnetwork/rust-ceramic/pull/427
add getters to fetch envelope info needed for validating signed events 
support did resolution in jwk : add functionality to resolve a did to extract it's didDocument ( contains signing info like public key needed to verify signature)


event/src/unvalidated/signed/mod.rs
```diff
@@ -192,6 +197,34 @@ impl Envelope {
                 })
         })
     }
+
+    /// Get the signature
+    pub fn get_signature(&self) -> &Bytes {
+        &self.signatures[0].signature
```
dav1do, line 203: This will panic if the vec is empty. Since we use a builder, we shouldn't have to worry about it right now, but maybe in the future we can add a custom field deserializer to error if it's ever empty so we are guaranteed to have one at least one afterward.

But more to the point, maybe we can make this `signatures(&self) ->&[Signature]` and let the caller deal with it.

core/src/jwk.rs
```diff
@@ -65,6 +67,24 @@ impl Jwk {
         jwk.key_id.clone_from(&self.key_id);
         Ok(Self(jwk))
     }
+
+    /// Resolve a DID and return the DIDDocument
+    ///
+    /// The DID Document contains the public key. For more info, see:
+    /// https://www.w3.org/TR/did-core/#dfn-did-documents
+    pub async fn resolve_did(did: &str) -> Option<DidDocument> {
+        let resolver = &did_method_key::DIDKey;
```
dav1do, line 76: Were you thinking this would support did:pkh as well (if not, we should probably rename it). I'm not convinced it belongs in here and instead should be separate but this is fine until we start using it. 

Some other thoughts we can address in a follow up: The resolver should maybe go in a lazy cell so we don't need to create one every time (but I'm not 100% sure the impl details). If we keep it here, we could check the did return before we allocate so we don't need to create something we don't use.

core/src/jwk.rs
```diff
@@ -65,6 +67,24 @@ impl Jwk {
...
+        let input_metadata = ResolutionInputMetadata::default();
+        let base_did = match did.split('#').next() {
```
dav1do, line 78: `let base = did.split_once("#").map_or(did, |(b, _| b)` is probably nicer/more correct. I'm not sure if it's allowed, but this would still work if it didn't include the #.

event/src/unvalidated/signed/mod.rs
```diff
@@ -192,6 +197,34 @@ impl Envelope {
...
+    }
+
+    /// Get the signed payload
+    pub fn get_payload(&self) -> &Bytes {
+        &self.payload
+    }
+
+    /// Construct the jws header from the signature protected bytes
+    pub fn get_header(&self) -> Result<ssi::jws::Header, anyhow::Error> {
```
dav1do, line 212: we don't usually include `get` in the names.. maybe call this `jws_header()` or `header_as_jws`? same about removing `get` from `get_payload` and `get_signature`

event/src/unvalidated/signed/mod.rs
```diff
@@ -68,6 +68,11 @@ impl<D: serde::Serialize> Event<D> {
         &self.payload
     }
 
+    /// Get the Envelope
+    pub fn envelope(&self) -> &Envelope {
```
dav1do, line 72: Another one I needed and can add unless you want to add it here is 

```
    /// Get the capability if one exists
    pub fn capability(&self) -> Option<&Capability> {
        self.capability.as_ref().map(|(_, ref c)| c)
    }
    ```

dav1do (commented):
Thanks for doing this, a few comments/more things we'll need, but feel free to comment if you want to skip them and I'll add them in my next PR.

core/src/jwk.rs
samika98: Nice catch! This is a better implementation!

core/src/jwk.rs
samika98: Agree with the addition of an early return before we do any allocations! Thanks for pointing that out! Will add that!

After looking into the current implementation, it seems that DIDKey and DIDPKH are already being used efficiently as static references.
The DIDPKH struct is actually a zero-sized type, which means it's incredibly cheap to construct. Given this, creating a lazy cell might not provide significant performance benefits and could potentially add unnecessary complexity to the code.
That being said, I'm curious about your thoughts on this. Did you have any specific scenarios in mind where a lazy cell might be particularly beneficial? Are there any performance bottlenecks you've observed that led to this suggestion? It would be great to hear more about your reasoning, as there might be aspects we haven't considered yet.

dav1do (approved):

core/src/jwk.rs
dav1do: Nice, this seems good then. My only reasoning was seeing that it was shared in the private data poc and I assumed it was at least an allocation to box it but if we can just use &dyn and it's free, let's do that :).

PR #493, opened 2024-08-19
chore: Add logic to dispatch validation based on event type

stbrody:


service/src/event/service.rs
```diff
@@ -129,23 +129,72 @@ impl CeramicEventService {
...
+                Ok(insertable) => parsed_events.push(insertable),
+                Err(err) => invalid_events.push(InvalidItem::InvalidFormat {
                     key: event.key.clone(),
                     reason: err.to_string(),
                 }),
             }
         }
 
+        // Group events by their type
+        let mut valid_events = Vec::new();
```
stbrody, line 162: there's a lot of allocations in this function with the way the various vectors are initialized and added to.

I'm definitely open to ideas on how to optimize this.

dav1do: I'm not too worried about the allocations yet, but we should try to use `with_capacity()` whenever we have a decent guess because it defaults to 0 and only grows by 2x, so if we have to put in 100 events we'll have to reallocate the vec with sizes 1,2,4,8,16,32,64,128.

I think we can figure out how to be smarter about allocations/resource utilization after we get things working, maybe we send on a channel in the loop, or use an arena to get nice allocations, or we restructure things a different way. Someone has to loop through and split things up, but it'd be nice if we could reuse the allocations across the different jobs (validate, order, etc).

service/src/event/service.rs
```diff
@@ -129,23 +129,72 @@ impl CeramicEventService {
...
+                    signed_events.push(event);
+                }
+                Event::Unsigned(_) => {
+                    // Unsigned events need no extra validation.
+                    valid_events.push(event);
+                }
+            }
+        }
+
+        let (valid_signed, invalid_signed) = Self::validate_signed_events(signed_events).await?;
```
dav1do, line 180: maybe use try_join! on the signed/time event futures?

dav1do (commented):
A couple small comments but it seems like a good start to delegate to the validation components.

service/src/event/service.rs
```diff
@@ -129,23 +129,72 @@ impl CeramicEventService {
...
-        items: &[recon::ReconItem<EventId>],
-        source: DeliverableRequirement,
-    ) -> Result<InsertResult> {
-        let mut to_insert = Vec::new();
-        let mut invalid = Vec::new();
+    async fn validate_signed_events(
+        events: Vec<EventInsertable>,
+    ) -> Result<(Vec<EventInsertable>, Vec<InvalidItem>)> {
+        // TODO: IMPLEMENT THIS
+        Ok((events, Vec::new()))
```
AaronGoldman, line 136: should this be `unimplemented!();` or `todo!()`

stbrody: That will cause things to start breaking since it's not behind a feature flag yet.  I'm not too worried about us forgetting to do this 😅

service/src/event/service.rs
stbrody: ooh yes good call, updated

service/src/event/service.rs
```diff
@@ -129,23 +130,77 @@ impl CeramicEventService {
...
+                Err(err) => invalid_events.push(InvalidItem::InvalidFormat {
                     key: event.key.clone(),
                     reason: err.to_string(),
                 }),
             }
         }
 
+        // Group events by their type
+        let mut valid_events = Vec::with_capacity(parsed_events.len());
+        let mut signed_events = Vec::with_capacity(parsed_events.len());
```
stbrody, line 164: I'm over-allocating here, since events will either be signed or time, each of these will definitely have less than `parsed_events.len()`.  But I figured over-allocating a little bit is better than having multiple allocations as the vector grows, and keeping these allocations a somewhat consistent size might make it easier for the allocator to be more efficient with re-using the memory.

dav1do (approved):

PR #495, opened 2024-08-20
feat: eth and sol pkh cacao validation

dav1do:
Creates a new `validation` crate and provides a `Verifier` trait implemented on CACAOs. It currently only understands Ethereum and Solana PKH and is not used outside of tests. I added some new crates for crypto libraries to the root (we mostly had them transitively through spruceID `ssi` but didn't have direct access). To validate a CACAO we must first rebuild the SIWx message the user saw, which is included. The CAIP-2 (chain)/CAIP-10 (account) parsing comes from the `ssi-caips` crate. During testing, we included a solana address that wasn't valid and it failed, but the tests need to be updated to include this.

This uses the test CAR files generated by the [ceramic-sdk](https://github.com/ceramicstudio/ceramic-sdk/pull/17) and parses them to build events of different types/signing methods. Currently it only uses the valid eth/sol events and I need to update the tests for the invalid signatures and did:key methods (those are simply included to demonstrate it builds/parses). There are tests for an existing CAR file that checks the `VerifierOpts` logic around skew, grace period, "at time" for ethereum (that logic is shared for all CACAOs).


dav1do:
bleh, actually want ticket 32 so closing and re-creating

PR #496, opened 2024-08-20
feat: eth and sol pkh cacao validation

dav1do:
Creates a new `validation` crate and provides a `Verifier` trait implemented on CACAOs. It currently only understands Ethereum and Solana PKH and is not used outside of tests. I added some new crates for crypto libraries to the root (we mostly had them transitively through spruceID `ssi` but didn't have direct access). To validate a CACAO we must first rebuild the SIWx message the user saw, which is included. The CAIP-2 (chain)/CAIP-10 (account) parsing comes from the `ssi-caips` crate. During testing, we included a solana address that wasn't valid and it failed, but the tests need to be updated to include this.

This uses the test CAR files generated by the [ceramic-sdk](https://github.com/ceramicstudio/ceramic-sdk/pull/17) and parses them to build events of different types/signing methods. Currently it only uses the valid eth/sol events and I need to update the tests for the invalid signatures and did:key methods (those are simply included to demonstrate it builds/parses). There are tests for an existing CAR file that checks the `VerifierOpts` logic around skew, grace period, "at time" for ethereum (that logic is shared for all CACAOs).


validation/src/verifier/cacao_verifier.rs
```diff
@@ -0,0 +1,77 @@
...
+    fn verify_time_checks(&self, opts: &VerifyOpts) -> anyhow::Result<()>;
+}
+
+impl Verifier for Capability {
+    fn verify(&self, opts: &VerifyOpts) -> anyhow::Result<()> {
+        // verify signed from js-did is not required as it won't deserialize without a signature
+        // is that something that is ever expected?
+        self.verify_time_checks(opts)?;
+
+        // TODO: move tests to this and remove time_check from pkh impls
```
dav1do, line 41: clean up... this is done

validation/src/signature/pkh_ethereum.rs
```diff
@@ -0,0 +1,259 @@
...
+                )
+                .unwrap_or_default();
+            }
+        }
+
+        // 2022-09-20 in milliseconds
+        // TODO: use lazy static/const for LEGACY_CHAIN_ID_REORG_DATE
+        if recovered != issuer
+            && cacao.payload.issued_at
+                < chrono::DateTime::from_timestamp_millis(1663632000000).unwrap()
```
dbcfd, line 48: `std::once_cell::Lazy` for this?

validation/src/signature/pkh_ethereum.rs
```diff
@@ -0,0 +1,259 @@
...
+        Ok(())
+    }
+
+    fn verify_message(msg: &str, signature: &str) -> Result<String> {
+        // Create the prefixed message as per the Ethereum standard
+        let digest = keccak::hash_personal_message(msg);
+        let signature_bytes = hex::decode(signature.strip_prefix("0x").unwrap_or(signature))?;
+
+        // eth offsets the 0-1 recovery ID by 27 and doesn't use 2 or 3
+        let mut v = signature_bytes[64];
```
dbcfd, line 69: Should we worry that there aren't 64 bytes?

validation/src/siwx_message.rs
```diff
@@ -0,0 +1,258 @@
...
+// use ssi_caips::{caip10::BlockchainAccountId, caip2::ChainId};
+use crate::cacao::Capability;
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct SiwxMessage<'a> {
+    ///RFC 4501 dns authority that is requesting the signing.
+    domain: &'a str,
+    /// Address performing the signing conformant to capitalization
+    /// encoded checksum specified in EIP-55 where applicable if Ethereum.
+    address: String,
```
dbcfd, line 15: Not a huge fan of the inconsistence between `&'a str` and `String`. Seems like the owned fields are coming from the account, so maybe split this up?

validation/src/verifier/cacao_verifier.rs
```diff
@@ -0,0 +1,76 @@
...
+    pub clock_skew: chrono::Duration,
+    /// Do not verify expiration time when false
+    pub check_exp: bool,
+}
+
+impl Default for VerifyOpts {
+    fn default() -> Self {
+        Self {
+            at_time: None,
+            revocation_phaseout_secs: chrono::Duration::new(0, 0).expect("0 is a valid duration"),
```
dbcfd, line 22: lazy this?

validation/src/signature/pkh_ethereum.rs
dav1do: Yeah, I'll add a check. I thought this was being hashed and would end up the right length but that's the other values in this function.

validation/src/siwx_message.rs
dav1do: Agreed.. I'll see what I can do about nesting owned things. I took this straight from js-did in terms of format. I tried not to clone the entire cacao and borrow what I could easily, but I can't extend the lifetimes of strings I have to generate from it and didn't feel like adding `'b` (or using a bunch of `Option<'&a str>`... it's internal details at this point and only used to generate the siwx message, which ends up allocating a bunch of strings anyway.

dav1do: used all borrows except for account (replaced address/chain_id) and kept DateTime references that are converted for the message

dbcfd (approved):

PR #497, opened 2024-08-20
feat: jws signature validation

dav1do:
Implements event validation using did:key/JWS. This verifies the envelope and makes sure the cacao grants access to the key that signed the payload. This builds on #495 (current target) and I had to make the cacao/pkh `Verifier` async since it can be a JWS based on the metadata.

This is not invoked anywhere and the tests currently only verify signatures for the valid events from the ceramic-sdk test vectors.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1722,6 +1722,7 @@ name = "ceramic-validation"
 version = "0.33.0"
 dependencies = [
  "anyhow",
+ "async-trait",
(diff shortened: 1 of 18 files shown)


dav1do:
OHMYGSH how can I never get the linear tickets right??? This is 2**8**3 not 2**9**3.. since there's no unliking or changing the branch name... closing and trying again 😭 😭 💀

PR #498, opened 2024-08-20
feat: jws signature validation

dav1do:
Implements event validation using did:key/JWS. This verifies the envelope and makes sure the cacao grants access to the key that signed the payload. This builds on #495 (current target) and I had to make the cacao/pkh `Verifier` async since it can be a JWS based on the metadata.

This is not invoked anywhere and the tests currently only verify signatures for the valid events from the ceramic-sdk test vectors.


validation/src/verifier/key_verifier.rs
```diff
@@ -0,0 +1,76 @@
...
+    header: &[u8],
+    payload: &[u8],
+    alg: ssi::jwk::Algorithm,
+    signature: &[u8],
+    delegated_kid: Option<&str>,
+) -> Result<()> {
+    let did_doc = Jwk::resolve_did(did)
+        .await
+        .context("failed to resolve did")?;
+    if delegated_kid.is_some() && delegated_kid.unwrap() != did_doc.id {
```
dav1do, line 53: `map_or(false, |kid| kid != did_doc.id)`

validation/src/verifier/key_verifier.rs
```diff
@@ -0,0 +1,76 @@
...
+            jws_header.algorithm,
+            signature.signature.as_slice(),
+            delegated_kid,
+        )
+        .await
+    }
+}
+
+pub(crate) async fn verify_did_jws(
+    did: &str,
```
dav1do, line 43: maybe wrap these params in an input struct

validation/src/verifier/key_verifier.rs
```diff
@@ -0,0 +1,76 @@
...
+        bail!("jws kid not delegated access",);
+    }
+
+    let jwk = Jwk::new(&did_doc)
+        .await
+        .map_err(|e| anyhow!("failed to generate jws for did: {}", e))?;
+
+    let header_str = engine::general_purpose::STANDARD_NO_PAD.encode(header);
+    let payload_cid = base64::engine::general_purpose::URL_SAFE_NO_PAD.encode(payload);
+    let digest = format!("{header_str}.{payload_cid}");
```
samika98, line 68: Could you add a comment and doc link explaining this?

validation/src/verifier/key_verifier.rs
```diff
@@ -0,0 +1,76 @@
...
+            self.payload().as_slice(),
+            jws_header.algorithm,
+            signature.signature.as_slice(),
+            delegated_kid,
+        )
+        .await
+    }
+}
+
+pub(crate) async fn verify_did_jws(
```
samika98, line 42: Can we add a docstring here?

validation/src/verifier/key_verifier.rs
```diff
@@ -0,0 +1,76 @@
...
+    did: &str,
+    header: &[u8],
+    payload: &[u8],
+    alg: ssi::jwk::Algorithm,
+    signature: &[u8],
+    delegated_kid: Option<&str>,
```
samika98, line 48: why is this called delegated_kid, rather than just a kid?

validation/src/verifier/key_verifier.rs
```diff
@@ -0,0 +1,76 @@
...
+
+#[async_trait::async_trait]
+pub trait Verifier {
+    /// Verify the signature. Optoinally requiring it matches a delegated KID
+    async fn verify_signature(&self, delegated_kid: Option<&str>) -> Result<()>;
+}
+
+#[async_trait::async_trait]
+impl Verifier for Envelope {
+    async fn verify_signature(&self, delegated_kid: Option<&str>) -> Result<()> {
```
samika98, line 15: Should return a valid event struct in the future. just putting this here so that we don't forget about it.

samika98 (approved):
Looks great! Just a nit about adding dcstrings and comments, and also a question about naming.

validation/src/verifier/opts.rs
```diff
@@ -0,0 +1,22 @@
+#[derive(Clone, Debug)]
+pub struct VerifyOpts {
+    /// The point in time at which the capability must be valid, defaults to `now`
+    pub at_time: Option<chrono::DateTime<chrono::Utc>>,
+    /// How long the capability stays valid for after it was expired
+    pub revocation_phaseout_secs: chrono::Duration,
+    /// The clock tolerance when verifying iat, nbf, and exp
+    pub clock_skew: chrono::Duration,
+    /// Do not verify expiration time when false
+    pub check_exp: bool,
```
dav1do, line 10: remove this boolean and use at_time without a default (none being do not check exp)

validation/src/verifier/key_verifier.rs
dav1do: yeah, we could add something in the future. I was holding off on a "validated event" because that requires more than just a signature (and this is a "can you trust what was inside this"). I was hoping to add that concept to the service in my next PR and make any changes here as needed.

validation/src/verifier/key_verifier.rs
dav1do: this is related to step 4 of [cacao verification](https://developers.ceramic.network/docs/protocol/js-ceramic/accounts/object-capabilities#verification).. even if the cacao was signed and the envelope was signed correctly, the cacao had to actually delegate to the key that signed the envelope. I don't care for this API signature but it's mostly private/implementation details right now.

validation/src/verifier/opts.rs
dav1do: actually, this won't quite work since we use the time (now) to check the nbf/iat fields in addition to exp.. so leaving as is for now

validation/src/verifier/key_verifier.rs
dav1do: I tried adding an enum but that was maybe overkill and kind of clunky.. anyway, i can drop the enum if we just want a `delegated_kid: Option<&'a str>` on the input struct 🤷

validation/src/verifier/event_verifier.rs
```diff
@@ -0,0 +1,61 @@
+use anyhow::Result;
+use ceramic_event::unvalidated::signed;
+use ipld_core::ipld::Ipld;
+
+use super::{cacao_verifier::Verifier as _, key_verifier::Verifier as _, opts::VerifyOpts};
+
+#[async_trait::async_trait]
+pub trait Verifier {
+    async fn verify_signature(&self, opts: &VerifyOpts) -> Result<()>;
```
dav1do, line 9: I'm pretty sure this is going to need more parameters or to return something so we can make sure to verify controller and more things but I'm planning to adjust that when I start using it in the next PR.

PR #501, opened 2024-08-22
feat: migrate from file list

smrz2001:
This PR adds support for migrating IPFS blocks from an input file list. It also introduces both an `offset` and a `limit` that can be used to paginate migrations, whether from a directory or from a file list.

This is needed for being able to apply incremental ZFS snapshots without having to re-migrate the entire blockstore. The list of added blocks can be extracted from the incremental snapshot and input to the migration so that only new blocks are migrated. This will dramatically reduce the amount of time needed for migrating new blocks added after the first snapshot was taken.

Note that paginating over a directory might have unpredictable results if it is receiving live updates.

### DIFF
--- a/one/src/migrations.rs
+++ b/one/src/migrations.rs
@@ -11,7 +11,7 @@ use clap::{Args, Subcommand};
 use futures::{stream::BoxStream, StreamExt};
 use multihash_codetable::{Code, Multihash, MultihashDigest};
 use tokio::io::AsyncBufReadExt;
-use tracing::{debug, info, trace};
+use tracing::{debug, info};
 
 use crate::{default_directory, DBOpts, Info, LogOpts};
 
@@ -64,7 +64,31 @@ pub struct FromIpfsOpts {
     #[command(flatten)]
     log_opts: LogOpts,
 
-    /// Path of file containing list of newline-delimited absolute file paths to migrate
+    /// Path of file containing list of newline-delimited absolute file paths to migrate.
+    ///
+    /// See below for example usage when running a migration for a live IPFS node. Multiple migration runs using lists
+    /// of files that have changed between runs is useful for incremental migrations. This method can also be used for
+    /// a final migration after shutting down the IPFS node so that all inflight blocks are migrated to the new C1 node.
+    ///
+    /// # Get list of files in sorted order
+    ///
+    /// find ~/.ipfs/blocks -type f | sort > first_run_files.txt
+    ///
+    /// # Run the migration
+    ///
+    /// ceramic-one migrations from-ipfs --input-ipfs-path ~/.ipfs/blocks --input-file-list-path first_run_files.txt
+    ///
+    /// # Get updated list of files in sorted order
+    ///
+    /// find ~/.ipfs/blocks -type f | sort > second_run_files.txt
+    ///
+    /// # Use comm to get the list of new files
+    ///
+    /// comm -13 first_run_files.txt second_run_files.txt > new_files.txt
+    ///
+    /// # Re-run the migration
+    ///
+    /// ceramic-one migrations from-ipfs --input-ipfs-path ~/.ipfs/blocks --input-file-list-path new_files.txt
     #[clap(long, short = 'f', env = "CERAMIC_ONE_INPUT_FILE_LIST_PATH")]
     input_file_list_path: Option<PathBuf>,
 
@@ -197,7 +221,7 @@ impl BlockStore for FSBlockStore {
 }
 
 impl FSBlockStore {
-    fn blocks_from_dir(&self) -> BoxStream<'static, anyhow::Result<(Cid, Vec<u8>)>> {
+    fn blocks_from_dir(&self) -> BoxStream<'static, Result<(Cid, Vec<u8>)>> {
         // the block store is split in to 1024 directories and then the blocks stored as files.
         // the dir structure is the penultimate two characters as dir then the b32 sha256 multihash of the block
         // The leading "B" for the b32 sha256 multihash is left off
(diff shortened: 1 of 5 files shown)


one/src/migrations.rs
```diff
@@ -150,6 +158,85 @@ impl BlockStore for FSBlockStore {
...
+            .expect("input_file_list_path should have been checked before calling this function");
+        info!(path = %input_file_list_path.display(), "reading IPFS file list");
+
+        let offset = self.file_offset;
+        let limit = self.file_limit;
+        (try_stream! {
+            let file = tokio::fs::File::open(input_file_list_path).await?;
+            let mut lines = tokio::io::BufReader::new(file).lines();
+            let mut i = 0;
+            while let Some(line) = lines.next_line().await? {
```
nathanielc, line 216: There are [skip](https://docs.rs/tokio-stream/latest/tokio_stream/trait.StreamExt.html#method.skip) and [take](https://docs.rs/tokio-stream/latest/tokio_stream/trait.StreamExt.html#method.take)  methods on streams using [StreamExt](https://docs.rs/tokio-stream/latest/tokio_stream/trait.StreamExt.html). Which should simplify this logic to `let lines = lines.skip(offset).take(limit);` and the simply have the while loop without any bounds checks.

one/src/migrations.rs
```diff
@@ -150,6 +158,85 @@ impl BlockStore for FSBlockStore {
...
+        // the block store is split in to 1024 directories and then the blocks stored as files.
+        // the dir structure is the penultimate two characters as dir then the b32 sha256 multihash of the block
+        // The leading "B" for the b32 sha256 multihash is left off
+        // ~/.ipfs/blocks/QV/CIQOHMGEIKMPYHAUTL57JSEZN64SIJ5OIHSGJG4TJSSJLGI3PBJLQVI.data // cspell:disable-line
+        info!(path = %self.input_ipfs_path.display(), "opening IPFS repo");
+
+        let mut dirs = Vec::new();
+        dirs.push(self.input_ipfs_path.clone());
+
+        let offset = self.file_offset;
```
nathanielc, line 172: I don't this we should support limit/offset for the read dir case. From the docs:
 
> The order in which this iterator returns entries is platform and filesystem dependent.

We cannot guarantee that limit/offset will be meaningful. And in practice the order in which blocks where traversed appeared to be random. Meaning it was not in lexicographical order on my local filesystem nor in various production filesystems we used.

I propose if limit/offset are not None and there is no input file list then we error.

nathanielc (changes requested):
Good addition, some UX changes proposed but nothing significant

one/src/migrations.rs
nathanielc: https://doc.rust-lang.org/nightly/std/fs/fn.read_dir.html

one/src/migrations.rs
AaronGoldman: [tokio](https://docs.rs/tokio/latest/tokio/index.html)::[io](https://docs.rs/tokio/latest/tokio/io/index.html)::[Lines](https://docs.rs/tokio/latest/tokio/io/struct.Lines.html#) did not have the StreamExt functions. [tokio_stream](https://docs.rs/tokio-stream/0.1.15/tokio_stream/index.html)::[wrappers](https://docs.rs/tokio-stream/0.1.15/tokio_stream/wrappers/index.html)::[LinesStream](https://docs.rs/tokio-stream/0.1.15/tokio_stream/wrappers/struct.LinesStream.html#) might let us do it that way.

one/src/migrations.rs
AaronGoldman: I will likely be nearly stable on any one disk as the tree in the filesystem try's to not change much. If the user wants a guaranty that they cover the entire ipfs repo then they should use make a file list file.
I wander if we should make a CLI argument for making a sorted file list file and a CLI for subtracting one from another

or maybe in the make file
```
# List files in directory at migration_1 and sort them alphabetically
ls -l ipfs/blocks | sort > migration_1.txt
 
CERAMIC_ONE_INPUT_FILE_LIST_PATH=migration_1.txt \
ceramic-one migrations from-ipfs 

# List files in directory at migration_2 and sort them alphabetically
ls -l ipfs/blocks | sort > migration_2.txt

# Find files in migration_2 that are not in migration_1
comm -23  migration_2.txt migration_1.txt > delta

CERAMIC_ONE_INPUT_FILE_LIST_PATH=delta.txt \
ceramic-one migrations from-ipfs 
```

nathanielc: Do we need the feature? I'd rather not add an unused feature that has a potential to break in mysterious ways.

one/src/migrations.rs
smrz2001: I figured it out.

smrz2001: > [tokio](https://docs.rs/tokio/latest/tokio/index.html)::[io](https://docs.rs/tokio/latest/tokio/io/index.html)::[Lines](https://docs.rs/tokio/latest/tokio/io/struct.Lines.html#) did not have the StreamExt functions. [tokio_stream](https://docs.rs/tokio-stream/0.1.15/tokio_stream/index.html)::[wrappers](https://docs.rs/tokio-stream/0.1.15/tokio_stream/wrappers/index.html)::[LinesStream](https://docs.rs/tokio-stream/0.1.15/tokio_stream/wrappers/struct.LinesStream.html#) might let us do it that way.

Oh wait, lol, that's what I ended up using.

one/src/migrations.rs
smrz2001: I took it out for now and added the check.

nathanielc (approved):

PR #502, opened 2024-08-23
feat: check cacao resources against stream information

dav1do:
Adds a method to the CACAO verifier trait to check that the provide resources are valid for the given stream. I may end up changing the parameters but these should all be known so I think this will work. Based on the [js-ceramic](https://github.com/ceramicnetwork/js-ceramic/blob/a342f204dfcc0952aa691249e17d09d53af35036/packages/stream-handler-common/src/signature_utils.ts#L77) code. Targets #498.

### DIFF
--- a/event/src/unvalidated/signed/cacao.rs
+++ b/event/src/unvalidated/signed/cacao.rs
@@ -49,31 +49,28 @@ impl From<chrono::DateTime<chrono::Utc>> for CapabilityTime {
     }
 }
 
-impl CapabilityTime {
-    /// Returns the time as a UTC DateTime
-    pub fn as_utc_dt(&self) -> anyhow::Result<chrono::DateTime<chrono::Utc>> {
-        let ts = Self::parse_timestamp(&self.0)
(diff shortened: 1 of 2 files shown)


event/src/unvalidated/signed/cacao.rs
```diff
@@ -37,6 +37,46 @@ pub struct Header {
...
+            .map_err(|e| anyhow::anyhow!("invalid time format for '{}': {}", self.0, e))?;
+        Ok(ts.to_utc())
+    }
+
+    /// Returns a string representation of the time
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+
+    fn parse_timestamp(input: &str) -> anyhow::Result<chrono::DateTime<chrono::Utc>> {
```
stbrody, line 65: do you think we should support try_into for DateTime to make this conversion easier/more native?

dav1do: yeah, probably nicer to do that. reworked in https://github.com/ceramicnetwork/rust-ceramic/pull/502/commits/7150a969eb4f97df82d0fdbed24f4582aec1667d

smrz2001 (approved):
LGTM!

PR #503, opened 2024-08-23
feat: validate signed events

dav1do:
Initial implementation of validation for signed events. The Validation module accepts the list of events, partitions them into "time" and "signed" events and delegates the validation to the appropriate handler. 

The signed validator will validate events against the init controller and if it's unable to find the init event, will consider them unvalidated. The service decides whether this is an error or acceptable based on the `ValidationRequirement` `allow_pending` value. If pending is supported, they are tracked in memory until the required init event is found, otherwise they are considered an error.

When verifying against the init event we check the CACAO resources, the CACAO and envelope signatures and the delegation chain, but we currently assume all streams are MIDs when constructing the stream ID to use.

This targets ~~#505~~ #526 and is not enabled by default yet as it is behind an experimental feature flag: `--experimental-feature-flags event-validation`.

### DIFF
--- a/event-svc/src/event/service.rs
+++ b/event-svc/src/event/service.rs
@@ -110,9 +110,9 @@ impl EventService {
         Self::try_new(pool, false, true).await
     }
 
-    /// Currently, we track events that were ASAP deliverable but we don't know the init event.
-    /// Immediately deliverable require the init event existing so don't need tracking.
-    /// Lazy deliverable don't require the init event to exist for validation so don't need tracking.
+    /// Currently, we track events when the [`ValidationRequirement`] allows. Right now, this applies to
+    /// recon discovered events when the init event isn't yet known to the node due to out of order sync
+    /// but might apply to other cases in the future.
     fn track_pending(&self, pending: Vec<UnvalidatedEvent>) {
         let mut map = self.pending_writes.lock().unwrap();
         if map.len() + pending.len() >= PENDING_VALIDATION_QUEUE_DEPTH {
--- a/event-svc/src/event/service.rs
+++ b/event-svc/src/event/service.rs
@@ -223,7 +223,7 @@ impl EventService {
 
         let ValidatedEvents {
             valid,
-            unvalidated: pending,
+            unvalidated,
             invalid,
         } = EventValidator::validate_events(&self.pool, validation_requirement, to_validate)
             .await?;
@@ -231,7 +231,7 @@ impl EventService {
 
         Ok(ValidatedEvents {
             valid,
-            unvalidated: pending,
+            unvalidated,
             invalid: invalid_events,
         })
     }
--- a/event-svc/src/event/service.rs
+++ b/event-svc/src/event/service.rs
@@ -437,7 +437,7 @@ impl EventService {
                 self.send_discovered_event(DiscoveredEvent {
                     cid: *ev.cid(),
                     prev: ev.event().prev().copied(),
-                    id: Some(*ev.event().id()),
+                    id: *ev.event().id(),
                     known_deliverable: ev.deliverable(),
                 })
                 .await?;
@@ -488,18 +488,9 @@ pub(crate) struct DiscoveredEvent {
     /// The prev event that this event builds on.
     pub(crate) prev: Option<Cid>,
     /// The Cid of the init event that identifies the stream this event belongs to.
-    pub(crate) id: Option<Cid>,
+    pub(crate) id: Cid,
     /// Whether this event is known to already be deliverable.
     pub(crate) known_deliverable: bool,
-}
-
-impl DiscoveredEvent {
-    pub(crate) fn stream_cid(&self) -> Cid {
-        match self.id {
-            None => self.cid, // init event
-            Some(id) => id,
-        }
-    }
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
--- a/event-svc/src/event/service.rs
+++ b/event-svc/src/event/service.rs
@@ -197,7 +197,7 @@ impl EventService {
     async fn validate_events(
         &self,
         items: &[ReconItem<EventId>],
-        validation_req: Option<ValidationRequirement>,
+        validation_req: Option<&ValidationRequirement>,
     ) -> Result<ValidatedEvents> {
         let mut parsed_events = Vec::with_capacity(items.len());
         let mut invalid_events = Vec::with_capacity(items.len() / 4); // assume most of the events are valid
@@ -245,17 +245,31 @@ impl EventService {
     ) -> Result<InsertResult> {
         let ValidatedEvents {
             valid,
-            unvalidated: pending,
+            unvalidated,
             mut invalid,
-        } = self.validate_events(items, validation_req).await?;
+        } = self.validate_events(items, validation_req.as_ref()).await?;
 
         let to_insert: Vec<EventInsertable> = valid
             .into_iter()
             .map(|e| ValidatedEvent::into_insertable(e, informant))
             .collect();
 
+        // we consider these "invalid" to tell the caller what happened.
+        // it may be okay and we'll track them below and try to find events we need in the future,
+        // but if this isn't allowed, we won't track them and just error
+        unvalidated.iter().for_each(|p| {
+            invalid.push(ValidationError::RequiresHistory {
+                key: p.order_key().to_owned(),
+            })
+        });
+
+        // if we're not validating, there's no need to pend anything
+        if validation_req.map_or(false, |v| v.allow_pending) {
+            self.track_pending(unvalidated);
+        }
+
         let store_result = self
-            .persist_events(to_insert, deliverable_req, pending, &mut invalid)
+            .persist_events(to_insert, deliverable_req, &mut invalid)
             .await?;
 
         Ok(InsertResult {
@@ -264,27 +278,15 @@ impl EventService {
         })
     }
 
-    /// Persists events to disk, tracks 'pending' events and notifies the background ordering task when appropriate
+    /// Persists events to disk and notifies the background ordering task when appropriate
     async fn persist_events(
         &self,
         to_insert: Vec<EventInsertable>,
         deliverable_req: DeliverableRequirement,
-        pending: Vec<UnvalidatedEvent>,
         invalid: &mut Vec<ValidationError>,
     ) -> Result<crate::store::InsertResult> {
         let ordered = OrderEvents::try_new(&self.pool, to_insert).await?;
 
-        // we consider pending "invalid" to tell the caller what happened.
-        // for recon, it's okay and we'll track them below and try to find events we need in the future,
-        // but if this is API (Immediate), we won't track them because they shouldn't exist yet.
-        pending.iter().for_each(|p| {
-            invalid.push(ValidationError::RequiresHistory {
-                key: p.order_key().to_owned(),
-            })
-        });
-
-        // api writes shouldn't have any missed history so we don't insert those events and
-        // we can skip notifying the ordering task because it's impossible to be waiting on them
         match deliverable_req {
             DeliverableRequirement::Immediate => {
                 let to_insert = ordered.deliverable().iter();
@@ -296,8 +298,6 @@ impl EventService {
                 Ok(CeramicOneEvent::insert_many(&self.pool, to_insert).await?)
             }
             DeliverableRequirement::Lazy | DeliverableRequirement::Asap => {
-                self.track_pending(pending);
-
                 let to_insert = ordered
                     .deliverable()
                     .iter()
@@ -496,10 +496,11 @@ pub(crate) struct DiscoveredEvent {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct ValidationRequirement {
     /// Whether we should check the signature is currently valid or simply whether it was once valid
-    check_exp: bool,
-    /// If true: the init event must be currently known to the node or it's invalid.
-    /// If false, it may be "pended" until the init event is discovered and the signature can be validated.
-    require_local_init: bool,
+    pub check_exp: bool,
+    /// Whether validation must succeed or events can be "pended" until init events are discovered.
+    /// If true: the init event may not yet be known to the node and we'll store it in memory until we discover it.
+    /// If false, we fail validation if we can't find the init event to validate the signature.
+    pub allow_pending: bool,
 }
 
 impl ValidationRequirement {
@@ -507,7 +508,7 @@ impl ValidationRequirement {
     pub fn new_local() -> Self {
         Self {
             check_exp: true,
-            require_local_init: true,
+            allow_pending: false,
         }
     }
 
@@ -515,7 +516,7 @@ impl ValidationRequirement {
     pub fn new_recon() -> Self {
         Self {
             check_exp: false,
-            require_local_init: false,
+            allow_pending: true,
         }
     }
 }
--- a/event-svc/src/event/validator/event.rs
+++ b/event-svc/src/event/validator/event.rs
@@ -15,7 +15,7 @@ use crate::{
     Result,
 };
 
-use super::grouped::{GroupedEvents, SignedEvents, Time, Unsigned};
+use super::grouped::{GroupedEvents, SignedValidationBatch, TimeValidationBatch};
 
 #[derive(Debug)]
 pub struct ValidatedEvents {
@@ -152,8 +152,8 @@ impl<'a> EventValidator<'a> {
         let grouped = GroupedEvents::from(parsed_events);
 
         let (validated_signed, validated_time) = try_join!(
-            validator.validate_signed_events(grouped.signed, grouped.unsigned),
-            validator.validate_time_events(grouped.time)
+            validator.validate_signed_events(grouped.signed_batch),
+            validator.validate_time_events(grouped.time_batch)
         )?;
         validated.extend_with(validated_signed);
         validated.extend_with(validated_time);
@@ -166,8 +166,7 @@ impl<'a> EventValidator<'a> {
 
     async fn validate_signed_events(
         &self,
-        events: SignedEvents,
-        unsigned: Vec<Unsigned>,
+        events: SignedValidationBatch,
     ) -> Result<ValidatedEvents> {
         let opts = if self.check_exp {
             ceramic_validation::VerifyJwsOpts::default()
@@ -177,13 +176,13 @@ impl<'a> EventValidator<'a> {
                 ..Default::default()
             }
         };
-        SignedValidator::validate_events(self.pool, &opts, events, unsigned).await
+        SignedValidator::validate_events(self.pool, &opts, events).await
     }
 
-    async fn validate_time_events(&self, events: Vec<Time>) -> Result<ValidatedEvents> {
+    async fn validate_time_events(&self, events: TimeValidationBatch) -> Result<ValidatedEvents> {
         // TODO: IMPLEMENT THIS
         Ok(ValidatedEvents {
-            valid: events.into_iter().map(ValidatedEvent::from).collect(),
+            valid: events.0.into_iter().map(ValidatedEvent::from).collect(),
             unvalidated: Vec::new(),
             invalid: Vec::new(),
         })
--- a/event-svc/src/event/validator/event.rs
+++ b/event-svc/src/event/validator/event.rs
@@ -4,12 +4,18 @@ use ceramic_core::{Cid, EventId, NodeId};
 use ceramic_event::unvalidated;
 use ipld_core::ipld::Ipld;
 use recon::ReconItem;
+use tokio::try_join;
 
 use crate::{
-    event::service::{ValidationError, ValidationRequirement},
+    event::{
+        service::{ValidationError, ValidationRequirement},
+        validator::signed::SignedValidator,
+    },
     store::{EventInsertable, SqlitePool},
     Result,
 };
+
+use super::grouped::{GroupedEvents, SignedEvents, Time, Unsigned};
 
 #[derive(Debug)]
 pub struct ValidatedEvents {
@@ -30,18 +36,16 @@ pub struct ValidatedEvent {
 }
 
 impl ValidatedEvent {
-    pub fn order_key(&self) -> &EventId {
-        &self.key
-    }
-
+    /// Convert this ValidatedEvent into an EventInsertable
     pub fn into_insertable(value: Self, informant: Option<NodeId>) -> EventInsertable {
         EventInsertable::new(value.key, value.cid, value.event, informant, false)
             .expect("validated events must be insertable")
     }
 
-    /// Skip the validation process. `unchecked` has a "memory unsafety" implication typically, but
-    /// this is safe code, however, doing this is not protocol safe. Mainly used in tests and by anything
-    /// that skips the validation process (e.g. an ipfs -> ceramic-one migration).
+    /// The way to convert into a validated event. Normally `unchecked` has a "memory unsafety" implication typically,
+    /// but this is safe code, however, doing this outside of the validation flow is not protocol safe.
+    /// It is used to covert an event after validation succeeds currently, but is also used in tests and
+    /// by anything that skips the validation process (e.g. an ipfs -> ceramic-one migration).
     pub(crate) fn from_unvalidated_unchecked(event: UnvalidatedEvent) -> Self {
         Self {
             key: event.key,
@@ -112,25 +116,74 @@ impl ValidatedEvents {
 pub struct EventValidator<'a> {
     pool: &'a SqlitePool,
     /// Whether we should check the signature is currently valid or simply whether it was once valid
-    requirement: ValidationRequirement,
+    check_exp: bool,
 }
 
 impl<'a> EventValidator<'a> {
-    fn new(pool: &'a SqlitePool, requirement: ValidationRequirement) -> Self {
-        Self { pool, requirement }
+    fn new(pool: &'a SqlitePool, requirement: &ValidationRequirement) -> Self {
+        Self {
+            pool,
+            check_exp: requirement.check_exp,
+        }
     }
+
     pub(crate) async fn validate_events(
-        _pool: &'a SqlitePool,
-        _validation_req: Option<ValidationRequirement>,
+        pool: &'a SqlitePool,
+        validation_req: Option<&ValidationRequirement>,
         parsed_events: Vec<UnvalidatedEvent>,
     ) -> Result<ValidatedEvents> {
-        // let _validator = Self::new(pool, validation_req);
+        let validation_req = if let Some(req) = validation_req {
+            req
+        } else {
+            // we don't validate so we just return done
+            return Ok(ValidatedEvents {
+                valid: parsed_events
+                    .into_iter()
+                    .map(ValidatedEvent::from_unvalidated_unchecked)
+                    .collect(),
+                unvalidated: Vec::new(),
+                invalid: Vec::new(),
+            });
+        };
+        let validator = Self::new(pool, validation_req);
+
+        let mut validated = ValidatedEvents::new_with_expected_valid(parsed_events.len());
+        // partition the events by type of validation needed and delegate to validators
+        let grouped = GroupedEvents::from(parsed_events);
+
+        let (validated_signed, validated_time) = try_join!(
+            validator.validate_signed_events(grouped.signed, grouped.unsigned),
+            validator.validate_time_events(grouped.time)
+        )?;
+        validated.extend_with(validated_signed);
+        validated.extend_with(validated_time);
+
+        if !validated.invalid.is_empty() {
+            tracing::warn!(count=%validated.invalid.len(), "invalid events discovered");
+        }
+        Ok(validated)
+    }
+
+    async fn validate_signed_events(
+        &self,
+        events: SignedEvents,
+        unsigned: Vec<Unsigned>,
+    ) -> Result<ValidatedEvents> {
+        let opts = if self.check_exp {
+            ceramic_validation::VerifyJwsOpts::default()
+        } else {
+            ceramic_validation::VerifyJwsOpts {
+                at_time: ceramic_validation::AtTime::SkipTimeChecks,
+                ..Default::default()
+            }
+        };
+        SignedValidator::validate_events(self.pool, &opts, events, unsigned).await
+    }
+
+    async fn validate_time_events(&self, events: Vec<Time>) -> Result<ValidatedEvents> {
         // TODO: IMPLEMENT THIS
         Ok(ValidatedEvents {
-            valid: parsed_events
-                .into_iter()
-                .map(ValidatedEvent::from_unvalidated_unchecked)
-                .collect(),
+            valid: events.into_iter().map(ValidatedEvent::from).collect(),
             unvalidated: Vec::new(),
             invalid: Vec::new(),
         })
--- a/event-svc/src/event/validator/event.rs
+++ b/event-svc/src/event/validator/event.rs
@@ -127,6 +127,8 @@ impl<'a> EventValidator<'a> {
         }
     }
 
+    /// Validates the events with the given validation requirement
+    /// If the [`ValidationRequirement`] is None, it just returns every event as valid
     pub(crate) async fn validate_events(
         pool: &'a SqlitePool,
         validation_req: Option<&ValidationRequirement>,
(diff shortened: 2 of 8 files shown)


service/src/event/service.rs
```diff
@@ -130,76 +177,35 @@ impl CeramicEventService {
...
-    }
-
-    async fn validate_time_events(
-        events: Vec<EventInsertable>,
-    ) -> Result<(Vec<EventInsertable>, Vec<InvalidItem>)> {
-        // TODO: IMPLEMENT THIS
-        Ok((events, Vec::new()))
-    }
-
-    pub(crate) async fn validate_events(
```
stbrody, line 147: any reason to remove this helper function?  `insert_events` is quite long, having it broken down into a series of helper functions I think improves readability.

service/src/event/service.rs
```diff
@@ -130,76 +177,35 @@ impl CeramicEventService {
...
-    ) -> Result<InsertResult> {
-        let (to_insert, mut invalid) = Self::validate_events(items).await?;
+        let history_req = source.into();
+        let pending_to_insert = self.remove_pending(&parsed_events);
+        let to_validate = parsed_events.into_iter().chain(pending_to_insert).collect();
+
+        let validator = EventValidator::new(&self.pool, history_req);
+        let (to_insert, pending) = validator.validate_events(to_validate, &mut invalid).await?;
+        pending.iter().for_each(|p| {
+            invalid.push(InvalidItem::RequiresHistory {
```
stbrody, line 204: doesn't whether pending events are considered invalid or not depend on the `DeliverableRequirement` in use?  It looks like below we are still inserting events with missing history so long as we're in Lazy or Asap mode.

service/src/event/validator.rs
```diff
@@ -0,0 +1,269 @@
...
+impl<'a> EventValidator<'a> {
+    pub fn new(pool: &'a SqlitePool, history_req: HistoryRequirement) -> Self {
+        Self { pool, history_req }
+    }
+
+    pub(crate) async fn validate_events(
+        &self,
+        parsed_events: Vec<EventInsertable>,
+        invalid_events: &mut Vec<InvalidItem>,
+    ) -> Result<(Vec<EventInsertable>, Vec<EventInsertable>)> {
```
stbrody, line 47: can we comment what the two different return vectors are?  And frankly just add some doc comments for this whole function

service/src/event/service.rs
```diff
@@ -74,9 +80,50 @@ impl CeramicEventService {
...
+                    Some(*e.cid())
+                } else {
+                    None
+                }
+            })
+            .collect();
+        let mut lock = self.pending_writes.lock().unwrap();
+        let mut to_add = Vec::new(); // no clue on capacity..
+        for new_cid in new_init_cids {
+            if let Some(unblocked) = lock.remove(&new_cid) {
```
stbrody, line 119: nitpick: weird to see `lock.remove()`.  At this point the "lock" object is really a reference to the underlying map right?  Can we call it "map" or something like that instead?

service/src/event/validator.rs
```diff
@@ -0,0 +1,269 @@
...
+        // first collect all the init events so we can refer to them when grouping by type
+        // in order to avoid querying the database or rejecting the events incorrectly.
+        // we will then add any missing init events from the DB for the next phase.
+        // guess at a capacity and just assume half are init events so we only need to re-allocate once at most
+        // TODO: we should validate these init events before we trust them for others.. could
```
stbrody, line 52: what's the plan for addressing this TODO?  Do we need a ticket for this?

service/src/event/validator.rs
```diff
@@ -0,0 +1,269 @@
...
+            if event.event().is_init() {
+                required_init_events.insert(*event.cid(), event.event().to_owned());
+            }
+        });
+        // Group events by their type (over allocate vectors for now)
+        let mut valid_events = Vec::with_capacity(parsed_events.len());
+        let mut signed_events = Vec::with_capacity(parsed_events.len());
+        let mut time_events = Vec::with_capacity(parsed_events.len());
+        let mut pending_events = match self.history_req {
+            HistoryRequirement::InitOnly => Vec::with_capacity(parsed_events.len() / 2),
```
stbrody, line 66: I don't quite follow the logic for different starting capacity based on history requirement here.

service/src/event/validator.rs
```diff
@@ -0,0 +1,269 @@
...
+            _ => Vec::new(),
+        };
+        for event in parsed_events {
+            match event.event().as_ref() {
+                unvalidated::Event::Time(time) => {
```
stbrody, line 71: why do time events even need their init event to be present to be validated?  What stops you from validating a TimeEvent even without having the init event?  For data events you need the init because you need the controller, but TimeEvents don't have that requirement.

service/src/event/validator.rs
```diff
@@ -0,0 +1,269 @@
...
+                    let init_known = if required_init_events.contains_key(time.id()) {
+                        true
+                    } else {
+                        let (exists, _) =
+                            CeramicOneEvent::deliverable_by_cid(self.pool, time.id()).await?;
+                        exists
+                    };
+                    if init_known || matches!(self.history_req, HistoryRequirement::None) {
```
stbrody, line 79: if there's no history requirement, then it seems like we can skip querying the database for the init event entirely yes?  Maybe we can restructure this with an early return to avoid doing unnecessary work?

service/src/event/validator.rs
```diff
@@ -0,0 +1,269 @@
...
+                        time_events.push(event);
+                    } else if matches!(self.history_req, HistoryRequirement::InitOnly) {
+                        pending_events.push(event);
+                    } else {
+                        invalid_events.push(InvalidItem::RequiresHistory {
```
stbrody, line 84: this feels like more of a IOD concern than a validation concern.  I feel like all the validation code shouldn't need to know anything about the HistoryRequirement frankly.  From validation's perspective it either has everything it needs to do the validation, or it doesn't and is still missing something.  But whether the fact that missing something is an error or not is a question for a higher layer, not something the validation system should be concerning itself with.

stbrody: Maybe validation should return 3 vectors?  valid, invalid, and pending?  Then the caller can decide if it wants to turn pending into invalid.

stbrody (commented):
I'm worried we're leaking too much IOD logic into the validation layer.  Can we keep these more isolated?

service/src/event/validator.rs
dav1do: It's silly and I'll remove it - not worth the mental overhead. The reasoning was that no history required or all history required were either valid or invalid, not pending. 

I'd like to figure out how to encapsulate the pending stuff more nicely because it sort of bleeds into IOD and makes assumptions based on the source that are more implicit than I'd like right now.

service/src/event/validator.rs
dav1do: I was on the fence here.. we don't need it, so maybe it's better to just validate it right away. I assumed if we don't have the init event we probably don't care to validate the event because we're not going to be able to order it and have no reason to store it.. but if we found it, we probably want to, so it'd be nice to simplify this to just data events.

service/src/event/validator.rs
dav1do: yeah, good call

service/src/event/validator.rs
dav1do: Yeah, I need to encapsulate this better and think about it a bit more. Right now it does return 3 vectors (sort of): `valid`, `pending` return by value and it access the mutable reference to invalid to add to it. I'm basically trying to avoid doing signature validation until we care/need to. If I can make a type that encapsulates the phases of validation, it'll be store the partially validated event and pick up where we need to.

The reason this exists is pretty much because of the migration. We can't validate migrated data events completely without the init events, but we're currently not willing to wait for them because it's already so time consuming and IO intensive and we get random ordering, so we want to validate signatures etc without checking things against the init, so at least we have some confidence the data was made by js-ceramic and is valid.

service/src/event/validator.rs
dav1do: yeah, I was planning to use a struct return here. I first wanted to poc things and makes sure the validation traits etc seem reasonable and it should actually work.. but this is definitely a hack job 😅  

I'll work on making this more reasonable today, if you have ideas on how to represent the events without needing to match and unreachable!() time/init, that'd be appreciated.. not sure if we need to add something that consumes the unvalidated event, or put arcs around some fields.. right now we get references and can't move/clone things easily.

service/src/event/service.rs
dav1do: yea, for sure. this is a weird shorthand I use around mutex guards for some reason.

service/src/event/service.rs
dav1do: I basically just moved it to the other file. It's currently not a static method and is `let validator = EventValidator::new(..); validator.validate_events(..).await;`. I'm going to rework things today and see how it shakes out - I didn't want to have this function that just called another function immediately and thought having a dedicated file might be nice since this one is getting long.

Maybe it makes sense to add a validator trait for this crate and inject the validator from recon/api/migration that composes the pieces rather than trying to have a smart validator that knows all of the nuances based on history requirements etc.

service/src/event/validator.rs
dav1do: maybe it should be a ticket.. I was thinking to just do things in two passes initially: split init events out and check their signatures before using them in the next phase.. it's a bit of a bummer but isn't hard, I just didn't want to forget so I wrote this TODO.

service/src/event/service.rs
dav1do: Yeah, this is weird and needs adjustment. The deliverable requirement is how I got the "history requirement" enum.. I wanted something else but it may not be different enough to be useful. Right now, we turn these into invalid items and the into API/Recon result turns them into an error/pending count increment.

Pending is only supposed to ever happen for recon events, we don't want to pend migration data (just get it over with asap) and API events should have everything on the node and be rejected. Pending is different than missing history for IOD but I need to figure out how to use/share/encapsulate things across validation and IOD more nicely. IOD requires complete history, while data event validation requires only the init event, which we expect to find _soon_ if we don't already have it, so `pending`. But if we have no init event, no need to try ordering. 

Right now, I have a bit of a mix of doing the same work in both places and not knowing anything about the other or state of the event. I think tracking the validation process as a new type (e.g. including the init cid when it's used), might make it easier to reason about and be useful for the IOD stuff too.

service/src/event/validator/signed.rs
```diff
@@ -0,0 +1,252 @@
...
+                unvalidated::Payload::Init(init) => {
+                    self.verify_signature_against_init(init, signed, opts).await
+                }
+            },
+            Some(unvalidated::Event::Unsigned(init)) => {
+                self.verify_signature_against_init(init, signed, opts).await
+            }
+            None => {
+                // TODO: we skip validating the CACAO resources without the init because we don't even know
+                // what stream we're on, so it will fail if we don't have * resources access and the data
```
dav1do, line 204: this should actually say what model (sepKey/sepValue) rather than stream.. we can assume our `id` CID is the real stream ID but we may not be able to pass the time checks so I still think we should skip checking the cacao resources

event-svc/src/event/validator/signed.rs
```diff
@@ -0,0 +1,265 @@
...
+                self.add_invalid(InvalidItem::InvalidSignature {
+                    key: event.key,
+                    reason,
+                })
+            }
+            ValidationStatus::Pending => self.add_pending(event),
+        }
+    }
+
+    /// Returns valid = Ok(true), pending = Ok(false), invalid = Err(Reason)
```
dav1do, line 137: fix doc comment as it's out of date with an enum return

event-svc/src/event/validator/signed.rs
```diff
@@ -0,0 +1,265 @@
...
+                unvalidated::Payload::Init(init) => {
+                    self.verify_signature_against_init(init, signed, opts).await
+                }
+            },
+            Some(unvalidated::Event::Unsigned(init)) => {
+                self.verify_signature_against_init(init, signed, opts).await
+            }
+            None => {
+                // TODO: we skip validating the CACAO resources without the init because we don't even know
+                // what stream we're on, so it will fail if we don't have * resources access and the data
```
dav1do, line 217: this comment is wrong.. we know what the stream should be but not the model

event-svc/src/tests/mod.rs
```diff
@@ -61,15 +61,9 @@ pub(crate) struct TestEventInfo {
...
-    let controller = thread_rng()
-        .sample_iter(&rand::distributions::Alphanumeric)
-        .take(32)
-        .map(char::from)
-        .collect::<String>();
-
     let unique = gen_rand_bytes::<12>();
     let init = ceramic_event::unvalidated::Builder::init()
-        .with_controller(controller)
+        .with_controller(CONTROLLER.to_owned())
```
dav1do, line 66: tests were failing because a random controller isn't valid and events were no longer valid

service/src/event/validator.rs
dav1do: added in this PR

event-svc/src/event/validator/signed.rs
```diff
@@ -0,0 +1,265 @@
...
+                    )
+                })?;
+                let model_id = match StreamId::try_from(init.header().model()) {
+                    Ok(m) => Some(m),
+                    Err(e) => {
+                        warn!(error=?e, %init_cid, "failed to parse init event model field as stream");
+                        None
+                    }
+                };
+                // TODO: this assumes all streams are MIDs
```
stbrody, line 250: can we add a ticket number for this TODO?

stbrody (commented):
sorry for so many comments, this PR is really massive.  I feel like maybe if we pulled out the implementation of the validator crate into its own PR it would be a bit more manageable?

event-svc/src/event/validator/signed.rs
```diff
@@ -0,0 +1,265 @@
...
+
+use super::{
+    event::{ValidatedEvent, ValidatedEvents},
+    grouped::SignedEvents,
+    UnvalidatedEvent,
+};
+
+#[derive(Debug)]
+pub struct SignedValidator {
+    init_map: HashMap<Cid, Arc<unvalidated::Event<Ipld>>>,
```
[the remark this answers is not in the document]
dav1do: this is created and dropped for a given "validation" call right now. we're only adding the init events for data events in the batch and when we're done with validation, they are all dropped in the insert function

event-svc/src/event/validator/signed.rs
```diff
@@ -0,0 +1,265 @@
...
+                    None
+                };
+
+                match self.verify_signature(init, signed, opts).await {
+                    Ok(_) => Ok(ValidationStatus::ValidData),
+                    Err(err) => Ok(ValidationStatus::InvalidData(err)),
+                }
+            }
+            unvalidated::Payload::Init(init) => {
+                // yes, we are the init in this case but it still works
```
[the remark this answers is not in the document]
dav1do: will clarify.. it was mostly cheek. we call `verify_signature_against_init` where init and payload are the same thing.

event-svc/src/event/validator/signed.rs
```diff
@@ -0,0 +1,265 @@
...
+                    Err("init event pointer cannot reference a data event".to_string())
+                }
+                unvalidated::Payload::Init(init) => {
+                    self.verify_signature_against_init(init, signed, opts).await
+                }
+            },
+            Some(unvalidated::Event::Unsigned(init)) => {
+                self.verify_signature_against_init(init, signed, opts).await
+            }
+            None => {
```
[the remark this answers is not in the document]
dav1do: migrations (when require_init is false) we don't check the controller info against the init (currently)

event-svc/src/event/validator/signed.rs
```diff
@@ -0,0 +1,265 @@
...
+                // should have been valid previously and is just being migrated, so we don't want to unnecessarily fail
+                signed
+                    .verify_signature(None, opts)
+                    .await
+                    .map_err(|e| e.to_string())
+            }
+        }
+    }
+
+    async fn verify_signature_against_init(
```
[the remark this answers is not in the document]
dav1do: yeah, will do.. I already confused myself reading these names

event-svc/src/event/validator/grouped.rs
```diff
@@ -0,0 +1,125 @@
...
+            unvalidated::Event::Signed(signed) => match signed.payload() {
+                unvalidated::Payload::Data(_) => Self::SignedData(SignedData(value)),
+                unvalidated::Payload::Init(_) => Self::SignedInit(SignedInit(value)),
+            },
+            unvalidated::Event::Unsigned(_) => Self::None(Unsigned(value)),
+        }
+    }
+}
+
+macro_rules! unvalidated_wrapper {
```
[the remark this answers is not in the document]
dav1do: but this is "simple macros" ;) i'll try to clarify this file, it was supposed to be magic/internal helpers to remove the need to match on event types and make it impossible to represent them wrong. That is, instead of adding `unvalidated::Event<D>`  that we know is a time event and then having to do the match and panic on the other branches, this takes care of hiding that so when we build it and it declares itself a time event, we can just call `as_time()` and get what we want.

event-svc/src/event/validator/grouped.rs
```diff
@@ -0,0 +1,125 @@
...
+pub struct GroupedEvents {
+    pub time_events: Vec<Time>,
+    pub signed_events: SignedEvents,
+}
+
+#[derive(Debug)]
+pub struct SignedEvents {
+    pub data: Vec<SignedData>,
+    pub init: Vec<SignedInit>,
+    pub unsigned: Vec<Unsigned>,
```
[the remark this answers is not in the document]
dav1do: yeah, we could but we'd need to pass it to the signed event validator still. maybe the name isn't great, but it's intended to partition everything into the two validation buckets, and event signatures need all the init events (signed and unsigned) since they aren't in the database yet to look up

event-svc/src/event/validator/event.rs
```diff
@@ -0,0 +1,187 @@
...
+        }
+        Ok(validated)
+    }
+
+    async fn validate_signed_events(&self, events: SignedEvents) -> Result<ValidatedEvents> {
+        let opts = if self.check_exp {
+            ceramic_validation::VerifyJwsOpts::default()
+        } else {
+            ceramic_validation::VerifyJwsOpts {
+                at_time: ceramic_validation::AtTime::SkipTimeChecks,
```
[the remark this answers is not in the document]
dav1do: I'm not sure.. do we want to throw out events discovered on the network that have expired cacaos or allow them and then let the conclusion be that they're expired? I think I'd prefer to ignore time information from recon events because it will add an entire new level of ordering to validation that might be a more natural fit for aggregation/conclusions.

Also for did:key, we don't want to check if the DID is expired so I think skip (or some variation on it) will probably make sense for recon discovered data.

event-svc/src/event/validator/event.rs
```diff
@@ -0,0 +1,187 @@
...
+}
+
+#[derive(Debug)]
+pub struct EventValidator<'a> {
+    pool: &'a SqlitePool,
+    /// Whether to require the init event exists, and if we should check the the signature against
+    /// the init controller. True in all cases except a data migration.
+    require_init_event: bool,
+    /// Whether we should check the signature is currently valid or simply whether it was once valid
+    check_exp: bool,
```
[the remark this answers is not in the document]
dav1do: for recon we might get events that are old and we haven't yet decided to tie rpc time stamping in as a blocker for validating the event signatures (more below)... the API uses the wallclock time, but recon and migrations set this to false currently

event-svc/src/event/validator/event.rs
```diff
@@ -0,0 +1,187 @@
...
+
+impl From<ValidatedEvent> for EventInsertable {
+    fn from(value: ValidatedEvent) -> Self {
+        EventInsertable::new(value.key, value.cid, value.event, false)
+            .expect("validated events must be insertable")
+    }
+}
+
+#[derive(Debug)]
+pub struct UnvalidatedEvent {
```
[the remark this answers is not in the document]
dav1do: events can't really track their event ID can they? I mean, I guess we can assign one based on the network and state of the system but it's not part of the event so it feels a bit weird. I agree this name is weird, was just trying to wrap things together in something for the crate to use rather than have `(eventId, event)` tuples like we had all over. but with the CID stuff being known now, maybe this can be removed. I'll see what I can do.

event-svc/src/event/service.rs
```diff
@@ -141,44 +193,15 @@ impl EventService {
...
-        // TODO: IMPLEMENT THIS
-        Ok((events, Vec::new()))
-    }
-
-    pub(crate) async fn validate_events(
+    pub(crate) async fn insert_events(
+        &self,
         items: &[ReconItem<EventId>],
-    ) -> Result<(Vec<EventInsertable>, Vec<InvalidItem>)> {
+        source: DeliverableRequirement,
```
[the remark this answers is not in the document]
dav1do: yeah, it's not right.. iI called it source originally because it was like API/recon and then we made the deliverable enum instead and I didn't rename the variable. but it's simple to change as it's just internal to the function.

event-svc/src/event/service.rs
```diff
@@ -187,47 +210,35 @@ impl EventService {
...
-        Ok((valid_events, invalid_events))
-    }
-
-    pub(crate) async fn insert_events(
-        &self,
-        items: &[ReconItem<EventId>],
-        source: DeliverableRequirement,
-    ) -> Result<InsertResult> {
-        let (to_insert, mut invalid) = Self::validate_events(items).await?;
+        let to_insert: Vec<EventInsertable> = validated.into_iter().map(|e| e.into()).collect();
```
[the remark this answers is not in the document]
dav1do: yeah, probably makes sense.. I have been trying to avoid pass through functions but I think this has gotten big enough now. I didn't expect parsing to happen before originally and thought everything was going to be in the validator file.

event-svc/src/event/validator/event.rs
```diff
@@ -0,0 +1,187 @@
...
+    pub valid: Vec<ValidatedEvent>,
+    /// We don't have enough information to validate these events yet.
+    /// e.g. The init event is required and the node has not yet discovered it.
+    pub pending: Vec<UnvalidatedEvent>,
+    /// Events that failed validation
+    pub invalid: Vec<InvalidItem>,
+}
+
+#[derive(Debug)]
+pub struct ValidatedEvent {
```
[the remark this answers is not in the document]
dav1do: I'm open to ideas.. I originally planned to go that route, but I couldn't really figure out how to define or transition it at the time.. I didn't want to make event dependent on the validator crate, allow constructing them manually, and wasn't sure if we needed "validated::event" as a core construct to be passed around between other crates. 

Like I could see providing something from validation that basically implements the verify_signature function from here that uses the existing traits that verify cacaos, event signatures and jws stuff in one place using the init event/controller.. but it seems like that is just tying the core concepts from that crate together into one place (which is currently the service) and doesn't need to be bigger than that.. I might be wrong or totally missing an elegant solution but I just went the route of making new types internal to this module for now.

event-svc/src/event/validator/signed.rs
```diff
@@ -0,0 +1,265 @@
...
+            events.data.len(),
+            require_init,
+        );
+        validator.init_map.extend(
+            events
+                .unsigned
+                .iter()
+                .map(|i| (i.as_inner().cid, i.as_inner().event.to_owned())),
+        );
+        validator.add_valid_iter(events.unsigned.into_iter().map(|i| i.into_inner().into()));
```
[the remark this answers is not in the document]
dav1do: talked about above.. but we need to know if any init events are in our batch but not in the db yet

dav1do:
> sorry for so many comments, this PR is really massive. I feel like maybe if we pulled out the implementation of the validator crate into its own PR it would be a bit more manageable?

I'll see what I can do in the morning.. the bulk of this PR is just the validator module but there is some behavior around pending and other changes that might be a distraction. The validator crate to do signature verification already exists, this module is mostly to group/find/orchestrate everything needed to be able to verify a envelope and cacao against the init controller and figure out if it's a hard error or a "maybe error".

event-svc/src/event/validator/event.rs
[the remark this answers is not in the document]
stbrody: yeah, I think this comes back to what we started talking about at the end of the standup call earlier today.  Recon still needs to validate expiration times, it can't just totally ignore them.  I do think the logic for the api and recon likely should be the same and unified, I'm not sure we need this boolean.  But really we need to figure out how CACAO enforcement with time information from anchors is going to work.

event-svc/src/event/validator/grouped.rs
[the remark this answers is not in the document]
stbrody: why do unsigned events need to go to the SignedEventValidator?  Can't they just skip validation entirely?

event-svc/src/event/validator/event.rs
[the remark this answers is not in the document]
stbrody: I still don't see why recon is any different from the API with regard to CACAO expiration.  Either the event is valid or it isn't, and the logic for determining if it's valid depends on timestamps from anchors.

Also I don't see how the existence of did:key signed events factors into this?

event-svc/src/event/validator/grouped.rs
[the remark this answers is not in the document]
stbrody: oh I see, they need to be passed in not to be validated, but as context for validating other events in the batch.  Okay makes sense, probably just a comment to that effect will help

event-svc/src/event/validator/signed.rs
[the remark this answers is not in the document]
stbrody: ah okay nice.  Then some comments should make that clear

event-svc/src/event/validator/signed.rs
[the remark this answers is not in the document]
stbrody: I'm starting to feel like migrations should maybe just bypass validation entirely so we can take out any migration-related special casing out of all of the validation code...

event-svc/src/event/validator/event.rs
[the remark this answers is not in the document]
dav1do: did:keys can also expire, be rotated or revoked, though the exact fields to check seem a bit dependent on your use case. We check `nextUpdate` in [js-did](https://github.com/ceramicnetwork/js-did/blob/3a8ad6d2c743b9b22c067961ae0e71818ff76443/packages/dids/src/did.ts#L353).

the rules should be the same, but the current flow is the problem.. one is happening in parallel (or in the future) and the other is being applied to the stream tip. also, if an event is invalid because a cacao is expired when we find it on the network, should we drop it or store it? in the future we *may* discover the time event that makes it valid, so drawing a conclusion that it's not valid as part of the stream makes sense, but rejecting it completely seems a bit excessive.

event-svc/src/event/validator/signed.rs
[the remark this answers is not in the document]
dav1do: yea, I'm coming to that conclusion too (I think I wrote something similar above in another comment).. it seems like using the service for migrations is weird when it literally just skips every piece so it can go straight to the database call

dav1do:
> > sorry for so many comments, this PR is really massive. I feel like maybe if we pulled out the implementation of the validator crate into its own PR it would be a bit more manageable?
> 
> I'll see what I can do in the morning.. the bulk of this PR is just the validator module but there is some behavior around pending and other changes that might be a distraction. The validator crate to do signature verification already exists, this module is mostly to group/find/orchestrate everything needed to be able to verify a envelope and cacao against the init controller and figure out if it's a hard error or a "maybe error".

Split some changes out in #526. Most of the feedback about clarifying comments/names were addressed in that PR. This PR actually does the signature validation. If it's easier to close and reopen a new one, I can do that. I still have the original branch/commits locally but after rebasing onto main and splitting things out, the original breakdown of changes wasn't super helpful so I squashed them and split them up differently.

event-svc/src/tests/mod.rs
dav1do: This is why the conclusion feed test expects were regenerated as well

event-svc/src/event/validator/signed.rs
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    fn add_valid_iter<I>(&mut self, item: I)
+    where
+        I: Iterator<Item = ValidatedEvent>,
+    {
+        self.validated.valid.extend(item);
+    }
+
+    pub async fn validate_events(
```
nathanielc, line 81: Can we add a test for this method? It appears to be the main logic added in this PR.

event-svc/src/event/validator/signed.rs
```diff
@@ -0,0 +1,267 @@
...
+        pool: &SqlitePool,
+        opts: &VerifyJwsOpts,
+        events: SignedEvents,
+        unsigned: Vec<Unsigned>,
```
nathanielc, line 85: nit, its confusing that we are passing in events based on their signed status. I get that its helpful for the algorithm to know which is which. Maybe the `grouped` module can have a `Group` type that contains three sets of events, data, init, and unsigned init? That way this method has only a single parameter for the input events. This should help clarify the relationship between the events.

nathanielc: Ah I found GroupedEvents in this PR. Why not use this type for the argument to this method?

nathanielc (commented):
Looks good, I had some code readability questions, but after spending some time with the code it makes sense. Once a test is added this should be good to go

event-svc/src/event/validator/signed.rs
dav1do: Yeah, probably worthwhile. I didn't add a new one because the existing tests all go through this logic (and started failing before updating the controller to a real one). I'll look at putting something dedicated in place.

event-svc/src/event/validator/signed.rs
dav1do: Yeah, I'll see what I can do. I originally just passed `SignedEvents` in, but Spencer (reasonably) found it confusing that SignedEvents contained unsigned init events, so I split it into three things (time, signed, unsigned) and passed two in here. I could probably take a reference to grouped, and maybe it's better since we _may_ want the time events to add timestamps someday. Currently those are being validated in parallel and I didn't want to fight over ownership but putting things in an Arc if needed is simple. Alternatively, I might be able to come up with a better name than `SignedEvents` and just pass that in here. 

The initial types/layout (like grouped) was split into #526, which this is targeting since this PR was getting big.

event-svc/src/event/validator/event.rs
```diff
@@ -0,0 +1,187 @@
...
+    key: EventId,
+    cid: Cid,
```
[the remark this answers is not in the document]
dav1do: They still don't know their CID without recalculating it so I'm going to leave it on here for now.

event-svc/src/event/validator/signed.rs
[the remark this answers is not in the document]
dav1do: Added some comments, hopefully it's better now.

event-svc/src/event/validator/grouped.rs
```diff
@@ -0,0 +1,125 @@
...
+        let mut grouped = GroupedEvents::new(value.len() / 2);
+
+        value
+            .into_iter()
+            .for_each(|v| grouped.add(ValidationNeeded::from(v)));
+        grouped
+    }
+}
+
+impl GroupedEvents {
```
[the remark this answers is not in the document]
dav1do: Added some more comments that hopefully clear things up a bit... if it's still confusing, I can just write out the types, these macros aren't essential, they just gets rid of some boilerplate.

event-svc/src/event/service.rs
```diff
@@ -33,13 +33,20 @@ const MAX_ITERATIONS: usize = 100_000_000;
 /// This is currently 304 bytes per event, so this is 3 MB of data
 const PENDING_EVENTS_CHANNEL_DEPTH: usize = 1_000_000;
 
+/// The max number of events that can be waiting on a previous event before we can validate them
```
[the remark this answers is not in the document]
dav1do: Added more information about the use and what happens.

event-svc/src/event/validator/signed.rs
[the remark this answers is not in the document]
dav1do: adjusted to simply skip doing any work when if we're not requiring validation so this path has been removed. we now just covert things to "valid" if the caller says, "I want this to be implied so give me back things so I can move on"

event-svc/src/event/validator/signed.rs
dav1do: added some tests at the layer the service calls into (`EventValidator::validate_events`)

event-svc/src/event/validator/signed.rs
dav1do: I wanted to keep ownership of the events to make it easy for now, so I [renamed the types](https://github.com/ceramicnetwork/rust-ceramic/pull/503/commits/89ae7c556a3655c068117c04e571e548013c425c) on `GroupedEvents` `SignedValidationBatch` and `TimeValidationBatch` and pass those in.. lmk what you think.

event-svc/src/event/service.rs
```diff
@@ -99,6 +110,47 @@ impl EventService {
         Self::try_new(pool, false, true).await
     }
 
+    fn track_pending(&self, pending: Vec<UnvalidatedEvent>) {
```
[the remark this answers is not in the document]
dav1do: I changed how the pending requirement is determined (must be passed as part of `ValidationRequirement` and updated the doc to match)

event-svc/src/event/service.rs
```diff
@@ -187,47 +210,35 @@ impl EventService {
...
-
-    pub(crate) async fn insert_events(
-        &self,
-        items: &[ReconItem<EventId>],
-        source: DeliverableRequirement,
-    ) -> Result<InsertResult> {
-        let (to_insert, mut invalid) = Self::validate_events(items).await?;
+        let to_insert: Vec<EventInsertable> = validated.into_iter().map(|e| e.into()).collect();
 
         let ordered = OrderEvents::try_new(&self.pool, to_insert).await?;
```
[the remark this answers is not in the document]
dav1do: it's been refactored and more like that now. validate, order, persist. The clarity about ordering has been refactored more in #530 that will come after.

event-svc/src/event/validator/signed.rs
dav1do: there are tests related to different signature validation success/failure (did, cacao, invalid) in the Validation crate so I didn't focus on that here

PR #504, opened 2024-08-23
fix: use cacao timestamp strings and parse

dav1do:
In order to ensure that CACAOs round trip without changing the CID, we use a string when deserializing rather than a chrono::DateTime which can change the second precision that was originally included. It's a bit clunky, open to better ideas.

### DIFF
--- a/event/src/unvalidated/signed/cacao.rs
+++ b/event/src/unvalidated/signed/cacao.rs
@@ -37,9 +37,6 @@ pub struct Header {
     pub r#type: HeaderType,
 }
 
-/// Time format for capability
-pub type CapabilityTime = chrono::DateTime<chrono::Utc>;
-
 /// Payload for a CACAO
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Payload {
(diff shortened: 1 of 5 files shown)


smrz2001 (approved):
Ty for the quick fix!! LGTM 🥳

event/src/unvalidated/signed/cacao.rs
```diff
@@ -51,20 +48,29 @@ pub struct Payload {
     pub domain: String,
 
     /// Expiration time
+    /// Not using a chrono::DateTime because we need to round trip the exact
+    /// value we receive without modifying precision and changning the CID.
+    /// The new CAIP proposes using ints but most of our cacaos still use ISO 8601 values.
     #[serde(rename = "exp", skip_serializing_if = "Option::is_none")]
-    pub expiration: Option<CapabilityTime>,
+    pub expiration: Option<String>,
```
stbrody, line 55: Instead of using a raw String and getting no type safety, would it make sense to introduce a new type, that uses string as the underlying storage but provides a method to convert to a DateTime?

dav1do: Yeah, it might be nice to use a tuple struct with serde transparent. Or at least make these fields private. I can add something like that in the cacao branch I have open.. wanted to fix it before it got released Monday to avoid any weird issues with event serialization.

dav1do: Made the change here: https://github.com/ceramicnetwork/rust-ceramic/pull/502/commits/1d602df9dde4bceaeb25b755d523211822a8b70f

PR #505, opened 2024-08-26
feat: groundwork for event validation and some clean up

dav1do:
Cleaned up some oversights like not adding the new crate to the workspace dependencies list and the benchmark not compiling. Also put an `Arc` around the event on `EventInsertable` so I can clone it to track in a map for the event validation work.

The main change is adding a new experimental feature flag that event validation is behind. It's a sub optimal representation, but my intention is that it's temporary and allows me to merge changes in #503 so we can benchmark and test more thoroughly before it's on everywhere, at which point the feature flag will be removed and it's the configured behavior. It's enabled with `--experimental-feature-flags event-validation`.

All of these changes were part of #503, so I'll rebase that one to use this as the source since these should be standalone and we can agree on this structure without getting into the weeds of the validation implementation. This should also allow us to merge imperfect changes and follow up rather than having to get everything right on the first attempt.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,6 +62,7 @@ ceramic-one = { path = "./one" }
 ceramic-p2p = { path = "./p2p" }
 ceramic-service = { path = "./service" }
 ceramic-store = { path = "./store" }
+ceramic-validation = { path = "./validation" }
 chrono = "0.4.31"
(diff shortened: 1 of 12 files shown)


stephhuynh18 (approved):
🚀

AaronGoldman (approved):

PR #508, opened 2024-08-29
feat: api to transform events to conclusion events

samika98:
API that processes a series of raw events, transforming them into a structured vector of conclusion events. The transformation involves taking a high watermark as input, fetching events from the database based on this marker, and converting these events into the ConclusionEvent type. 

Added transform_raw_events_to_conclusion_events function which converts events based on the provided high watermark
Modified event and payload structures
Updated CeramicEventService to include new transformation logic

TODO : Add time event creation logic for test


service/src/event/service.rs
```diff
@@ -237,6 +241,168 @@ impl CeramicEventService {
...
+
+                Ok(ConclusionEvent::Time(ConclusionTime {
+                    event_cid: raw_event.id,
+                    init: ConclusionInit {
+                        stream_cid: *init_event.unwrap().id().unwrap(),
+                        stream_type: 0, // You may need to determine the correct stream type
+                        controller: event.get_controller().unwrap(),
+                        dimensions: vec![], // You may need to populate this if needed
+                    },
+                    // TODO_Discuss: How can this ever have more than one value? prev return a single cid
```
samika98, line 298: prev returns a single cid

service/src/event/service.rs
```diff
@@ -237,6 +241,168 @@ impl CeramicEventService {
...
+                    // Soln_proposed : ?
+                    previous: vec![*time_event.prev()],
+                    //TODO_Discuss: How to determine an ever incrementing index.
+                    //Soln_proposed : We have to store this on disk and increment with every event seen ?
+                    //Define purpose of this counter and how it can be used
```
samika98, line 303: ** how it can be useful to the consumer of this api

service/src/event/service.rs
```diff
@@ -237,6 +241,168 @@ impl CeramicEventService {
...
+        unimplemented!("Implement get_event_by_cid method")
+    }
+
+    async fn conclusion_events_since(
+        &self,
+        highwater_mark: i64,
+        limit: i64,
+        include_data: ceramic_api::IncludeEventData,
+    ) -> anyhow::Result<Vec<ConclusionEvent>> {
+        let raw_events = ceramic_api::EventStore::events_since_highwater_mark(
```
nathanielc, line 394: We do not want to rely on the API methods here. We should likely move the body of `ceramic_api::EventStore::events_since_highwater_mark` to a method on the struct directly. Both this method and the EventStore implementation can call it. This way we do not overload the ceramic_api types.

service/src/event/service.rs
```diff
@@ -237,6 +241,168 @@ impl CeramicEventService {
...
+    fn get_event_by_cid(&self, cid: &Cid) -> Result<ceramic_event::unvalidated::Event<Ipld>> {
+        // Implement this method to fetch the event from your storage
+        unimplemented!("Implement get_event_by_cid method")
+    }
+
+    async fn conclusion_events_since(
+        &self,
+        highwater_mark: i64,
+        limit: i64,
+        include_data: ceramic_api::IncludeEventData,
```
nathanielc, line 392: We always want the data for the conclusion API so we should not let this be a parameter on the function.

service/src/event/service.rs
```diff
@@ -237,6 +241,168 @@ impl CeramicEventService {
         })
     }
 
+    // We map over this to get a vec of conclusion events
+    fn transform_raw_events_to_conclusion_events(
+        &self,
+        raw_event: ceramic_api::EventDataResult,
```
nathanielc, line 247: As noted, we should not use the ceramic_api types here but instead use the `unvalidated::Event` directly.

service/src/event/service.rs
```diff
@@ -237,6 +238,126 @@ impl CeramicEventService {
...
+                            event_cid,
+                            init,
+                            previous: vec![],
+                            data: init_event.data().to_json_bytes().map_err(|e| {
+                                Error::new_app(anyhow::anyhow!(
+                                    "Failed to serialize IPLD data: {}",
+                                    e
+                                ))
+                            })?,
+                            index: delivered as u64, // TODO: Implement proper indexing
```
samika98, line 288: nit : remove todo

event/src/unvalidated/event.rs
```diff
@@ -229,9 +248,11 @@ impl<D> From<Box<TimeEvent>> for Event<D> {
     }
 }
 
-impl<D> From<init::Payload<D>> for Event<D> {
-    fn from(value: init::Payload<D>) -> Self {
-        Self::Unsigned(value)
+// TODO : Delete this? then delete all
+// Delete event with new
```
samika98, line 252: @nathanielc All cannot be delete since rest of the two impls are being used in migration code. Makes sense to keep all 3 then?

event/src/unvalidated/event.rs
```diff
@@ -92,16 +107,17 @@ where
         }
     }
 
+    // TODO : Returns the cid no option, update the comment
```
nathanielc, line 110: nit, this todo is complete

event/src/unvalidated/event.rs
```diff
@@ -92,16 +107,17 @@ where
...
     /// Returns the 'id' field of the event, which is the Cid of the stream's init event.
     /// If this event *is* the init event, then it doesn't know its own Cid and returns None.
```
nathanielc, line 112: Update this comment to indicate we always return the cid because and init event does know its cid

event/src/unvalidated/event.rs
nathanielc: Yes lets keep them

flight/src/types.rs
```diff
@@ -22,11 +25,11 @@ impl ConclusionEvent {
     }
 }
 
+//TODO : Add Stream Type here
```
nathanielc, line 28: we have an issue to track this, so we can remove this comment for now

service/src/event/service.rs
```diff
@@ -237,6 +238,126 @@ impl CeramicEventService {
...
+    #[allow(dead_code)]
+    async fn transform_raw_events_to_conclusion_events(
+        &self,
+        raw_event: ceramic_event::unvalidated::Event<Ipld>,
+        delivered: i64,
+    ) -> Result<ConclusionEvent> {
+        let stream_cid = raw_event.id();
+        let init_event = self.get_event_by_cid(stream_cid).await?;
+        let init = ConclusionInit::try_from(init_event).unwrap();
+        let event_cid = *raw_event.id();
```
nathanielc, line 250: This is incorrect, the event_cid is not the stream_cid (unless its an init event). Does the raw_event know its own cid? That is what we need to use. Maybe we need a new method on unvalidated::Event to get the cid of itself.

store/src/sql/access/event.rs
```diff
@@ -295,48 +295,57 @@ impl CeramicOneEvent {
...
 
-    /// Returns the root CIDs and the data of all the events found after the given delivered value.
+    /// Returns the root CIDs and values of all the events found after the given delivered value.
     pub async fn new_events_since_value_with_data(
         pool: &SqlitePool,
-        delivered: i64,
+        highwater: i64,
         limit: i64,
-    ) -> Result<(i64, Vec<(Cid, unvalidated::Event<Ipld>)>)> {
+    ) -> Result<(i64, Vec<(Cid, unvalidated::Event<Ipld>, i64)>)> {
```
nathanielc, line 303: Can you change this to use a struct instead of a tuple? Its confusing what the different i64s are for. I get it after reading the code, however a struct would make it more clear.

store/src/sql/access/event.rs
```diff
@@ -295,48 +295,57 @@ impl CeramicOneEvent {
...
 
         let mut all_blocks: Vec<DeliveredEventBlockRow> =
             sqlx::query_as(EventQuery::new_delivered_events_with_data())
-                .bind(delivered)
+                .bind(highwater)
                 .bind(limit)
                 .fetch_all(pool.reader())
                 .await?;
 
-        // default to the passed in value if there are no new events to avoid the client going back to 0
```
nathanielc, line 330: why was this comment removed? It still seems valid.

store/src/sql/access/event.rs
```diff
@@ -295,48 +295,57 @@ impl CeramicOneEvent {
...
         let max_highwater = all_blocks
             .iter()
             .map(|row| row.new_highwater_mark + 1)
             .max()
-            .unwrap_or(delivered);
+            .unwrap_or(highwater);
         all_blocks.sort_by(|a, b| a.new_highwater_mark.cmp(&b.new_highwater_mark));
-        let blocks = all_blocks.into_iter().map(|b| b.block).collect();
+
+        let blocks = all_blocks.clone().into_iter().map(|b| b.block).collect();
```
nathanielc, line 341: Why do we have to clone the blocks?

nathanielc (changes requested):
I like it, a few questions below. The refactor worked out well[

nathanielc:
I'd like to see a test that shows we are correctly converting from events to conclusion events

service/src/event/service.rs
samika98: we can also pass it as a parameter into this method

store/src/sql/access/event.rs
```diff
@@ -295,48 +295,57 @@ impl CeramicOneEvent {
...
         let values = ReconEventBlockRaw::into_events(blocks).await?;
-        Ok((max_highwater, values))
+        let result = values
+            .into_iter()
+            .zip(all_blocks.iter())
```
samika98, line 363: we use them here as well, we add the delivered field to all the blocks here

service/src/event/ordering_task.rs
```diff
@@ -544,12 +544,8 @@ impl OrderingState {
...
 
             event_cnt += 1;
-            let stream_cid = parsed_event
-                .id()
-                .expect("id must exist for non-init events");
-            let prev = parsed_event
-                .id()
-                .expect("prev must exist for non-init events");
+            let stream_cid = parsed_event.id();
+            let prev = parsed_event.id();
```
dav1do, line 548: This should be `prev()`.. it looks like we introduced a bug here that was released in v0.33 (made latest this week)... fortunately it's the startup task code and has probably never done anything. We don't have long streams currently so things shouldn't be getting picked up at startup, meaning we we're probably okay without doing any clean up, but I'm going to add a ticket to investigate how it got through/improve the tests.

dav1do: fixed in #516

nathanielc (approved):

PR #509, opened 2024-08-29
refactor: remove recon client/server and send empty ranges when we don't need anything

dav1do:
This has two main changes, the first is to remove the recon client/server and the second is to reply with empty ranges when possible.

1) We no longer need exclusive access to recon so removing the client/server loop abstraction as it currently acts as an extra layer to update. I added Clone to a couple things and had to adjust generics.. I think it's okay but probably merits a close review. The tests now use worker threads instead of spawning the server in the background. This changed the order that things get sent, but it looks like all the same messages are exchanged.

2) When the remote is missing a range, we now send an empty response when we don't want anything. We still have to send some ranges in these cases if we're missing the bounding key and need to discover it. This seems to have reduced the round trips in some cases. The tests for this change was updated in the second commit, so looking at each commit to see the test changes is probably wise (or I can split it into two PRs if preferred). The goal of this change is to remove the requirement that the peer "read their writes". Previously, we would send a bunch of values and a hash and they would have to persist them _before_ they could reply to indicate we're in sync. Now we assume the range is done and move on, coming back in a future conversation if necessary.


nathanielc (changes requested):
This is great, happy to see the simplification here. 

I was concerned about the new reply_with field in remote missing as it didn't quite fit in the Recon model. I played around with it and found a simple way to avoid having to change the SyncState enum. 

Here is a PR to this branch that shows my thinking. https://github.com/ceramicnetwork/rust-ceramic/pull/510

nathanielc (approved):

PR #512, opened 2024-09-03
feat: use key value pairs for experimental feature flags

stephhuynh18:
I removed none because I didn't know what to do with it
Used https://github.com/ceramicnetwork/rust-ceramic/pull/448#pullrequestreview-2213083696 as reference 
Can comma separate or set `--experimental-feature-flags` multiple times

### DIFF
--- a/one/src/feature_flags.rs
+++ b/one/src/feature_flags.rs
@@ -17,9 +17,16 @@ impl std::str::FromStr for ExperimentalFeatureFlags {
         }
 
         match parts[0].to_ascii_lowercase().as_str() {
-            "none" => Ok(ExperimentalFeatureFlags::None),
-            "authentication" => Ok(ExperimentalFeatureFlags::Authentication),
-            "event-validation" => Ok(ExperimentalFeatureFlags::EventValidation),
+            "authentication" => match parts[1].trim().to_ascii_lowercase().as_str() {
+                "true" => Ok(ExperimentalFeatureFlags::Authentication),
+                "false" => Ok(ExperimentalFeatureFlags::None),
+                _ => Err(anyhow!("invalid value for the 'authentication' flag, should be 'true' or 'false'")),
+            },
+            "event-validation" => match parts[1].trim().to_ascii_lowercase().as_str() {
+                "true" => Ok(ExperimentalFeatureFlags::EventValidation),
+                "false" => Ok(ExperimentalFeatureFlags::None),
+                _ => Err(anyhow!("invalid value for the 'event-validation' flag, should be 'true' or 'false'")),
+            },
             _ => Err(anyhow!("invalid value")),
         }
     }
--- a/one/src/feature_flags.rs
+++ b/one/src/feature_flags.rs
@@ -10,8 +10,13 @@ pub(crate) enum ExperimentalFeatureFlags {
 impl std::str::FromStr for ExperimentalFeatureFlags {
     type Err = anyhow::Error;
 
-    fn from_str(value: &str) -> Result<Self> {
-        match value.to_ascii_lowercase().as_str() {
+    fn from_str(s: &str) -> Result<Self> {
+        let parts: Vec<&str> = s.splitn(2, '=').collect();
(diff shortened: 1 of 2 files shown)


one/src/feature_flags.rs
```diff
@@ -10,11 +10,27 @@ pub(crate) enum ExperimentalFeatureFlags {
 impl std::str::FromStr for ExperimentalFeatureFlags {
     type Err = anyhow::Error;
 
-    fn from_str(value: &str) -> Result<Self> {
-        match value.to_ascii_lowercase().as_str() {
-            "none" => Ok(ExperimentalFeatureFlags::None),
-            "authentication" => Ok(ExperimentalFeatureFlags::Authentication),
-            "event-validation" => Ok(ExperimentalFeatureFlags::EventValidation),
+    fn from_str(s: &str) -> Result<Self> {
```
AaronGoldman, line 13: Do we need a `from_str`? I think that if you just don't have a from string that CLAP will parse out the comma separated list? 

did we try making 
```rs
#[derive(Clone, Debug, PartialEq)]
pub(crate) enum ExperimentalFeatureFlags {
    Authentication(Bool),
    EventValidation(Bool),
}
```
and just seeing if clap gets it write. also 
https://github.com/ceramicnetwork/rust-ceramic/blob/main/one/src/lib.rs#L125C7-L125C7

Do we want to have a `ExperimentalFeatureFlags`
with https://docs.rs/clap/latest/clap/struct.Arg.html#method.requires we could have a single bool flag `ExperimentalFeatures` and then just require the `ExperimentalFeatures` flag to use those features. the individual flags are just normal CLAP arguments with a requirement. This pattern will also allow us to have a `DeprecatedFeatures` flag for flags we are panning to remove. 

when a flag gos from ExperimentalFeatures to release feature the operator dose not need to change anything it still works with the ExperimentalFeatures=true it is just no longer required. Same for the DeprecatedFeatures the operator can have DeprecatedFeatures enabled in prod so things don't brake suddenly and disabled in test so they have warning of DeprecatedFeatures.

I think we should consider the `ExperimentalFeatures`, `DeprecatedFeatures`, `method.requires` model with Clap doing the parsing.

AaronGoldman:
```rs
    /// Unique key used to find other Ceramic peers via the DHT
    #[arg(long, default_value = "testnet-clay", env = "CERAMIC_ONE_NETWORK")]
    network: Network,

    /// Unique id when the network type is 'local'.
    #[arg(long, env = "CERAMIC_ONE_LOCAL_NETWORK_ID", requires = "network")]
    local_network_id: Option<u32>,
```
You can't local_network_id if the network is not local.
```rs

local_network_id
```rs
/// allow experimental features
#[arg(long, default_value = "false", env = "CERAMIC_ONE_EXPERIMENTAL_FEATURES")]
experimental_features: bool,
/// allow deprecated features
#[arg(long, default_value = "false", env = "CERAMIC_ONE_DEPRECATED_FEATURES")]
deprecated_features: bool,
/// enable authentication
#[arg(long, default_value = "false", env = "CERAMIC_ONE_AUTHENTICATION", requires = "experimental_features")]
authentication: bool,
/// enable event validation
#[arg(long, default_value = "false", env = "CERAMIC_ONE_EVENT_VALIDATION", requires = "experimental_features")]
event_validation: bool,
```

stbrody:
Recapping a conversation from discord: I'd love to eliminate "experimental" feature flags and collapse them into regular feature flags.  

my vote would be for a single flag, `--feature-flags` that can be set multiple times, where each flag is a key value pair.
For example: 
`ceramic-one daemon --feature-flags validation=enabled --feature-flags anchoring=delegated`

or possibly: `ceramic-one daemon --feature-flags validation=enabled,anchoring=delegated`

I don't feel strongly on comma separated vs specify the flag multiple times.  I lean towards specifying multiple times though as I think it will make it easier to have flags that are themselves lists and need comma separation

stephhuynh18:
Closing in favor of aarons suggestion

PR #513, opened 2024-09-05
refactor: create explicit event and interest services with separate stores.

nathanielc:
Establishes a pattern of services and stores.

A service is a public API to be consumed by other services in the application.
A store is direct access to a sqlite database.

Services have their own stores. Stores never touch tables that belong to other services.


dav1do (approved):
:shipit: Bring on the merge conflicts!

stbrody (commented):
nice

stbrody (approved):

Cargo.toml
```diff
@@ -6,22 +6,23 @@ members = [
     "car",
     "core",
     "event",
```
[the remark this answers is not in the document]
nathanielc: The `ceramic-event` crate contains the event types with no behavior other than encoding/decoding them. This is useful for other downstream projects (like potential future clients) that want to leverage the event types without bringing all the dependencies of the `event-svc` (sqlite for example).

event-svc/src/store/metrics.rs
```diff
@@ -123,9 +123,9 @@ impl<S: Send + Sync> StoreMetricsMiddleware<S> {
 }
 
 #[async_trait]
-impl<S> ceramic_api::InterestStore for StoreMetricsMiddleware<S>
+impl<S> ceramic_api::InterestService for StoreMetricsMiddleware<S>
```
smrz2001, line 126: This is still needed in the Event Service?

interest-svc/src/store/metrics.rs
```diff
@@ -0,0 +1,336 @@
...
+            &self.metrics,
+            "api_interest_range",
+            self.store.range(start, end, offset, limit),
+        )
+        .await
+    }
+}
+
+#[async_trait]
+impl<S> ceramic_api::EventService for StoreMetricsMiddleware<S>
```
smrz2001, line 157: This is still needed in the Interest Service?

event-svc/src/store/metrics.rs
nathanielc: nope good catch will remove, and the mirror as well.

PR #514, opened 2024-09-05
fix: updates detection of tile docs and doesn't log them by default

nathanielc:
With this change we catch two more cases where we can detect tile docs. Additionally by default we only count the number of tile docs we find and do not log any errors about them. This way any tile docs do not create a feeling of failure when migrating.


dav1do (approved):
Makes sense to me. Error enums are great :shipit:

PR #515, opened 2024-09-06
refactor: simplify daemon logic

nathanielc:
This removes some of the abstraction around sqlite to make creating and initializing services more straightforward. We will likely need to add new abstractions when we want to support parquet or other storage formats however those can be service focused instead of db focused.


stbrody (approved):

dav1do (approved):
:shipit:

PR #516, opened 2024-09-06
fix: start up ordering task bug

dav1do:
During a recent refactor (released in 0.33), prev was incorrectly set to the init_cid in the startup ordering task. This would have caused events to have been "delivered" in a possibly random order, as everything is after the init event. As this was just introduced and we generally have short streams that should be ordered inline, it is extremely unlikely that it caused any problems that merit a clean up migration (i.e. delete the current value, read all events to rewrite the deliverable order and tell js-ceramic to indexing again) so that is not included. 

I also deleted a log statement I accidentally left as info during debugging that shouldn't have landed. It could probably be trace but for now I just removed it as we have prometheus and other metrics around that.

### DIFF
--- a/event-svc/src/event/ordering_task.rs
+++ b/event-svc/src/event/ordering_task.rs
@@ -549,7 +549,7 @@ impl OrderingState {
                 .id()
                 .expect("id must exist for non-init events");
             let prev = parsed_event
-                .id()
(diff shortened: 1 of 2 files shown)


smrz2001 (approved):
:shipit:

PR #518, opened 2024-09-09
chore: remove release event trigger for deb repo action

dav1do:
We don't currently need this action to run on releases (and it's failing), so it's manual only now.


dav1do:
Closing in favor of #519

PR #520, opened 2024-09-09
feat: rpc client for eth blockchains

dav1do:
This PR adds an http client that knows how to talk to Ethereum RPC providers. It also validates that a time event proof correctly points at `prev` when parsing. The RPC client will be used to timestamp events which is coming in a follow up PR. That change will require storing more information and is related to both anchoring and timestamping. Event validation itself doesn't strictly require that we're able to follow a time event to the chain and prove that the root is in the proof. Without that, we have no time information but the event is "valid" from a structural perspective, and it's possible users will not all use the same chain, and we may want to gossip events even if we're not using them to timestamp ourselves.

The intention is to have the service (likely event) that is actually adding time information register a eth provider for every chain ID that they're interested in, and then they can query the blockchain and persist the block number and time information and add it to any events as needed. These changes are consumed by #522.

I expect that we will swap this out for a full eth client (e.g. alloy-rs) when we do self anchoring and need signing and such, but that should be a relatively small change as we can just impl the trait in terms of that client and it should work.

~~This currently targets #503 but can likely be rebased onto main at this point if that's preferred.~~


validation/src/blockchain/eth_rpc.rs
```diff
@@ -0,0 +1,295 @@
...
+    block_hash: Option<String>,
+    /// Transaction hash
+    hash: String,
+    /// contract input data
+    input: String,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[serde(rename_all = "camelCase")]
+pub struct EthBlock {
```
stbrody, line 81: some comments on why this is different than BlockResponse and the relationship between those two types would be helpful

validation/src/blockchain/eth_rpc.rs
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    /// Get the latest block number
+    ///
+    /// curl https://mainnet.infura.io/v3/{api_token} \
+    /// -X POST \
+    /// -H "Content-Type: application/json" \
+    /// -d '{"jsonrpc":"2.0","method":"eth_blockNumber","params": [],"id":1}'
+    /// >> {"jsonrpc": "2.0", "id": 1, "result": "0x127cc18"}
+    async fn _eth_block_number(&self) -> Result<RpcResponse<String>> {
```
stbrody, line 160: is this used anywhere?  Do we need it?

stbrody (commented):
Tagging in @AaronGoldman as I believe there may already be code checked in somewhere that does this?

validation/src/blockchain/eth_rpc.rs
```diff
@@ -0,0 +1,295 @@
...
+
+static HTTP_CLIENT: Lazy<reqwest::Client> = Lazy::new(reqwest::Client::new);
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub struct ChainTransaction {
+    /// Transaction hash
+    pub hash: String,
+    /// Transaction contract input
+    pub input: String,
+    pub block: Option<BlockResponse>,
```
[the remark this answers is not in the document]
dav1do: it can be done if the transaction hasn't been mined yet.. not sure if we actually want to deal with that or just error, but the current code has these cases:

1. transaction exists but not mined (no timestamp yet)
2. block mined -> timestamp available
3. transaction not found (may be an error, may just not exist on our provider for some reason)
4. error communicating with rpc endpoint

validation/src/blockchain/eth_rpc.rs
dav1do: yea, fair enough. this is the actual RPC response and I wanted the public type the trait relies on to convert the blockchain representation (hex strings) to integers before going up a level

validation/src/blockchain/eth_rpc.rs
dav1do: no, probably not... it was originally used to make sure we had mined "enough blocks" to consider the timestamp valid.. I left if in here in case we want to do that but it's simple to add back later.

event/src/unvalidated/event.rs
```diff
@@ -48,6 +46,29 @@ fn get_time_event_witness_blocks(
...
+            *parts
+                .last()
+                .ok_or_else(|| anyhow!("Time Event path must have at least one item"))?,
+        )?
+        .ok_or_else(|| anyhow!("Time Event path failed to resolve last element"))?
+    {
+        Ipld::Link(cid) => {
+            if *cid != event.prev() {
+                bail!(
+                    "the anchor commit proof {} with path {} points to invalid 'prev' commit",
```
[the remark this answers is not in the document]
dav1do: yeah, will update

dav1do:
> Tagging in @AaronGoldman as I believe there may already be code checked in somewhere that does this?

This is taken from that code and adjusted to a bit (mostly just more serde and a trait).. it was removed from main a while back.

validation/src/blockchain/eth_rpc.rs
stbrody: yeah makes sense and is what I figured, just would be nice to make it clearer with a couple of comments

validation/src/blockchain/eth_rpc.rs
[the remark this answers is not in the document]
stbrody: Makes sense.  I think a simple comment saying it's None if the txn hasn't been mined yet would be fine

event-svc/benches/sqlite_store.rs
```diff
@@ -8,10 +8,10 @@ use ceramic_event::unvalidated::{
 };
 use criterion2::{criterion_group, criterion_main, BatchSize, Criterion};
 use ipld_core::ipld::Ipld;
-use itertools::Itertools;
+use itertools::Itertools as _;
```
AaronGoldman, line 11: why _

dav1do: There's really not point to this change.. it was showing up as unneeded so I deleted it but it *was* needed so my IDE added it back with _.. but generally traits only need to be in scope, so using `as _` can be nice to avoid name collisions (plus it makes it clear it's a trait by convention) but it's same same.

validation/src/blockchain/eth_rpc.rs
```diff
@@ -0,0 +1,283 @@
...
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+/// The block information returned by eth_getBlockByHash
+#[serde(rename_all = "camelCase")]
+struct EthBlock {
+    hash: String,
+    /// hexadecimal block number
+    number: String,
+    /// hexademical representing unix epoch time
```
AaronGoldman, line 93: '0x' prefixed hexademical representing unix epoch time

stephhuynh18 (approved):

validation/src/blockchain/eth_rpc.rs
dav1do: updated

PR #521, opened 2024-09-09
feat: make all feature flags top level 

stephhuynh18:
Experimental features will require the flag `--experimental-features` to be set
Relevant comment here: https://github.com/ceramicnetwork/rust-ceramic/pull/512#discussion_r1745802124


one/src/daemon.rs
```diff
@@ -139,23 +139,35 @@ pub struct DaemonOpts {
...
 
-    /// Enable feature flags
+    /// Enable authentication; Requires using the experimental-features flag
     #[arg(
         long,
-        use_value_delimiter = true,
-        value_delimiter = ',',
-        env = "CERAMIC_ONE_FEATURE_FLAGS"
+        default_value_t = false,
+        env = "CERAMIC_ONE_AUTHENTICATION",
```
stbrody, line 158: huh, I know this isn't really the focus of this PR, but why is this flag called "authentication"?  Isn't this about enabling the access control experiment that @dbcfd was working on?  That's more about access control / authorization than authentication, right?

PR #523, opened 2024-09-10
feat: add flight sql server to expose conclusion feed.

nathanielc:
The Flight SQL Server allows using SQL to query the conclusion feed. Events are read from the db and transformed into conclusion events and returned out the feed.

Closes AES-288


flight/src/server.rs
```diff
@@ -0,0 +1,263 @@
...
+                    DataType::List(Arc::new(Field::new("item", DataType::Binary, false))),
+                    true,
+                ),
+            ])),
+        }
+    }
+    fn highwater_mark_from_expr(expr: &Expr) -> Option<u64> {
+        let find_highwater_mark = |col: &Expr, lit: &Expr| {
+            col.try_as_col()
+                .map(|column| column.name == "index")
```
dav1do, line 104: nit: `map_or(false, || )` instead of `unwrap_or(false)`

dav1do (commented):
Looks good, a couple minor questions. The table provider and execution plan implementations seem reasonable, but I don't have the context to provide much insight there.

flight/src/server.rs
```diff
@@ -0,0 +1,263 @@
...
+
+async fn run_service(addr: SocketAddr, svr: Router) -> anyhow::Result<()> {
+    info!(%addr, "FlightSQL server listening");
+    Ok(svr.serve(addr).await?)
+}
+
+/// A ConclusionFeed provides access to [`ConclusionEvent`]s.
+#[async_trait::async_trait]
+pub trait ConclusionFeed: Send + Sync {
+    /// Produce a set of conlcusion events up to the limit with an index greater than the highwater_mark
```
stbrody, line 56:
```suggestion
    /// Produce a set of conclusion events up to the limit with an index greater than the highwater_mark
```

flight/src/types.rs
```diff
@@ -17,41 +24,58 @@ impl AsRef<ConclusionEvent> for ConclusionEvent {
...
         }
     }
 }
 
-// Dimension is a tuple of (name, value) : way to identify a stream
+/// ConclusionInit is static metadata about a stream.
+/// All events within a stream have the same ConclusionInit.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct ConclusionInit {
+    /// The CID of the stream. Can be used as a unique identifier for the stream.
```
stbrody, line 39:
```suggestion
    /// The CID of the init event of the stream. Can be used as a unique identifier for the stream.
```

flight/src/types.rs
```diff
@@ -17,41 +24,58 @@ impl AsRef<ConclusionEvent> for ConclusionEvent {
...
+    /// This is not the StreamId as it does not contain the stream type.
```
stbrody, line 40:
```suggestion
    /// This is not the StreamId as it does not contain the StreamType.
```

flight/src/types.rs
```diff
@@ -17,41 +24,58 @@ impl AsRef<ConclusionEvent> for ConclusionEvent {
...
+    /// Index of the event. See [`ConclusionEvent`] for invariants about the index.
+    pub index: u64,
+    /// The CID of the event itself. Can be used as a unique identifier for the event.
     pub event_cid: Cid,
+    /// The stream metadata of the event.
     pub init: ConclusionInit,
-    // TODO : rethink this, add a checkpoint to make it work in datafusion query
+    /// Ordered list of previous events this event references.
     pub previous: Vec<Cid>,
+    /// Raw bytes of the event data.
```
stbrody, line 61: maybe add some information about how this is encoded?

one/src/daemon.rs
```diff
@@ -168,6 +168,14 @@ pub struct DaemonOpts {
         requires = "experimental_features"
     )]
     event_validation: bool,
+
+    /// Flight SQL bind address; Requires using the experimental-features flag
```
stbrody, line 172: do we want a separate flag for setting the port number?

stbrody (commented):
lgtm, a few small comment suggestions

one/src/daemon.rs
nathanielc: I prefer fewer flags, this handle both host and port in a single flag.

flight/tests/server.rs
```diff
@@ -0,0 +1,190 @@
...
+    FlightSqlServiceClient::new(channel(&addr).await)
+}
+
+async fn execute_flight(
+    client: &mut FlightSqlServiceClient<Channel>,
+    info: FlightInfo,
+) -> Result<RecordBatch> {
+    let schema = Arc::new(Schema::try_from(info.clone()).context("valid schema")?);
+    let mut batches = Vec::with_capacity(info.endpoint.len() + 1);
+    batches.push(RecordBatch::new_empty(schema.clone()));
```
samika98, line 50: why are we pushing an empty record batch here. Is it a placeholder for the schema? Can we add a comment here that explains the reasoning.

flight/src/server.rs
```diff
@@ -0,0 +1,267 @@
...
+        filters: &[Expr],
+        limit: Option<usize>,
+    ) -> datafusion::common::Result<Arc<dyn ExecutionPlan>> {
+        let schema = projection
+            .map(|projection| self.schema.project(projection))
+            .transpose()?
+            .map(Arc::new)
+            .unwrap_or_else(|| self.schema.clone());
+        debug!(?schema, "projected schema");
+        Ok(Arc::new(FeedExec {
```
samika98, line 160: nit : I see that the FeedExec implements the ExecutionPlan. Can we add comments as to when the execute gets invoked. Also some explanatory comments in the execute method might help. Why are we setting the limit to 10000 by default?

samika98 (approved):
Some nit's about more explanatory comments

dav1do (approved):

flight/tests/server.rs
nathanielc: This is test code, the empty batch ensures that when we print out the results the headers always print even if the results are empty. I'll add a comment

flight/src/server.rs
nathanielc: I'll add some comments, but these type implement the traits from datafusion which are well commented. I recommend reading the TableProvider and ExecutionPlan trait comments.

PR #525, opened 2024-09-11
feat: infer stream type from raw event model and add it to conclusion event + tests

samika98:
Infer stream type from the header of the init event. Apply this type while creating conclusion events.


flight/src/types.rs
```diff
@@ -60,9 +62,19 @@ impl TryFrom<unvalidated::Event<Ipld>> for ConclusionInit {
     fn try_from(event: unvalidated::Event<Ipld>) -> Result<Self> {
         let init_payload = event
             .init_payload()
-            .ok_or_else(|| anyhow!("no init payload found"))?;
+            .ok_or_else(|| anyhow!("malformed event: no init payload found"))?;
+        let model = init_payload.header().model();
+        let stream_type = StreamId::try_from(model)
```
samika98, line 67: Do we want to move this logic to `init.rs` or should it live here? I prefer it to be here since this seems like an inference of what we pass into the struct. The init header struct does not add any checks while populating this field. Any thoughts on this?

nathanielc: I like the logic here as its part of the mapping from raw event to conclusion event. However this logic is incorrect and only passes tests because the tests are also incorrect.

This is subtle, let me start with the structure of an stream id.

As you have seen a stream id is an encoded tuple of (stream type, init cid).
There are two stream ids at play in an init event. First, the stream id of the init event itself, lets call this one the init_stream_id. Second the stream id of the model that the stream belongs to, lets call this one the model_stream_id.

Therefore a given init event the model_stream_id is found in the model field of the header. This is the one the code above is decoding. The init_stream_id is not part of the init event since it is the cid of the init event itself. 

The stream id we need to construct for a conclucsion event is the init_stream_id, i.e. the stream id of the stream to which the event belongs. 

The init_stream_id can have two possible stream types, Model or ModelInstanceDocument depending. However the model_stream_id is always a stream type of Model. This is because a stream cannot belong to a ModelInstanceDocument stream. 

So back to the code above. The code is decoding the model_stream_id type and using that as the stream type for the init_stream_id which is incorrect. However if you look into the [init_event](https://github.com/ceramicnetwork/rust-ceramic/blob/9c2575963f421bf8052966aafc577403dd649e25/event-svc/src/tests/mod.rs#L144) function is takes a stream id and uses it as the model field for the init event. However it does not guarantee that the stream type is always a Model type. If you add these lines to that function the test fails.

```rust
    assert_eq!(
        model.r#type,
        StreamIdType::Model,
        "an init event's model stream id must always be a model stream type"
    );
```

This stems from the `create_deterministic_stream_id_model` and `create_deterministic_stream_id_document` functions  and how they are used. The test creates two stream ids. Both of these stream ids are used as the model field in the init event. Therefore both of these stream ids should be model stream ids. This would make the above assertion pass but then the `ConclucsionInit::try_from` code would then always produce a stream type of model for the init_stream_id, which is incorrect. 

So how can we determine the type of the init_stream_id? This needs one more piece of information. Since model streams must also belong to a model stream there is a special known stream (sometimes called the metamodel) to which all model streams belong. We can use this information to recover the type of the init_stream_id. If the model of the init event is equal to the metamodel stream id then we known that the init_stream_id type is model. If its any other stream id then its a document type. 

All of this is to say the code above should do a simple equality comparison of the init event model field with the metamodel stream id to determine the stream type.

The metamodel stream id is `kh4q0ozorrgaq2mezktnrmdwleo1d` we should create a constant of this value along side the StreamId type. It comes from [here](https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/stream-model/src/model.ts#L121)

event-svc/src/tests/mod.rs
```diff
@@ -243,8 +243,8 @@ pub(crate) async fn generate_chained_events() -> Vec<ReconItem<EventId>> {
     let mut events: Vec<ReconItem<EventId>> = Vec::with_capacity(5);
 
     let signer = Box::new(signer().await);
-    let stream_id_1 = create_deterministic_stream_id(&[0x01]);
-    let stream_id_2 = create_deterministic_stream_id(&[0x02]);
+    let stream_id_1 = create_deterministic_stream_id_document(&[0x01]);
+    let stream_id_2 = create_deterministic_stream_id_model(&[0x02]);
```
nathanielc, line 247: per explanation above, both of these should be model stream ids as they are the id of the stream that these new event belong to. Streams can only belong to model stream, not document streams.

flight/src/types.rs
nathanielc: As an aside we should probably create an MID builder that expects an actual stream id for the model value and can verify that it is a model stream instead of using the vanilla event builder. Not something we should do in this PR, but would be a good API design to catch this bug at compile time.

nathanielc (changes requested):
Overall looks good with exception to the bug called out below.

flight/src/types.rs
stbrody: Note the metamodel streamid is of type UNLOADABLE, not Model

event-svc/src/tests/mod.rs
```diff
@@ -330,10 +333,29 @@ pub(crate) async fn generate_chained_events() -> Vec<ReconItem<EventId>> {
...
+/// A `StreamId` representing the metamodel stream.
+///
+/// # Example
+///
+/// ```rust
+/// let stream_id = create_meta_model_stream_id();
+/// assert_eq!(stream_id.to_string(), "kh4q0ozorrgaq2mezktnrmdwleo1d");
+/// ```
+fn create_meta_model_stream_id() -> StreamId {
+    StreamId::from_str(METAMODEL_STREAM_ID).unwrap()
```
samika98, line 360: Is there a better way to create tests for this @nathanielc

flight/src/types.rs
```diff
@@ -82,11 +89,34 @@ impl TryFrom<unvalidated::Event<Ipld>> for ConclusionInit {
...
             .init_payload()
-            .ok_or_else(|| anyhow!("no init payload found"))?;
+            .ok_or_else(|| anyhow!("malformed event: no init payload found"))?;
+
+        // Get the model from the init header
+        // The model indicates the creator of the stream
+        let model = init_payload.header().model();
+
+        // Define the metamodel stream ID
+        let meta_model_stream_id = StreamId::from_str(METAMODEL_STREAM_ID)?;
```
stbrody, line 102: can we make *this* the constant we define at the top of the file?  Instead of converting from string to StreamId for every event.

flight/src/tests.rs
```diff
@@ -0,0 +1,220 @@
+use super::*;
```
samika98, line 1: Will remove this file

flight/src/types.rs
nathanielc: I did this here, https://github.com/ceramicnetwork/rust-ceramic/pull/529 turns out it was a bit more complicated that I expected but we have it now.

nathanielc (approved):

PR #527, opened 2024-09-12
feat: add olap aggregation function

nathanielc:
The aggregation function compute the stream state for each new event into a stream. All stream states are persisted into a parquet database.

TODO:

- [x] Use schema version naming
- [x] CLI args for flight sql path

Closes AES-289
Closes AES-290


olap/src/aggregator/ceramic_patch.rs
```diff
@@ -0,0 +1,144 @@
...
+            .map_err(|err| exec_datafusion_err!("Error parsing previous state: {err}"))?;
+        json_patch::patch(&mut new_state, &patch)
+            .map_err(|err| exec_datafusion_err!("Error applying JSON patch: {err}"))?;
+        serde_json::to_string(&new_state)
+            .map_err(|err| exec_datafusion_err!("Error JSON encoding: {err}"))
+    }
+}
+
+impl PartitionEvaluator for CeramicPatchEvaluator {
+    fn evaluate_all(&mut self, values: &[ArrayRef], num_rows: usize) -> Result<ArrayRef> {
```
samika98, line 81: Using topological sorting here (resolving events in order such that no event resolved before its dependency) will be a more optimal algorithm here. 
If there were cyclic dependencies in the current approach, the code will not terminate. 
Can we do this using topological sort + cycle detection.

samika98 (changes requested):
Few questions about the evaluate_all code. 
I believe using topological sort would be a neater solution than the current iterative + hashmap approach. 
Also concerned about termination during cycles in the current code. We might need to handle that and a return exec error

olap/src/aggregator/ceramic_patch.rs
nathanielc: Yeah we sort by index now so I can rely on in order delivery of the events. I'll update the algo.

CODEOWNERS
```diff
@@ -11,11 +11,13 @@
 /ci-scripts/        @smrz2001       @ceramicnetwork/eng-reviewers
 /core/              @nathanielc     @ceramicnetwork/eng-reviewers
 /event/             @nathanielc     @ceramicnetwork/eng-reviewers
+/flight/            @nathanielc     @ceramicnetwork/eng-reviewers
 /fpm/               @dbcfd          @ceramicnetwork/eng-reviewers
 /kubo-rpc-server/   @nathanielc     @ceramicnetwork/eng-reviewers
 /kubo-rpc/          @nathanielc     @ceramicnetwork/eng-reviewers
 /metrics/           @dav1do         @ceramicnetwork/eng-reviewers
 /migrations/        @dav1do         @ceramicnetwork/eng-reviewers
+/olap/              @nathanielc     @ceramicnetwork/eng-reviewers
```
samika98, line 20: same as line 26

nathanielc:
@samika98 I have updated the algo to do a single pass.

olap/src/aggregator/ceramic_patch.rs
```diff
@@ -0,0 +1,147 @@
...
+                //Init event, patch value is the initial state
+                if patches.is_null(i) {
+                    // If we have an init event without data use an empty object as the initial
+                    // state. This feels like a leaky abstraction is this expected based on the
+                    // Ceramic spec?
+                    new_states.append_value("{}");
+                } else {
+                    new_states.append_value(patches.value(i));
+                }
+            } else if let Some(previous_state) = if !previous_states.is_null(i) {
```
dav1do, line 98: I found this `else if let Some() = if/else {}` nesting hard to follow at first.. but looking at it in my IDE it's alright.. and I'm not sure what I'd prefer so just mentioning it as an observation without any expectation

olap/src/aggregator/ceramic_patch.rs
```diff
@@ -0,0 +1,147 @@
...
+            .map_err(|err| exec_datafusion_err!("Error parsing previous state: {err}"))?;
+        json_patch::patch(&mut new_state, &patch)
+            .map_err(|err| exec_datafusion_err!("Error applying JSON patch: {err}"))?;
+        serde_json::to_string(&new_state)
+            .map_err(|err| exec_datafusion_err!("Error JSON encoding: {err}"))
+    }
+}
+
+impl PartitionEvaluator for CeramicPatchEvaluator {
+    fn evaluate_all(&mut self, values: &[ArrayRef], num_rows: usize) -> Result<ArrayRef> {
```
dav1do, line 81: The main thing I struggled with was understanding the where things are coming from and what we're actually working with. The datafusion docs are pretty good for how this trait is used but but including a doc comment about what the expected input looks like would be helpful. Then again, the assignments into the fields makes it fairly clear, so maybe it's not useful 🤷

(these assignments)
```rust
        let event_cids = as_binary_array(&values[0])?;
        let previous_cids = as_binary_array(&values[1])?;
        let previous_states = as_string_array(&values[2])?;
        let patches = as_string_array(&values[3])?;

dav1do (approved):
I like the test. Some questions as it's a bit dense to those of us uninitiated with datafusion (their docs seem really good though so I just need to spend some time reading them).

olap/src/aggregator/ceramic_patch.rs
nathanielc: Same not sure how to best simplify, i'll play around with a few thing

PR #530, opened 2024-09-16
refactor: return reference to inserted item from event store and reduce unnecessary ordering work 

dav1do:
This PR refactors some of the code for the event service and store to do less cloning/copying of data, and do less database IO when we don't actually care. The commits should all pass tests independently, so we don't have to take all of them (or if any of the later commits are more contentious, I can move them to a separate PR).

Now when creating `EventInsertable` objects during validation, we mark the event as deliverable if it's an init event. This was one of the responsibilities of the first pass of "ordering" that has been obviated. Next, the first pass of ordering was modified to support only ordering against the in memory set as a "best effort" or using the database to ensure the complete history exists. For recon and migrations, this means we can do less work. 

Previously, we would attempt to find an events history in memory and in the database to see if we should mark it ready for the feed during writes. For recon, anything that couldn't be ordered would be sent to the ordering task while the migration writes wouldn't. Now we only check the in memory set for recon and let the ordering task deal with all IO, while for migrations we don't attempt to do any ordering. 

I based this on #503, as I thought more about it related to the deliverable/validation options (and in the context of needing to speed migrations up). It could probably be extracted but it will have conflicts. I'm considering those PRs more important so opening for review but it's not the most urgent IMO.

### DIFF
--- a/event-svc/src/event/order_events.rs
+++ b/event-svc/src/event/order_events.rs
@@ -6,6 +6,17 @@ use ceramic_sql::sqlite::SqlitePool;
 use crate::store::{CeramicOneEvent, EventInsertable};
 use crate::Result;
 
+/// Groups the events into lists of those with a delivered prev and those without. This can be used to return an error if the event is required to have history.
+/// The events will be marked as deliverable so that they can be passed directly to the store to be persisted. It assumes init events have already been marked deliverable.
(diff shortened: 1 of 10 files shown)


stephhuynh18 (approved):
😸

PR #531, opened 2024-09-17
fix: use build.rs to rebuild sql crate on migrations change

nathanielc:


dav1do (approved):
Bah, that's annoying. Good catch.

PR #535, opened 2024-09-18
feat: support dimensions in conclusion feed

nathanielc:
Adds support from the db out the flight API to handle dimensions. There is a mapping from existing events to dimensions on conclusion events.

The olap aggregator also preserves the dimensions during aggregation.


nathanielc (commented):
In live walk through we also discovered querying the dimensions in their current form is difficult. We should explore better options.

samika98 (approved):
LGTM! The cid_string_list really simplified the pretty_feed. Reading the unnest and aggregate was confusing this is way easier to read!

flight/src/conversion.rs
```diff
@@ -40,76 +53,54 @@ impl Default for ConclusionEventBuilder {
...
             }
+        };
+        self.stream_cid.append_value(init.stream_cid.to_bytes());
+        self.controller.append_value(&init.controller);
+        self.stream_type.append_value(init.stream_type);
+        for (k, v) in &init.dimensions {
+            self.dimensions.keys().append_value(k);
+            self.dimensions.values().append_value(v);
         }
+        let _ = self.dimensions.append(!init.dimensions.is_empty());
```
dav1do, line 85: I had to go look at the MapBuilder docs to understand this.. makes sense now.

flight/src/server.rs
```diff
@@ -96,6 +96,32 @@ impl<T> FeedTable<T> {
...
+                    DataType::Map(
+                        Field::new(
+                            "entries",
+                            DataType::Struct(
+                                vec![
+                                    Field::new("key", DataType::Utf8, false),
+                                    Field::new(
+                                        "value",
+                                        DataType::Dictionary(
+                                            Box::new(DataType::Int8),
```
dav1do, line 110: Out of curiosity, will we support any other types? I'm curious why you choose Int8 instead of 32 or 64 and just binary strings instead of utf8.

arrow-test/src/lib.rs
```diff
@@ -72,6 +78,70 @@ impl ScalarUDFImpl for CidString {
...
+    }
+    fn return_type(&self, _args: &[DataType]) -> datafusion::common::Result<DataType> {
+        Ok(DataType::new_list(DataType::Utf8, true))
+    }
+    fn invoke(&self, args: &[ColumnarValue]) -> datafusion::common::Result<ColumnarValue> {
+        let args = ColumnarValue::values_to_arrays(args)?;
+        let all_cids = as_list_array(&args[0])?;
+        let mut strs = ListBuilder::new(StringBuilder::new());
+        for cids in ArrayIter::new(all_cids) {
+            if let Some(cids) = cids {
```
dav1do, line 123: I find this odd (that we have values that are None) but I guess it's expected based on the Nullable definition. Does the DataType need to be nullable (and when would it be in practice)? I know this for testing so it's not a big deal, just trying to understand for when I have to add something.

dav1do (approved):
LGTM. A couple questions about datafusion things I don't quite grok but the code looks good.

arrow-test/src/lib.rs
nathanielc: yeah not real use case for when a cid would be null. For example the event_cid column is not nullable. However a this is a function that can be used on any function its best to not assume data is not null. And just pass nulls along.

flight/src/server.rs
nathanielc: Int8 is about how many unique values we can have. Int8 gives us 256. And as I write this I realize this is not enough. I was thinking about the number of unique dimension values for a single stream. However this needs to cover many streams combined. I'll likely pick i32/i64 as its plenty large.

flight/src/types.rs
```diff
@@ -117,7 +117,27 @@ impl TryFrom<unvalidated::Event<Ipld>> for ConclusionInit {
                 .first()
                 .ok_or_else(|| anyhow!("no controller found"))?
                 .to_string(),
-            dimensions: vec![],
+            dimensions: vec![
+                ("model".to_string(), init_payload.header().model().to_vec()),
+                (
+                    "controller".to_string(),
```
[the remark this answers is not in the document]
nathanielc: Should Index is not a dimension as it is mutable. We will need to add it to a wrapper object over the data. I'll create an issue to handle that. Unique is an interesting case. We discussed at length let me summarize.

Dimensions are by definition the fields that users can query over. Unique is not useful in querying for streams. Rather its only purpose is to ensure the cid of the init event is unique compared to other streams with the same dimensions. As a result while it exists as a header on the init event it does not need to be a dimension.

The sep concept is being replaced by dimensions and is now redundant so its not needed either.

PR #536, opened 2024-09-18
refactor: teach event validator some more and store a copy on the service

dav1do:
The pending/invalid logic didn't fit very nicely as implemented in #503. With #530 (this one's target), the return types of some of these signatures have been adjusted and it made the change to have the validator know whether a missing init event should be an error or pending/unvalidated easier. This feels like a more appropriate place to understand this, and simplifies the service logic a bit. 

It also adjusts how the event validator is constructed by the service so that we create one that we reuse, which will make it easier to construct at start up with the eth RPC providers to actually do the time event chain inclusion validation.

### DIFF
--- a/event-svc/src/event/service.rs
+++ b/event-svc/src/event/service.rs
@@ -50,6 +50,7 @@ pub struct EventService {
     pub(crate) pool: SqlitePool,
     validate_events: bool,
     delivery_task: DeliverableTask,
+    event_validator: EventValidator,
     pending_writes: Arc<Mutex<HashMap<Cid, Vec<UnvalidatedEvent>>>>,
 }
 /// An object that represents a set of blocks that can produce a stream of all blocks and lookup a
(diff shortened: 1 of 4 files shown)


event-svc/src/event/service.rs
```diff
@@ -304,17 +306,29 @@ impl EventService {
...
             }
         }
     }
 
+    /// Returns two vectors of Event IDs representing (new, existed)
+    fn partition_store_result(store: crate::store::InsertResult) -> (Vec<EventId>, Vec<EventId>) {
+        store.inserted.into_iter().partition_map(|e| {
+            let key = e.inserted.order_key().clone();
+            if e.new_key {
+                itertools::Either::Left(key)
```
AaronGoldman, line 325: Why are we using an Either rather than a more meaningful enum, where we can name left and right.

dav1do: that's the signature of the Itertools [partition_map](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.partition_map) function.. I could make a struct return instead of the two vecs, but it was just a small internal helper and we don't need more info yet so I just used the vecs with `partition_map` to help split them

stbrody (approved):
lgtm

PR #538, opened 2024-09-23
feat: aes 291 continuous queries in olap aggregator

samika98:
The OLAP aggregator will continuously run the data event and time event queries and store the results into Basin. This code here adds logic to run the queries continuously. 

We opted not to utilize  Data Fusion's streaming capabilities, as they lack the features and flexibility required for our use case. Data Fusion's streaming functionality is relatively limited in scope.


olap/src/aggregator/mod.rs
```diff
@@ -102,26 +110,103 @@ pub async fn run(
...
+                Expr::Cast(Cast::new(Box::new(col("data")), DataType::Utf8)).alias("data"),
+                col("previous"),
+            ])?;
+        let batch = conclusion_feed
+            .filter(col("index").gt(lit(self.last_processed_index.get())))?
+            .limit(0, Some(100))?;
+
+        let df = batch.cache().await?;
+        let df_clone = df.clone();
+        process_feed_batch(self.ctx.clone(), df_clone).await?;
```
nathanielc, line 189: nit, since the name `df_clone` doesn't convey any meaning I'd say this should just be `process_feed_batch(self.ctx.clone(), df.clone()).await?;`

olap/src/aggregator/mod.rs
```diff
@@ -102,26 +110,103 @@ pub async fn run(
...
+                col("controller"),
+                col("event_cid"),
+                Expr::Cast(Cast::new(Box::new(col("data")), DataType::Utf8)).alias("data"),
+                col("previous"),
+            ])?;
+        let batch = conclusion_feed
+            .filter(col("index").gt(lit(self.last_processed_index.get())))?
+            .limit(0, Some(100))?;
+
+        let df = batch.cache().await?;
```
nathanielc, line 187: A comment here that we need a cache because we do two passes over the data would be good. First pass is the normal process_feed_batch pass while the second pass is to get the highest index

olap/src/aggregator/mod.rs
```diff
@@ -369,7 +454,38 @@ mod tests {
         Ok(pretty_format_batches(&doc_state)?)
     }
 
-    #[test(tokio::test)]
+    // TODO : Create a test that checks how many times quereies run. They should run continuously in a loop.
+    // #[tokio::test]
+    // async fn test_continuous_stream_processing() -> Result<()> {
```
nathanielc, line 459: Any luck on getting a test working with the custom context?

olap/src/aggregator/mod.rs
```diff
@@ -102,26 +110,103 @@ pub async fn run(
...
+            }
+        }
+    }
     Ok(())
 }
 
+/// Represents a processor for continuous stream processing of conclusion feed data.
+struct ContinuousStreamProcessor {
+    ctx: SessionContext,
+    last_processed_index: Cell<u64>,
```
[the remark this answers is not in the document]
samika98: We do not want ctx to be mutable on the struct. Taking &self can make the API more flexible. I think interior mutability can make the API more flexible. Example : 
```
    let mut processor = ContinuousStreamProcessor::new();
    
    // This works fine
    processor.process_batch(&[Data::new()]);
    
    // But what if we want to do this?
    let processors = vec![&processor, &processor];
    
    for p in &processors {
        p.process_batch(&[Data::new()]); // This won't compile
    }```
    Don't feel too strongly either ways.

nathanielc (approved):

PR #539, opened 2024-09-23
feat: validate time event chain inclusion proofs

dav1do:
The validation service now sets up and interacts with the ethereum RPC providers in order to validate time event inclusion proofs. I refactored some things a bit, like removing a trait in the service crate that isn't necessary currently as we already have one at the RPC provider level. 

The `ValidationRequirement` struct now has a `require_inclusion_proof` option that allows us to store and sync time events over recon even if we can't prove that their valid (e.g. we don't have an RPC provider for the chain). If we can prove they are invalid, we will reject them still. I also made an error enum that is specific to the chain inclusion result to make it easier to process different errors. It doesn't yet handle transient errors very well, which I plan to address in the RPC provider level when I switch the http client to use the alloy crate rather than a simple reqwest wrapper.

### DIFF
--- a/event-svc/src/event/validator/time.rs
+++ b/event-svc/src/event/validator/time.rs
@@ -53,6 +53,7 @@ pub struct Timestamp(i64);
 
 impl Timestamp {
     /// A unix epoch timestamp
+    #[allow(dead_code)]
     pub fn as_unix_ts(&self) -> i64 {
         self.0
     }
@@ -114,7 +115,7 @@ impl TimeEventValidator {
     }
 
     /// Get the CAIP2 Chain IDs that we can validate
-    pub fn supported_chains(&self) -> Vec<caip2::ChainId> {
+    fn _supported_chains(&self) -> Vec<caip2::ChainId> {
         self.chain_providers.keys().cloned().collect()
     }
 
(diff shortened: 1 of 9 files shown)


event-svc/src/event/validator/time.rs
```diff
@@ -79,22 +100,75 @@ impl<'a> EventTimestamper<'a> {
...
+            if block.number < *threshold {
+                return Ok(Timestamp(block.timestamp));
+            } else if event.proof().tx_type() != V1_PROOF_TYPE {
+                return Err(ChainInclusionError::InvalidProof(format!("Any anchor proofs created after block {threshold} for chain {} must include the txType field={V1_PROOF_TYPE}. Anchor txn blockNumber: {}", event.proof().chain_id(), block.number)));
+            }
+        }
+
+        Ok(Timestamp(block.timestamp))
+    }
+
```
dav1do, line 171: delete V0 proof type in `get_root_cid_from_input`

event-svc/src/event/validator/time.rs
```diff
@@ -79,22 +100,75 @@ impl<'a> EventTimestamper<'a> {
...
+        if root_cid != event.proof().root() {
+            return Err(ChainInclusionError::InvalidProof(format!(
+                "the root CID is not in the transaction (root={})",
+                event.proof().root()
+            )));
+        }
+
+        if let Some(threshold) = BLOCK_THRESHHOLDS.get(event.proof().chain_id()) {
+            if block.number < *threshold {
+                return Ok(Timestamp(block.timestamp));
```
dav1do, line 163: return invalid, no v0

stbrody (approved):

PR #540, opened 2024-09-25
feat: store doc_state table in object storage

nathanielc:
With this change the doc_state table is backed by an object store instead of the local filesystem.

We assume an S3 compatible API and credentials.

Fixes AES-284

### DIFF
--- a/olap/src/aggregator/mod.rs
+++ b/olap/src/aggregator/mod.rs
@@ -147,6 +147,7 @@ fn tx_error_to_df(err: tonic::transport::Error) -> DataFusionError {
 //  * have previous CIDs that either already exist in `doc_state` or be contained within the
 //  current conclusion_feed batch,
 //  * be valid JSON patch data documents.
+//  * use a qualified table name of `conclusion_feed`.
 async fn process_feed_batch(ctx: SessionContext, conclusion_feed: DataFrame) -> Result<()> {
     let doc_state = ctx
         .table("doc_state")
@@ -249,9 +250,10 @@ mod tests {
     use cid::Cid;
     use datafusion::{
         common::Constraints,
+        datasource::{provider_as_source, MemTable},
         logical_expr::{
             expr::ScalarFunction, CreateMemoryTable, DdlStatement, EmptyRelation, LogicalPlan,
-            ScalarUDF,
+            LogicalPlanBuilder, ScalarUDF,
         },
     };
     use expect_test::expect;
@@ -321,7 +323,17 @@ mod tests {
         ctx: SessionContext,
         conclusion_feed: RecordBatch,
     ) -> anyhow::Result<impl std::fmt::Display> {
-        let conclusion_feed = ctx.read_batch(conclusion_feed)?;
+        // Setup conclusion_feed table from RecordBatch
+        let provider = MemTable::try_new(conclusion_feed.schema(), vec![vec![conclusion_feed]])?;
+        let conclusion_feed = DataFrame::new(
+            ctx.state(),
+            LogicalPlanBuilder::scan(
+                "conclusion_feed",
+                provider_as_source(Arc::new(provider)),
+                None,
+            )?
+            .build()?,
+        );
         process_feed_batch(ctx.clone(), conclusion_feed).await?;
         let cid_string = Arc::new(ScalarUDF::from(CidString::new()));
         let doc_state = ctx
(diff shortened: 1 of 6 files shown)


stbrody (commented):
should we do both? Store it locally in C1 *and* push it out to S3?

nathanielc:
> should we do both? Store it locally in C1 and push it out to S3? I think we'll need it in C1 soon anyway to build the aggregated document feed for OrbisDB and others.

I don't think so. The only reason to store it in both places would be as a backup/redundancy strategy since you would not be able to dynamically query it from both locations. And it would be a poor backup strategy as there is no control over how it actually does the backup(i.e. frequency etc). And with the data being immutable files in an object store there are much better tools for managing backups.

stbrody:
I guess maybe "both" is less useful, but "either" might still be useful.  I'm trying to think about how folks who don't want the Basin integration, but do still want to use the aggregator from C1 (like OrbisDB) will do so.  They'll still need access to the aggregated document states somewhere.

nathanielc:
Agreed both would be a good feature however I'd say its out of scope for the MVP as its still unclear if we are going to actually bring the aggregator back into ceramic-one directly or not. That decision will impact how we store data locally for the aggregator.

olap/src/aggregator/mod.rs
```diff
@@ -56,10 +68,11 @@ pub async fn run(
     let listing_options =
         ListingOptions::new(Arc::new(file_format)).with_file_extension(".parquet");
 
-    // TODO place this table in object store, see AES-284
+    // Set the path within the bucket for the doc_state table
+    url.set_path("/ceramic/v0/doc_state/");
```
samika98, line 72: Can we get the String from a const?

olap/src/aggregator/mod.rs
```diff
@@ -94,7 +107,7 @@ pub async fn run(
                     true,
                 )),
                 Arc::new(Field::new("event_cid", DataType::Binary, false)),
-                Arc::new(Field::new("state", DataType::Utf8, false)),
+                Arc::new(Field::new("state", DataType::Utf8, true)),
```
samika98, line 110: Why are we changing the nullable to be true here?

olap/src/aggregator/mod.rs
```diff
@@ -173,36 +175,12 @@ async fn process_feed_batch(ctx: SessionContext, conclusion_feed: DataFrame) ->
...
-                relation: Some(TableReference::from("?table?")),
-                name: "dimensions".to_string(),
-            })
-            .alias("dimensions"),
-            col(Column {
-                relation: Some(TableReference::from("?table?")),
-                name: "event_cid".to_string(),
-            })
-            .alias("event_cid"),
+            col("conclusion_feed.index").alias("index"),
```
samika98, line 178: Thanks for this. Did the exact same thing in my test haha. Was getting an ambiguous reference error!

olap/src/aggregator/mod.rs
```diff
@@ -344,7 +323,17 @@ mod tests {
         ctx: SessionContext,
         conclusion_feed: RecordBatch,
     ) -> anyhow::Result<impl std::fmt::Display> {
-        let conclusion_feed = ctx.read_batch(conclusion_feed)?;
+        // Setup conclusion_feed table from RecordBatch
+        let provider = MemTable::try_new(conclusion_feed.schema(), vec![vec![conclusion_feed]])?;
+        let conclusion_feed = DataFrame::new(
+            ctx.state(),
+            LogicalPlanBuilder::scan(
```
samika98, line 330: Why do we need a separate logical plan builder for scan? What does this give us extra which the DDL does not?

samika98 (changes requested):
Looks good! Thanks for adding the aliases in the join! Just a question about adding a separate logical plan for scan, why do we need it?

olap/src/aggregator/mod.rs
nathanielc: We need this in order to specify the name of the table. Otherwise it gets the `?table?` name.

olap/src/aggregator/mod.rs
nathanielc: Ran into this with some real world bad data. If the events are broken this allows us to continue to process other events and just use null as the state for the broken ones.

samika98 (approved):

PR #541, opened 2024-09-26
feat: support non sharded IPFS block paths

nathanielc:
When migrating from IPFS there are two possible directory structures, a sharded and a non sharded form. This change now supports both defaulting to the sharded form.

### DIFF
--- a/one/src/migrations.rs
+++ b/one/src/migrations.rs
@@ -51,6 +51,11 @@ pub struct FromIpfsOpts {
     #[arg(long, env = "CERAMIC_ONE_LOCAL_NETWORK_ID")]
     local_network_id: Option<u32>,
 
+    /// Expect non sharded paths to blocks.
+    /// Sharded paths are organized into a two character prefix directories.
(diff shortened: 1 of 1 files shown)


one/src/migrations.rs
```diff
@@ -51,6 +51,11 @@ pub struct FromIpfsOpts {
     #[arg(long, env = "CERAMIC_ONE_LOCAL_NETWORK_ID")]
     local_network_id: Option<u32>,
 
+    /// Expect non sharded paths to blocks.
+    /// Sharded paths are organized into a two character prefix directories.
+    #[arg(long, env = "CERAMIC_ONE_NON_SHARDED_PATHS", default_value_t = false)]
+    non_sharded_paths: bool,
```
m0ar, line 57: FWIW, the sharding/key transformations are a configurable property of the blockstore and not a binary choice. But this should cover the "didn't actively configure sharding" case, and the most common option I guess.

https://github.com/ipfs/kubo/blob/master/docs/datastores.md

PR #544, opened 2024-09-30
feat: protect FlightSQL server with sql options.

nathanielc:
Disables all types of SQL queries except read only queries.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -88,12 +88,11 @@ crossterm = "0.25"
 ctrlc = "3.2.2"
 dag-jose = "0.2"
 datafusion = "42"
-#datafusion-federation = "0.2"
-#datafusion-flight-sql-server = "0.2"
-#datafusion-flight-sql-table-provider = { git = "https://github.com/datafusion-contrib/datafusion-federation.git", branch = "main" }
-datafusion-federation = { git = "https://github.com/datafusion-contrib/datafusion-federation.git", branch = "main" }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -88,12 +88,11 @@ crossterm = "0.25"
 ctrlc = "3.2.2"
 dag-jose = "0.2"
 datafusion = "42"
-#datafusion-federation = "0.2"
-#datafusion-flight-sql-server = "0.2"
-#datafusion-flight-sql-table-provider = { git = "https://github.com/datafusion-contrib/datafusion-federation.git", branch = "main" }
-datafusion-federation = { git = "https://github.com/datafusion-contrib/datafusion-federation.git", branch = "main" }
-datafusion-flight-sql-server = { git = "https://github.com/datafusion-contrib/datafusion-federation.git", branch = "main" }
-datafusion-flight-sql-table-provider = { git = "https://github.com/datafusion-contrib/datafusion-federation.git", branch = "main" }
+datafusion-federation = { git = "https://github.com/nathanielc/datafusion-federation.git", features = [
```
dav1do, line 91: A comment about which PR needs to land so we can switch to the main repo might be nice but not a big deal.

dav1do (approved):

Cargo.toml
nathanielc: PR landed will update to main now... but good call

PR #545, opened 2024-09-30
feat: add query command to ceramic-one

nathanielc:


dav1do (approved):

AaronGoldman (approved):

stbrody (commented):
is this going over http or via direct access to the underlying data store (sqlite)?  I believe it's the latter - if so, can it run at the same time as another C1 process is running in Daemon mode against the same data?

I feel like a query REPL would be best pointed against a running C1 daemon over the network using the Flight APIs, rather than by direct access to the local data files.

nathanielc:
> is this going over http or via direct access to the underlying data store (sqlite)? 

Good question, its neither. As written today it cannot access the data. When I demoed it last week I pointed it at my S3 bucket and then queried the data that way. 

To me that means we should add a bit to configure your S3 bucket as part of the query command so that its ready to go. And in that configuration it will not compete with the sqlite file or any other local resource. 

I'll change this to a draft until I can get to that change.

stbrody:
> To me that means we should add a bit to configure your S3 bucket as part of the query command so that its ready to go

Oh interesting, so then this wouldn't really be querying data in a Ceramic node at all, it'd be querying data in Parquet that was exported out from Ceramic via the OLAP aggregator?  TBH that feels like a weird thing to include in the Ceramic CLI since at that point it kind of isn't really Ceramic data anymore?

nathanielc:
>  TBH that feels like a weird thing to include in the Ceramic CLI since at that point it kind of isn't really Ceramic data anymore?

As we go down the path that more of the Ceramic data lives in parquet then this fits naturally. Additionally we are leaning towards bringing the OLAP aggregator into C1 and so such a distinction won't be meaningful to end users.

stbrody:
I imagine that we're going to want a way for users to query the data that actually lives inside the ceramic node no?  I would have assumed that that's what a `query` command in the ceramic CLI would do.  If we wind up both with a way to query data within the Ceramic node as well as data that has been exported from the Ceramic to parquet in S3, then I just want to make sure it isn't confusing to users which is which.  Maybe it's just a matter of thinking carefully about the names of the config flags for this command.

nathanielc:
@stbrody PR has been updated to auto configure connecting to the `conclusion_feed` table over flight sql and to the `doc_state` table over S3.

one/src/query.rs
```diff
@@ -0,0 +1,56 @@
...
+#[derive(Args, Debug)]
+pub struct QueryOpts {
+    /// Endpoint of a Flight SQL server for the conclusion feed.
+    #[arg(
+        short,
+        long,
+        default_value = "http://127.0.0.1:5102",
+        env = "CERAMIC_ONE_FLIGHT_SQL_ENDPOINT"
+    )]
+    flight_sql_endpoint: String,
```
stbrody, line 14: I feel like the fact that it uses Flight as the API format is an implementation detail, the important thing is that this is a C1 node at the other end of this endpoint right?  Should this be called ceramic_one_endpoint or something like that?

one/src/query.rs
```diff
@@ -0,0 +1,56 @@
...
+    ///
+    ///   * AWS_ACCESS_KEY_ID -> access_key_id
+    ///   * AWS_SECRET_ACCESS_KEY -> secret_access_key
+    ///   * AWS_DEFAULT_REGION -> region
+    ///   * AWS_ENDPOINT -> endpoint
+    ///   * AWS_SESSION_TOKEN -> token
+    ///   * AWS_ALLOW_HTTP -> set to "true" to permit HTTP connections without TLS
+    ///
+    #[arg(long, env = "CERAMIC_ONE_AWS_BUCKET")]
+    aws_bucket: String,
```
stbrody, line 29: kind of the inverse of my above comment.  AWS is a specific platform, one we are notably not using.  Should this say be s3_bucket or something like that?  We're using the S3 API, but we aren't using the AWS platform...

AaronGoldman:
I think we should have a http Api endpoint to run queries with your node so that all the data in the SQLite file and some data exported to parquet files can query identically. By querying with c1 you should not need to understand the storage location but obviously if I want to use DuckDB(bring your own query engine) then you will need to tell those tools where to find the state store.

AaronGoldman:
I think we should have a http Api endpoint to run queries with your node so that all the data in the SQLite file and some data exported to parquet files can query identically. By querying with c1 you should not need to understand the storage location but obviously if I want to use DuckDB(breing your own query engine) then you will need to tell those tools where to find the state store.

stbrody (approved):
lgtm mod questions about the argument names

stbrody:
> I think we should have a http Api endpoint to run queries with your node

Isn't that what the flight sql endpoint already is?

one/src/query.rs
nathanielc: Yes, however ceramic-one exposes multiple endpoints(p2p, REST, flight sql). Maybe we call it `ceramic_one_query_endpoint`? As its purpose is to expose query access to the data?

nathanielc:
> > I think we should have a http Api endpoint to run queries with your node
> 
> Isn't that what the flight sql endpoint already is?

In my opinion it is. Its not an HTTP/1 endpoint but trying to build an HTTP/1 endpoint to expose query access to the data will mean we basically reimplement flight sql.

one/src/query.rs
stbrody: yep I like that

nathanielc: Using `ceramic_one_query_endpoint` meant the env var was `ceramic_one_ceramic_one_query_endpoint` which is horrible. So I went with just `query_endpoint` as the `ceramic_one` context is already present

PR #547, opened 2024-10-01
feat: Ethereum RPC provider configuration

stbrody:


dav1do (approved):
:shipit: I think we'll need to come back to the local/in-memory stuff but this should be a good start. We'll connect to the chain even if we're not validating events but I'm okay with that.

one/src/daemon.rs
```diff
@@ -231,6 +231,53 @@ pub struct DaemonOpts {
...
+                    .supported_chain_ids()
+                    .map_or(true, |ids| ids.contains(provider_chain))
+                {
+                    providers.push(provider);
+                } else {
+                    warn!("Eth RPC provider {} uses chainid {} which isn't supported by Ceramic network {:?}", url, provider_chain,network);
+                }
+            }
+            Err(err) => {
+                warn!("failed to create RCP client with url: '{url}': {err}");
```
samika98, line 271: nit : RPC *

samika98 (approved):

one/src/lib.rs
```diff
@@ -93,6 +94,53 @@ impl Network {
...
+                Ok(vec![
+                    "https://ethereum-sepolia-rpc.publicnode.com".to_string()
+                ])
+            }
+            Network::Local => {
+                info!(
+                "using default Ganache Ethereum RPC URL for Local network: http://localhost:7545"
+            );
+                // Default Ganache port
+                Ok(vec!["http://localhost:8545".to_string()])
```
dav1do, line 119: I think this is failing CI because we have to address by container name. So we probably need to update the tests to use http://ganache:8545... a bit of a chicken and egg situation with the cli flag but we should be able to use the env var for now and it will be ignored before this is in.

one/src/daemon.rs
dav1do: Fixed in [543](https://github.com/ceramicnetwork/rust-ceramic/pull/543)

PR #550, opened 2024-10-04
refactor: modify eth rpc trait for other chains

dav1do:
I modified the EthRpc trait exposed by the validation crate to hide more chain information and return the specific information needed to do time event validation (a timestamp and a root CID). The parsing and understanding of various transaction input related information moved from the service to the validation crate. A lot of code was moved, some was deleted/renamed (specifically `EthRpc` to `ChainInclusion` trait and the related inputs/outputs). The `ChainInclusion` trait has an associated type for the input so chains without the same Eth RPC interface can request different information.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ sync_test.sh
 openapi-generator-cli.jar
 .env
 ceramic_cicddb.sqlite*
+event-svc/model_error_counts.csv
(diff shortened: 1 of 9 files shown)


stbrody (approved):
looks good to me, but I put together a follow-up PR to push things a little bit farther and hopefully avoid the event validation code needing to make any decisions at all based on which chain the anchor proof is on: https://github.com/ceramicnetwork/rust-ceramic/pull/570

nathanielc (approved):

PR #551, opened 2024-10-04
chore: record the per item duration of an insert_many request

dav1do:
We insert in batches that  are typically 3-100 items (depending if it's api or recon). This gives a better sense of how long each item takes in comparison to the duration of the entire request. It's kind of a clunky way to do this, but I didn't want to return a duration from the other function and then record a second item, but maybe that would be cleaner.

### DIFF
--- a/event-svc/src/store/metrics.rs
+++ b/event-svc/src/store/metrics.rs
@@ -114,6 +114,34 @@ impl<S: Send + Sync> StoreMetricsMiddleware<S> {
         metrics.record(&event);
         ret
     }
+
+    /// Record metrics for the duration of given a given API endpoint
+    /// and then record a second entry with the duration for an individual item.
+    async fn record_per_item<T>(
+        metrics: &Metrics,
+        name: &'static str,
+        name_per_item: &'static str,
+        count: u64,
+        fut: impl Future<Output = T>,
+    ) -> T {
+        let start = Instant::now();
+        let ret = fut.await;
+        let duration = start.elapsed();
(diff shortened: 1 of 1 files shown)


event-svc/src/store/metrics.rs
```diff
@@ -114,6 +114,34 @@ impl<S: Send + Sync> StoreMetricsMiddleware<S> {
...
+    /// and then record a second entry with the duration for an individual item.
+    async fn record_per_item<T>(
+        metrics: &Metrics,
+        name: &'static str,
+        name_per_item: &'static str,
+        count: u64,
+        fut: impl Future<Output = T>,
+    ) -> T {
+        let start = Instant::now();
+        let ret = fut.await;
```
gvelez17, line 128: no limit on how long to wait?

dav1do: No, not right now. This is middleware so we could add a max here but adding a timeout above at the API/recon/migrations level gives the caller more control (we do have one for API currently). That loses this timing and shows up as 500 errors instead. 

Right now this is wrapping a filesystem database so we're in a really bad state if we can't access it in a reasonable time. I'm not sure what a good timeout would be.. if this is taking a long time, someone asked the node to do something slow and other writes will be blocked in the meantime (or we can't access the file system and aborting the request isn't likely to help).

stbrody:
> @smrz2001 @samika98 this is to monitor the queue sizes for gitcoin - maybe we can get it onto the gitcoin node after integration tests?

@gvelez17 this PR is to rust-ceramic, which gitcoin doesn't use

nathanielc (approved):

PR #553, opened 2024-10-07
fix: subscribe to server shutdown signal in insert task

dav1do:
We previously relied on http server shutdown closing the channel to signal the task to stop, which could take at least one tick interval of the task to trigger. Now, we listen to the daemon shutdown signal so it's faster.

### DIFF
--- a/api/src/server.rs
+++ b/api/src/server.rs
@@ -410,10 +410,10 @@ where
             let mut shutdown = false;
             loop {
                 let mut buf = Vec::with_capacity(EVENTS_TO_RECEIVE);
-                let mut process_early = false;
+                let mut process_interval = false;
                 tokio::select! {
                     _ = interval.tick() => {
-                        process_early = true;
+                        process_interval = true;
                     }
                     val = event_rx.recv_many(&mut buf, EVENTS_TO_RECEIVE) => {
                         if val > 0 {
@@ -429,6 +429,8 @@ where
                     tracing::info!("Shutting down insert task after processing current batch");
                     if !event_rx.is_empty() {
                         let remaining_event_cnt = event_rx.len();
+                        // the buffer above gets moved into the select! even though we don't
+                        // follow that branch in this case, so we create a new buffer here
                         let mut buf = Vec::with_capacity(remaining_event_cnt);
                         tracing::info!(
                             "Receiving {remaining_event_cnt} remaining events for insert task before exiting"
@@ -443,7 +445,7 @@ where
                 // process events at the interval or when we're under heavy load.
                 // we do it outside the select! to avoid any cancel safety issues
                 // even though we should be okay since we're using tokio channels/intervals
-                if events.len() >= EVENT_INSERT_QUEUE_SIZE || process_early {
+                if events.len() >= EVENT_INSERT_QUEUE_SIZE || process_interval {
                     Self::process_events(&mut events, &event_store, node_id).await;
                 }
             }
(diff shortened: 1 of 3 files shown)


api/src/server.rs
```diff
@@ -399,7 +407,17 @@ where
             // rely on the channel depth for backpressure. the goal is to keep the queue close to empty
             // without processing one at a time. when we stop parsing the carfile in the store
             // i.e. validate before sending here and this is just an insert, we may want to process more at once.
+            let mut shutdown = false;
             loop {
+                let should_exit = shutdown || event_rx.is_closed();
+                // make sure the events queue doesn't get too deep when we're under heavy load
```
stbrody, line 413: nitpick, it confused me to have this code come in the middle of the shutdown logic.  Can we put line 412 and the block starting on line 417 next to each other?  Ideally as the first thing in the loop?

stbrody (commented):
core change looks good, but unit tests are failing to build

api/src/server.rs
dav1do: reworked the logic a bit to stop using channel openness at all or share anything between shutdown and normal processing. I added draining the channel before shutting down. lmk what you think.

api/src/server.rs
```diff
@@ -399,27 +407,45 @@ where
...
+                        tracing::debug!("Insert many task got shutdown signal");
+                        shutdown = true;
+                    }
+                };
                 if shutdown {
-                    tracing::info!("Shutting down insert task.");
+                    tracing::info!("Shutting down insert task after processing current batch");
+                    if !event_rx.is_empty() {
+                        let remaining_event_cnt = event_rx.len();
+                        let mut buf = Vec::with_capacity(remaining_event_cnt);
```
stbrody, line 432: can we not just use the buf that was already allocated on line 412?

api/src/server.rs
```diff
@@ -399,27 +407,45 @@ where
...
+                        events.extend(buf);
+                    }
+
+                    Self::process_events(&mut events, &event_store, node_id).await;
                     return;
                 }
+                // process events at the interval or when we're under heavy load.
+                // we do it outside the select! to avoid any cancel safety issues
+                // even though we should be okay since we're using tokio channels/intervals
+                if events.len() >= EVENT_INSERT_QUEUE_SIZE || process_early {
```
stbrody, line 446: what's "early" about the processing when `process_early` is true?  Isn't the "early" processing really when the events.len() is too long, but the processing when that bool is true is actually the regularly scheduled processing due to the interval ticking?

stbrody (commented):
there's enough weird rust and tokio things going on here that I feel like it's probably good if @nathanielc takes a look rather than me just approving it.

api/src/server.rs
dav1do: Fair. Process interval is a better name. I’ll rename it. It was early in my head because we want to get N requests before we do work so if we don’t get enough in time we process them anyway before that (ie early) but it’s not very clear.

api/src/server.rs
dav1do: Yeah, I makes sense. We can resize and use it (or just use it since recv_many might resize automatically - I’ll verify). It’s the last allocation before exiting so I didn’t think too closely about it.

dav1do: We can't use it actually. it gets moved into the select branch even though we don't follow that arm in this case. I added a comment to that effect. expanding the macro might make it easier to see but I didn't bother.

nathanielc (approved):

PR #558, opened 2024-10-08
fix: include unique in dimensions

nathanielc:
Turns out the unique field can contain useful information to query for example a stream id or a did.  This change exposes it in dimensions. 

Closes AES-402


event-svc/src/tests/event.rs
```diff
@@ -687,19 +687,19 @@ async fn test_conclusion_events_since() -> Result<(), Box<dyn std::error::Error>
...
-        +-------+------------+---------------------------------------------------------------+-------------+----------------------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+---------------------------------------------------------------+--------------------------------------------+----------------------------------------------------------------------+
-        | 0     | Data       | bagcqcerahx5i27vqxigdq3xulceu5qv6yzdvxzamfueubsyxam5kmjcpp45q | 3           | did:key:z6Mkk3rtfoKDMMG4zyarNGwCQs44GSQ49pcYKQspHJPXSnVw | [model: "ce0102015512204bf5122f344554c53bde2ebb8cd2b7e3d1600ad631c385a5d7cce23c7785459a", controller: "6469643a6b65793a7a364d6b6b337274666f4b444d4d47347a7961724e4777435173343447535134397063594b517370484a5058536e5677", context: ""] | bagcqcerahx5i27vqxigdq3xulceu5qv6yzdvxzamfueubsyxam5kmjcpp45q | 6e756c6c                                   | []                                                                   |
```
samika98, line 693: Can we have index starting with 1 here?

nathanielc: @samika98 The test does not control the index. It simply inserts the events into the store and then queries them backout. Has your change to start from 1 merged? If so then we likely still have a bug somewhere.

flight/src/types.rs
```diff
@@ -134,6 +134,14 @@ impl TryFrom<unvalidated::Event<Ipld>> for ConclusionInit {
                     init_payload
                         .header()
                         .context()
+                        .map(|context| context.to_vec())
+                        .unwrap_or_default(),
+                ),
+                (
+                    "unique".to_string(),
```
samika98, line 141: Should we have a test that shows that we can query over unique?

samika98: move unique to consts?

samika98 (approved):
We have a test which shows that we are including unique in dimension on the conclusion feed. Should we have a  test which checks querying over the "unique" in the conclusion feed ?

nathanielc:
@stbrody Brought up a good point that the value of the unique dimension is already inside the content (i.e. data ) of the event. So the question is do we need both the dimenstion and the raw data? Th dimension makes is straightforward to query over the unique value, which is the purpose of the dimensions? I don't see much down side to including it and since its valuable to include it I think we should. Thoughts?

stbrody:
my inclination would be to leave it out until/unless someone asks for it.  Always easier to add in later than to remove.  For now I'd err on the side of keeping the API clean and avoid noise if we don't need it.

stbrody:
but I also don't feel that strongly and could easily be swayed

nathanielc:
After some discussion we decided its not immediately useful to be able to query the unique dimension. We will need a system to be able to query any kind of relation and only set account relations are queryable via the unique dimension. As such exposing unique to be a dimension creates a different path for querying one of the types of relations and not the others. Therefore we are not going to expose unique dimensions until we have a clear use case for it and have designed how relations generally will be queried.

PR #561, opened 2024-10-10
fix: use correct index for the conclusion feed.

nathanielc:
We found a few bugs in the conclusion feed implementation on the event
service. Fixes will follow, however this change updates the test to make
the bugs obvious.

Bugs:
1. Indexes are duplicated
2. Unsigned event indexes are always zero
3. Off by one bug in the bounds of the highwater mark

The first commit in this change updates the test to expose all three bugs. The following commit each fix one of the bugs.

Turns out the delivered insertion was not test friendly. I added two more refactor commits that make the delivered counter isolated per EventService and another to rename some types whose names had become less clear because of other refactors.


event-svc/src/event/service.rs
```diff
@@ -403,7 +403,7 @@ impl EventService {
                         .map_err(|e| {
                             Error::new_app(anyhow::anyhow!("Failed to serialize IPLD data: {}", e))
                         })?,
-                    index: 0,
+                    index: delivered as u64,
```
samika98, line 406: !!

samika98: Tests should have caught this

nathanielc: No test included an unsigned init event. That is why I added it to the generated data.

event-svc/src/store/sql/access/event.rs
```diff
@@ -370,17 +371,30 @@ impl CeramicOneEvent {
             all_blocks.iter().map(|row| row.block.clone()).collect(),
         )
         .await?;
+
+        // We need to match up the delivered index with each event. However all_blocks contains an
```
samika98, line 397: We are keeping multiple copies of data in memory, can we move this logic in the SQL query itself? 
``` 
    let max_highwater = sqlx::query_as::<_, DeliveredEventBlockRow>(
        r#"
        SELECT
            event_id,
            MAX(delivered) AS max_delivered,
            block_data
        FROM
            events_table
        WHERE
            delivered > ?  -- Only fetch events with a delivered value greater than the highwater mark
        GROUP BY
            event_id
        ORDER BY
            max_delivered ASC
        LIMIT ?
        "#,
    ) 
    ```
    What do you think? Is this a better approach?

Comment (samika98) [reply event-svc/src/store/sql/access/event.rs:397]: Then we can call, 
```
.bind(highwater)
.bind(limit)
.fetch_all(pool.reader())
.await?
.into_iter()
.map(|row| {
        // Decode the CAR file from the block data
     ....

        // Add the parsed event to the list
     ....

      // Update the max highwater mark
    })
    .max()
    .unwrap_or(highwater); // Use the input highwater if no new events are found ```

event-svc/src/event/feed.rs
```diff
@@ -33,7 +33,9 @@ impl ConclusionFeed for EventService {
         limit: i64,
     ) -> anyhow::Result<Vec<ConclusionEvent>> {
         let raw_events = self
-            .fetch_events_since_highwater_mark(highwater_mark, limit)
+            // TODO: Can we make highwater_marks zero based?
```
samika98, line 36: Is this the correct behavior? If we want to fetch events since a highwatermark, we would get events since highwater_mark + 1

samika98:
Thank you for the detailed commit messages and explainations! Reviewing this code commit by commit was fun

event-svc/src/store/sql/access/event.rs
nathanielc: The challenge is that we need events, but we have to construct the events from the list of blocks that make up each event. So the `all_block` query needs to join with the blocks tables in order to get all the data. This PR doesn't change how much memory is needed (we were previously keeping a copy of the data in memory). I think optimizing this query should be future work.(If we find its a cause of slow down)

event-svc/src/event/feed.rs
nathanielc: Yes without this change you always get back the last row from the previous batch as the first row in the next batch.

nathanielc: If you look in the code for how the highwater mark is made you can see it already is adding one to it. This is what I mean by saying they are one based. Hence a TODO for later to investigate if we can simplify things to make everything zero based.

samika98 (approved):
LGTM!

PR #562, opened 2024-10-15
feat: turn on event validation by default

samika98:
Do not merge - for now
I am collecting evidence that this feature has been stable on durable for over a 1 week

### DIFF
--- a/one/src/daemon.rs
+++ b/one/src/daemon.rs
@@ -14,7 +14,7 @@ use ceramic_interest_svc::InterestService;
 use ceramic_kubo_rpc::Multiaddr;
 use ceramic_metrics::{config::Config as MetricsConfig, MetricsHandle};
 use ceramic_p2p::{load_identity, DiskStorage, Keychain, Libp2pConfig};
-use clap::Args;
+use clap::{ArgAction, Args};
 use recon::{FullInterests, Recon, ReconInterestProvider};
 use signal_hook::consts::signal::*;
 use signal_hook_tokio::Signals;
@@ -164,8 +164,15 @@ pub struct DaemonOpts {
     )]
     authentication: bool,
 
-    /// Enable event validation; Requires using the experimental-features flag
-    #[arg(long, default_value_t = true, env = "CERAMIC_ONE_EVENT_VALIDATION")]
+    /// Enable event validation
+    #[arg(
+        long,
+        default_value = "true",
+        value_parser = clap::value_parser!(bool),
+        num_args = 0..=1,
+        action = ArgAction::Set,
+        env = "CERAMIC_ONE_EVENT_VALIDATION"
+    )]
     event_validation: bool,
 
     /// Flight SQL bind address; Requires using the experimental-features flag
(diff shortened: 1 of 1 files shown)


AaronGoldman (approved):

nathanielc (commented):
LGTM, but can you test if you can still disable it? I have had trouble passing false to boolean args in clap. 

Also can you remove the language about the experimental flag being required?

samika98:
> LGTM, but can you test if you can still disable it? I have had trouble passing false to boolean args in clap.
> 
> Also can you remove the language about the experimental flag being required?

@nathanielc tested it and turns out clap is weird with booleans. See this discussion : https://github.com/clap-rs/clap/issues/1649#issuecomment-1144331401. Turns out we need to explicitly pass the boolean and set the action so that we can set the variable. 
So the current code leads us with : 
'event-validation' = None -> gets set to true
'event-validation' in the args can hold either true/false.
'-- event-validation = false' -> gets set to false
'--event-validation = true' -> gets set to true

one/src/daemon.rs
```diff
@@ -164,12 +164,14 @@ pub struct DaemonOpts {
...
-        default_value_t = false,
-        env = "CERAMIC_ONE_EVENT_VALIDATION",
-        requires = "experimental_features"
+        default_value = "true",
+        value_parser = clap::value_parser!(bool),
+        num_args = 0..=1,
+        action = ArgAction::Set,
+        env = "CERAMIC_ONE_EVENT_VALIDATION"
     )]
     event_validation: bool,
```
nathanielc, line 176: Re booleans being weird in clap. I think all we need to do re the issue you linked is set the to `Option<bool>` and then default it to true in our code like, `opts.event_validation.unwrap_or(true)`.

samika98: Yeah! we make it optional and don'e set a default value. Seems to work : 
```samikas@Samikas-MacBook-Pro rust-ceramic % RUST_LOG=info,ceramic_olap=debug cargo run --bin ceramic-one -- daemon | grep "Event service created"
warning: unused import: `ArgAction`
  --> one/src/daemon.rs:17:12
   |
17 | use clap::{ArgAction, Args};
   |            ^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: `ceramic-one` (lib) generated 1 warning (run `cargo fix --lib -p ceramic-one` to apply 1 suggestion)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.51s
     Running `target/debug/ceramic-one daemon`
  2024-10-22T18:08:19.257286Z  INFO ceramic_one::daemon: Event service created, event_validation: true
^C
samikas@Samikas-MacBook-Pro rust-ceramic % RUST_LOG=info,ceramic_olap=debug cargo run --bin ceramic-one -- daemon --event-validation=false | grep "Event service created"
warning: unused import: `ArgAction`
  --> one/src/daemon.rs:17:12
   |
17 | use clap::{ArgAction, Args};
   |            ^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: `ceramic-one` (lib) generated 1 warning (run `cargo fix --lib -p ceramic-one` to apply 1 suggestion)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.50s
     Running `target/debug/ceramic-one daemon --event-validation=false`
  2024-10-22T18:08:39.913484Z  INFO ceramic_one::daemon: Event service created, event_validation: false```

samika98: Added setting the default value using default_value. Added reasoning in the commit message incase anyone is interested in the workings of default_value vs default_value_t, why one works and the other does not

nathanielc (approved):

PR #563, opened 2024-10-16
fix: use tokio::time::Interval for scheduling anchor batches

nathanielc:
This leverages the time math of skipping missed ticks etc. We are only left with computing the time when we should start the first tick. For this we use chrono and then rely on Interval for the heavy lifting.

### DIFF
--- a/anchor-service/src/anchor_batch.rs
+++ b/anchor-service/src/anchor_batch.rs
@@ -9,7 +9,7 @@ use futures::pin_mut;
 use indexmap::IndexMap;
 use std::future::Future;
 use std::{sync::Arc, time::Duration};
-use tokio::time::{interval_at, Instant, MissedTickBehavior};
+use tokio::time::{interval_at, Instant, Interval, MissedTickBehavior};
 use tracing::{error, info};
 
 use crate::high_water_mark_store::HighWaterMarkStore;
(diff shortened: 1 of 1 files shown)


anchor-service/src/anchor_batch.rs
```diff
@@ -76,28 +77,42 @@ impl AnchorService {
         pin_mut!(shutdown_signal);
 
         info!("anchor service started");
+        let period =
```
stbrody, line 80: nitpick, but could we pull all this logic for building the interval into a helper function to improve readability?

anchor-service/src/anchor_batch.rs
```diff
@@ -76,28 +77,42 @@ impl AnchorService {
...
+            TimeDelta::from_std(self.anchor_interval).expect("anchor interval should be in range");
+        let buffer = period / 12;
```
stbrody, line 82: what is buffer doing?

stbrody (approved):

anchor-service/src/anchor_batch.rs
nathanielc: its the buffer time we give ourselves to actually process the batch. I'll add a comment. For 1hour it ends up being 5 minutes.

smrz2001 (approved):
🚀

PR #565, opened 2024-10-18
fix: size 1 anchor trees

AaronGoldman:


anchor-service/src/cas_mock.rs
```diff
@@ -75,7 +75,7 @@ impl Store for MockAnchorEventService {
         highwater: i64,
         limit: i64,
     ) -> Result<Vec<AnchorRequest>> {
-        let ints = ((highwater as u64)..self.anchor_req_count).take(limit as usize);
+        let ints = ((highwater as u64 + 1)..=self.anchor_req_count).take(limit as usize);
```
AaronGoldman, line 78: this is why all the test vectors channged

nathanielc: Can you add a comment why we are adding 1 to the water mark?

nathanielc (approved):
LGTM with a comment explaing the off by one.

AaronGoldman:
The integration tests didn't have an event with high water mark `0`

anchor-service/src/anchor.rs
```diff
@@ -191,15 +191,17 @@ impl TimeEventBatch {
...
-        for part in path.split('/') {
-            let merkle_node = merkle_nodes
-                .get(&current_node_cid)
-                .ok_or_else(|| anyhow!("missing merkle node for CID: {}", current_node_cid))?;
-            blocks.push(merkle_node.clone());
-            current_node_cid = match part {
-                "0" => merkle_node[0].context("missing left node")?,
-                "1" => merkle_node[1].context("missing right node")?,
-                _ => return Err(anyhow!("invalid path part in time event path: {}", part)),
+        if !path.is_empty() {
```
stbrody, line 194: huh, your answer to my question about the integration tests makes me think I don't actually understand what the bug was.  Wasn't this change here the actual bug fix?  What does the high water mark have to do with this?

PR #570, opened 2024-10-21
refactor: make ChainInclusion trait fully generic and change input to the raw AnchorProof

stbrody:
builds on https://github.com/ceramicnetwork/rust-ceramic/pull/550 making things more fully generic and preparing for the addition of an "adm" implementation of the `ChainProvider` trait.

### DIFF
--- a/event-svc/src/event/validator/event.rs
+++ b/event-svc/src/event/validator/event.rs
@@ -7,7 +7,7 @@ use ipld_core::ipld::Ipld;
 use recon::ReconItem;
 use tokio::try_join;
 
-use crate::event::validator::EthRpcProvider;
+use crate::event::validator::ChainInclusionProvider;
 use crate::store::EventAccess;
 use crate::{
     event::{
@@ -131,7 +131,7 @@ impl EventValidator {
     /// Create a new event validator
     pub async fn try_new(
         event_access: Arc<EventAccess>,
-        ethereum_rpc_providers: Vec<EthRpcProvider>,
+        ethereum_rpc_providers: Vec<ChainInclusionProvider>,
     ) -> Result<Self> {
         let time_event_verifier = TimeEventValidator::new_with_providers(ethereum_rpc_providers);
 
@@ -226,7 +226,7 @@ impl EventValidator {
         Ok(validated_events)
     }
 
-    /// Transforms the [`ChainInclusionError`] into a [`ValidationError`] with an appropriate message
+    /// Transforms the [`eth_rpc::Error`] into a [`ValidationError`] with an appropriate message
     fn convert_inclusion_error(err: eth_rpc::Error, order_key: &EventId) -> ValidationError {
         match err {
             eth_rpc::Error::TxNotFound { chain_id, tx_hash } => {
--- a/event-svc/src/event/validator/time.rs
+++ b/event-svc/src/event/validator/time.rs
@@ -5,9 +5,7 @@ use ceramic_core::ssi::caip2;
 use ceramic_event::unvalidated;
 use tracing::warn;
 
-use ceramic_validation::eth_rpc::{
-    self, ChainInclusion, EthProofType, EthTxProofInput, HttpEthRpc,
-};
+use ceramic_validation::eth_rpc::{self, ChainInclusion, HttpEthRpc};
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct Timestamp(u64);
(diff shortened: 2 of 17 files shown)


validation/src/blockchain/eth_rpc/types.rs
```diff
@@ -101,36 +93,25 @@ impl FromStr for EthProofType {
...
-    /// The input format needed to do the inclusion proof
-    type InclusionInput;
-
     /// Get the CAIP2 chain ID supported by this RPC provider
     fn chain_id(&self) -> &caip2::ChainId;
 
     /// Get the block chain transaction if it exists with the block timestamp information
-    async fn chain_inclusion_proof(&self, input: &Self::InclusionInput)
-        -> Result<TimeProof, Error>;
+    async fn chain_inclusion_proof(&self, input: &AnchorProof) -> Result<TimeProofResults, Error>;
```
stbrody, line 114: this is the core of the change in this PR.  Now instead of having the `ChainInclusion` type by generic over an `InclusionInput` type that must be specified for each implementation of the trait, now we just simply take the raw `AnchorProof` object and let the implementations decide how to interpret that object.

This avoids the callers of this function needing to know whether the AnchorProof is for an ethereum anchor or an Adm anchor and having to build up the appropriate input type accordingly.  Now the caller can just blindly pass the anchor proof down into this function and trust the underlying implementation to do the right thing.

The goal is that other than the initial setup to construct the `ChainInclusionProviders` from the given RPC urls at startup, none of the event validation code will need to be in any way aware of which chain its providers are for, or have any chain-specific logic. All of that gets pushed down to the two implementations of the `ChainInclusion` trait.

nathanielc, line 114: Why do we use the AnchorProof struct when it has more information than is needed by the implementations? With this API its not clear if the `chain_inclusion_proof` method is checking if the root cid returned matches the root cid of the input proof structure. 

This API is about only getting the proof result correct? In that case I think I prefer having the associated type InclusionInput that makes it explicit what information a given provider needs in order to get the proof result. 

Thoughts? Maybe I have missed the motivation behind this change.

nathanielc: Maybe we do `type InclusionInput: From<AnchorProof>` so that callers can still use the anchor proof but then its clear that the provider itself is not validating the proof but instead just retrieving the information the proof references?

validation/src/blockchain/eth_rpc/types.rs
```diff
@@ -101,36 +93,25 @@ impl FromStr for EthProofType {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
-/// A proof of time on the blockchain
-pub struct TimeProof {
+/// A proof of time derived from state on the blockchain
+pub struct TimeProofResults {
```
nathanielc, line 97: We have lots of names for what seem like the same thing, can we consolidate on a consistent naming?

For example we have AnchorProof, TimeProofResults and chain_inclusion_proof. What is the difference between anchor, time, and chain inclusion? Do we really have three different types of proofs? Why is the function named `chain_inclusion_proof` when it returns a TimeProofResults?

Should the names instead be, AnchorProof, ChainInclusionProof and get_chain_inclusion_proof? Where anchor proof is a chain agnostic chain inclusion proof?  I'd be fine with other names too, its just not clear how all these thing relate given their current names and the differences imply there are more entities than there really are.

nathanielc (changes requested):
Overall looks good, however I have some questions below to help clear up some of my confusion.

validation/src/blockchain/eth_rpc/types.rs
stbrody: yep agreed.  Took your suggested names, I think those are clearer.

validation/src/blockchain/eth_rpc/types.rs
stbrody: Per discussion on discord, there's unfortunately no way to do this without leaking details of the different ChainInclusion implementations for different chains into the calling code.  Decided the lesser evil is taking the AnchorProof, even though its taking in more information than is needed, as it allows the EventValidation code to avoid any special logic that depends on what chain is being validated.

nathanielc (approved):

AaronGoldman (commented):
Some nits but looks good.

event-svc/src/event/validator/event.rs
```diff
@@ -226,7 +226,7 @@ impl EventValidator {
         Ok(validated_events)
     }
 
-    /// Transforms the [`ChainInclusionError`] into a [`ValidationError`] with an appropriate message
+    /// Transforms the [`eth_rpc::Error`] into a [`ValidationError`] with an appropriate message
```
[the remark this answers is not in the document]
stbrody: How so?  Function takes an `eth_rpc::Error` and returns a `ValidationError`

event-svc/src/event/validator/time.rs
```diff
@@ -94,12 +92,7 @@ impl TimeEventValidator {
             .get(&chain_id)
             .ok_or_else(|| eth_rpc::Error::NoChainProvider(chain_id.clone()))?;
 
-        let input = EthTxProofInput {
-            tx_hash: event.proof().tx_hash(),
-            tx_type: EthProofType::from_str(event.proof().tx_type())
-                .map_err(|e| eth_rpc::Error::InvalidProof(e.to_string()))?,
-        };
-        let proof = provider.chain_inclusion_proof(&input).await?;
+        let proof = provider.chain_inclusion_proof(event.proof()).await?;
```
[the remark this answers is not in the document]
stbrody: good call, renamed `proof` to `chain_proof`, added a comment, and added the other root to the error message

PR #571, opened 2024-10-28
feat: expose all tables via flight sql

nathanielc:
This change moves the OLAP aggregation logic within ceramic-one. This means that all tables are available via Flight SQL endpoint.

TODO: 

- [x] Manually test query command


stbrody:
Going to wait to review until CI is green

nathanielc:
@stbrody Ready for review now. Turns out there was a subtle issue with the federation used with the `query` command. So instead of trying to fix it I have changed the query command to be a simple client of the FlightSQL endpoint instead of full featured REPL

one/src/daemon.rs
```diff
@@ -236,6 +237,33 @@ pub struct DaemonOpts {
...
     )]
     ethereum_rpc_urls: Vec<String>,
+
+    /// Enable the aggregator, requires Flight SQL to be enabled
+    #[arg(
+        long,
+        requires = "flight_sql_bind_address",
+        env = "CERAMIC_ONE_AGGREGATOR"
+    )]
+    aggregator: Option<bool>,
```
stbrody, line 247: I'm trying to think if this should be a state enum instead of a boolean, but I'm having a hard time thinking about what other modes the aggregator might have in the future.  I guess since I can't think of anything than just a bool is probably fine

one/src/daemon.rs
```diff
@@ -236,6 +237,33 @@ pub struct DaemonOpts {
...
+    ///   * AWS_SECRET_ACCESS_KEY -> secret_access_key
+    ///   * AWS_DEFAULT_REGION -> region
+    ///   * AWS_ENDPOINT -> endpoint
+    ///   * AWS_SESSION_TOKEN -> token
+    ///   * AWS_ALLOW_HTTP -> set to "true" to permit HTTP connections without TLS
+    ///
+    #[arg(
+        long,
+        requires = "experimental_features",
+        env = "CERAMIC_ONE_S3_BUCKET"
```
stbrody, line 264: in the future the data files might live in a different bucket than the aggregator output right?  So we might want this name to indicate that this specifically for the result of the aggregated state that we are exporting.

one/src/daemon.rs
```diff
@@ -236,6 +237,33 @@ pub struct DaemonOpts {
...
+    /// Enable the aggregator, requires Flight SQL to be enabled
+    #[arg(
+        long,
+        requires = "flight_sql_bind_address",
+        env = "CERAMIC_ONE_AGGREGATOR"
+    )]
+    aggregator: Option<bool>,
+
+    /// S3 bucket name.
+    /// When configured the aggregator will support storing data in S3 compatible object stores.
```
stbrody, line 250: will *support*?  or will actually do so?

one/src/daemon.rs
```diff
@@ -469,7 +497,7 @@ pub async fn run(opts: DaemonOpts) -> Result<()> {
         .await
         .map_err(|e| {
             anyhow!(
-                "Failed to start libp2p server using addresses: {}. {}",
+                "Failed to start linebp2p server using addresses: {}. {}",
```
stbrody, line 500: typo?

one/src/daemon.rs
```diff
@@ -236,6 +237,33 @@ pub struct DaemonOpts {
         env = "CERAMIC_ONE_ETHEREUM_RPC_URLS"
     )]
     ethereum_rpc_urls: Vec<String>,
+
+    /// Enable the aggregator, requires Flight SQL to be enabled
+    #[arg(
+        long,
+        requires = "flight_sql_bind_address",
```
stbrody, line 244: also requires the s3 bucket?

one/src/daemon.rs
```diff
@@ -495,18 +523,50 @@ pub async fn run(opts: DaemonOpts) -> Result<()> {
...
+    let (flight_handle, aggregator_handle) = if let Some(addr) = opts.flight_sql_bind_address {
         let addr = addr.parse()?;
         let feed = event_svc.clone();
-        let mut shutdown_signal = shutdown_signal.resubscribe();
-        Some(tokio::spawn(async move {
-            ceramic_flight::server::run(feed, addr, async move {
-                let _ = shutdown_signal.recv().await;
+        let bucket = opts
+            .s3_bucket
+            .ok_or_else(|| anyhow!("s3_bucket option is required when exposing flight sql"))?;
```
stbrody, line 531: *should* this be required?  What if you just want the run the aggregation locally and query out the aggregated state feed?

stbrody (approved):
a few small comments but generally LGTM.

One question: where does this leave folks who want to build their own custom aggregators?  Can they still do so out-of-process by consuming the conclusion feed over the network?  Does this PR make this more difficult?

one/src/daemon.rs
nathanielc: I am anticipating a single bucket for all the data. A single bucket and handle many tables. I don't want to complicate the API by expecting a different bucket for each table until that need arises in product discussions.

nathanielc: I'll update the comment to reflect this intention

one/src/daemon.rs
nathanielc: Today its required as there is no mechanism for storing the state locally. We should likely add that for dev purposes but that is a separate issue.

one/src/daemon.rs
nathanielc: The plan is also to remove this boolean once we make it always on. We will likely need some form of config for which parts of the pipeline to enable. That will likely be a net new config option once we have move than one stage.

nathanielc:
> One question: where does this leave folks who want to build their own custom aggregators? Can they still do so out-of-process by consuming the conclusion feed over the network? Does this PR make this more difficult?

This does not make it more difficult. Consumers can still build their own aggregators out of band. In fact this makes is easier as all the data for any stage in the pipeline can be found in a single location.

one/src/daemon.rs
stbrody: should the name of the flag be `enable_aggregator` maybe?

stbrody:
> This does not make it more difficult. Consumers can still build their own aggregators out of band. In fact this makes is easier as all the data for any stage in the pipeline can be found in a single location.

This does remove all the tooling for building an external validator though right?  Now someone would need to not just make their specific aggregator, but also a build process for a standalone aggregator binary

nathanielc:
> This does remove all the tooling for building an external validator though right? Now someone would need to not just make their specific aggregator, but also a build process for a standalone aggregator binary 

That was true before however they could have copied the olap binary. But it was quite small and didn't have anything interesting to preserve. It was a blind copy of the ceramic-one binary setup.

PR #573, opened 2024-11-05
fix: fix cacao model auth check

dav1do:
I included a $ in the string interpolation like I was writing javascript (that's what I get for copying from js-ceramic). Now that's removed so it actually compares the model stream ID instead of the literal `${streamID}`.


stbrody (approved):
LGTM

PR #576, opened 2024-11-06
feat: use cache table to make queries faster and writes to object store batched

nathanielc:
Prior to this change each small batch of writes into the API would result in a single parquet file being written to object store. This behavior quickly creates many files that are too small to gain any column storage benefits. 

This change instead caches writes locally into an in memory table that once large enough is written as a single file into object store. This has a few consequences:

* Files written to object store are large enough to be valuable. (defaults to 10,000 rows)
* There is always a set of data that lives only in memory.
    * If the process is stopped or crashes this data must be recreated on next startup
    * This data can only be queried if read directly from the node that holds the data

### DIFF
--- a/pipeline/src/aggregator/mod.rs
+++ b/pipeline/src/aggregator/mod.rs
@@ -5,7 +5,10 @@
 mod ceramic_patch;
 
 use anyhow::Context;
-use arrow::{array::UInt64Array, datatypes::DataType};
+use arrow::{
+    array::{RecordBatch, UInt64Array},
+    datatypes::DataType,
+};
 use ceramic_patch::CeramicPatch;
 use datafusion::{
     common::JoinType,
@@ -22,12 +25,16 @@ use datafusion::{
     sql::TableReference,
 };
 use std::{future::Future, sync::Arc};
-use tracing::{debug, error};
+use tracing::{debug, error, instrument, Level};
 
-use crate::Result;
+use crate::{schemas, Result, DOC_STATE_MEM_TABLE, DOC_STATE_PERSISTENT_TABLE};
+
+// Maximum number of rows to fetch per pass of the aggregator.
+// Minimum number of rows to have proccessed before writing a batch to object store.
+const BATCH_SIZE: usize = 10_000;
 
 pub async fn run(ctx: SessionContext, shutdown_signal: impl Future<Output = ()>) -> Result<()> {
-    run_continuous_stream(ctx, shutdown_signal, 10000).await?;
+    run_continuous_stream(ctx, shutdown_signal, BATCH_SIZE).await?;
     Ok(())
 }
 
@@ -94,6 +101,7 @@ impl ContinuousStreamProcessor {
         })
     }
 
+    #[instrument(skip(self), ret ( level = Level::DEBUG ))]
     async fn process_batch(&mut self, limit: usize) -> Result<()> {
         // Fetch the conclusion feed DataFrame
         let mut conclusion_feed = self
@@ -131,7 +139,7 @@ impl ContinuousStreamProcessor {
             )?
             .build()?,
         );
-        process_feed_batch(self.ctx.clone(), df.clone()).await?;
+        process_feed_batch(self.ctx.clone(), df.clone(), limit).await?;
 
         // Fetch the highest index from the cached DataFrame
         let highest_index = df
@@ -162,7 +170,12 @@ impl ContinuousStreamProcessor {
 //  current conclusion_feed batch,
 //  * be valid JSON patch data documents.
 //  * use a qualified table name of `conclusion_feed`.
-async fn process_feed_batch(ctx: SessionContext, conclusion_feed: DataFrame) -> Result<()> {
+#[instrument(skip(ctx, conclusion_feed), ret ( level = Level::DEBUG ))]
+async fn process_feed_batch(
+    ctx: SessionContext,
+    conclusion_feed: DataFrame,
+    max_cached_rows: usize,
+) -> Result<()> {
     let doc_state = ctx
         .table("doc_state")
         .await?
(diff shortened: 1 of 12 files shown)


pipeline/src/cache_table.rs
```diff
@@ -0,0 +1,658 @@
+//! [`CacheTable`] for querying `Vec<RecordBatch>` by DataFusion.
```
nathanielc, line 1: FYI, this is a copy of the MemTable implementation that allows for clearing all the data in the table. We could upstream that change however I think we will likely end up optimizing this implementation over time and so having our own copy as a starting point is useful.

pipeline/src/aggregator/mod.rs
```diff
@@ -356,7 +353,7 @@ mod tests {
         conclusion_feed: RecordBatch,
     ) -> anyhow::Result<impl std::fmt::Display> {
         let ctx = init_ctx_cont(conclusion_feed).await?;
-        let shutdown_signal = tokio::time::sleep(tokio::time::Duration::from_millis(100));
+        let shutdown_signal = tokio::time::sleep(tokio::time::Duration::from_millis(500));
```
dav1do, line 356: this is probably okay, but it might be nice to use a oneshot channel of () and just send that in at the end. if this seems reliable, i'm fine with leaving it as is.

pipeline/src/lib.rs
```diff
@@ -41,10 +48,26 @@ pub async fn session_from_config<F: ConclusionFeed + 'static>(
...
+    match config.conclusion_feed {
+        ConclusionFeedSource::Direct(conclusion_feed) => {
+            ctx.register_table(
+                "conclusion_feed",
+                Arc::new(conclusion::FeedTable::new(conclusion_feed)),
+            )?;
+        }
+        #[cfg(test)]
+        ConclusionFeedSource::InMemory(table) => {
+            debug_assert_eq!(
```
dav1do, line 60: should this be a regular assert in case we run tests in release mode? probably fine as is just a thought.

pipeline/src/lib.rs
```diff
@@ -41,10 +48,26 @@ pub async fn session_from_config<F: ConclusionFeed + 'static>(
...
+                schemas::conclusion_feed(),
+                datafusion::catalog::TableProvider::schema(&table)
+            );
+            ctx.register_table("conclusion_feed", Arc::new(table))?;
+        }
+    };
+    // Register the _internal schema
+    ctx.catalog("ceramic")
+        .unwrap()
```
dav1do, line 69: maybe `expect` since it seems like the ceramic catalog must exist? my preference is generally unwrap in tests and expect for "cannot fail/asserts" but again, fine to just close as is since we add it to the session above.

pipeline/src/lib.rs
```diff
@@ -69,13 +92,31 @@ pub async fn session_from_config<F: ConclusionFeed + 'static>(
     // Set the path within the bucket for the doc_state table
     const DOC_STATE_OBJECT_STORE_PATH: &str = "/ceramic/v0/doc_state/";
     url.set_path(DOC_STATE_OBJECT_STORE_PATH);
-    ctx.register_listing_table(
+    // Register doc_state_persistent as a listing table
+    ctx.register_table(
+        "_internal.doc_state_persistent",
```
dav1do, line 97: it seems like we have a lot of strings for table names. would it make sense to add a few `const`s so they're defined in one place?

pipeline/src/schemas.rs
```diff
@@ -3,8 +3,61 @@ use std::sync::{Arc, OnceLock};
 
 use datafusion::arrow::datatypes::{DataType, Field, Fields, SchemaBuilder, SchemaRef};
 
+static CONCLUCSION_FEED: OnceLock<SchemaRef> = OnceLock::new();
```
dav1do, line 6: typo in `CONCLUSION` here but it's private so nbd.

dav1do (commented):
A few minor things but overall looks like nice cleanup and a good improvement :shipit:

pipeline/src/aggregator/mod.rs
nathanielc: The continuous run loop doesn't end, so we need some kind of timeout to give it enough time to run. As I added an extra row to the test case I needed to give it more time. Alternatively poll the results until we get a passing result from the tables, I'll look into changing that.

pipeline/src/lib.rs
nathanielc: good call

nathanielc:
@dav1do Thanks for the review. I have updated the PR. Your suggestions really helped clean some things up.

pipeline/src/lib.rs
```diff
@@ -37,6 +37,11 @@ pub use conclusion::{
 };
 pub use config::{ConclusionFeedSource, Config};
 
+pub const CONCLUSION_FEED_TABLE: &str = "ceramic.v0.conclusion_feed";
+pub const DOC_STATE_TABLE: &str = "ceramic.v0.doc_state";
+pub const DOC_STATE_MEM_TABLE: &str = "ceramic._internal.doc_state_mem";
+pub const DOC_STATE_PRESISTENT_TABLE: &str = "ceramic._internal.doc_state_persistent";
```
dav1do, line 43: PRESISTENT -> PERSISTENT 😄

nathanielc: I can't type in all caps apparently, thanks I'll get this fixed.

pipeline/src/aggregator/mod.rs
```diff
@@ -222,113 +235,94 @@ async fn process_feed_batch(ctx: SessionContext, conclusion_feed: DataFrame) ->
...
-        // Write states to the doc_state table
-        .write_table("doc_state", DataFrameWriteOptions::new())
+        // Write states to the in memory doc_state table
+        .write_table(DOC_STATE_MEM_TABLE, DataFrameWriteOptions::new())
         .await
         .context("computing states")?;
+
+    let count = ctx.table(DOC_STATE_MEM_TABLE).await?.count().await?;
+    // If we have enough data cached in memory write it out to persistent store
+    if count >= max_cached_rows {
```
stbrody, line 245: I know I'm late to the party here, but is there an issue if no data gets written for a while that writes could sit in the memory table but never get flushed to disk?  Do we need a time-based flush condition in addition to a number of records based one?

nathanielc: Its possible that writes stay in memory. However I do not think we need a time based flush mechanism because the data is derived and can be recomputed easily and quickly. 

If the process exists then any data in memory is lost. However on startup that data is recomputed and repopulated into memory. This has the nice behavior that we do not end up with small files in low write use cases. And since the buffer size is the same as the batch size this is means that the amount of data that needs to be recomputed on startup is always manageable, meaning is never larger than a batch we would have computed during normal operation anyways.

stbrody: okay cool.  So I guess the risk is just that the data Parquet in S3 is out of date during periods of low writes.  Better than data loss, if the latency becomes an issue for someone we can add a time-based flush then I suppose.

nathanielc: Yep, and if they query the parquet data through the Flight SQL API on ceramc-one then the in memory data is used to serve the query

PR #578, opened 2024-11-07
chore: version v0.41.2

smrz2001:
### 🚀 Features

- Expose all tables via flight sql (#571)

### ⚙️ Miscellaneous Tasks

- Use larger github runner for image builds (#577)


nathanielc:
Should be a minor release not a patch release

PR #586, opened 2024-11-12
feat: add api to get stream state

### LINKED ISSUE #581: Ceramic-one HTTP exposes endpoint to get current state of a stream
In order to write data into the API a document state diff needs to be computed. This means we need easy access to the current state of a stream document. 

DOD
- [ ] HTTP API exists that can report the current state of a stream given a stream id
- [ ] Only the current stream state is returned, no information about history is returned.

nathanielc:
With this change it is now possible to get the current state of a stream via the HTTP JSON API.

Fixes #581


api/src/tests.rs
```diff
@@ -741,3 +837,136 @@ async fn test_events_event_id_get_by_cid_success() {
...
+            &Context,
+        )
+        .await;
+    expect![[r#"
+        Ok(
+            Success(
+                StreamState {
+                    id: "k2t6wzhjp5kk3zrbu5tyfjqdrhxyvwnzmxv8htviiganzacva34pfedi5g72tp",
+                    event_cid: "baeabeibisehf2y3f6wgyltbs27mg4qcw75vzhunpgzorzgdklzioliwtba",
+                    controller: "did:itwasntme",
```
dav1do, line 877: now I'm going to start humming shaggy songs 😂

one/src/daemon.rs
```diff
@@ -524,52 +524,59 @@ pub async fn run(opts: DaemonOpts) -> Result<()> {
...
                 .await
-                {
-                    error!(%err, "aggregator task failed");
-                }
-            }))
+            });
+            (pipeline_ctx, aggregator_handle, Some(flight_handle))
         } else {
-            None
+            // TODO
```
dav1do, line 572: what's this TODO represent? is there an issue or some more context you could add?

api/src/server.rs
```diff
@@ -356,6 +367,8 @@ pub struct Server<C, I, M> {
     insert_task: Arc<InsertTask>,
     marker: PhantomData<C>,
     authentication: bool,
+
+    pipeline: SessionContext,
```
dav1do, line 371: is it worth making this an option and just returning an error/bad request saying it's disabled if the context isn't set because the feature is off?

dav1do (approved):
Nice, looks good. A couple small comments and looks like there's a format issue to resolve but nothing major from me.

api/ceramic.yaml
```diff
@@ -632,3 +679,33 @@ components:
         name:
           type: string
           description: Name of the Ceramic network
+    StreamState:
+      title: State of a Ceramic stream
+      description: The state of a Ceramic stream as defined by the stream type aggregation and conflict resolution rules.
+      type: object
+      required:
+        - id
+        - event_cid
```
stbrody, line 688: historically we've called this `tip`, do we want to maintain that convention?

stbrody, line 688: also, should this be an array?  With multiprev we now support streams with multiple tips

api/ceramic.yaml
```diff
@@ -632,3 +679,33 @@ components:
...
+        - controller
+        - dimensions
+        - metadata
```
stbrody, line 691: what's the difference between dimensions and metadata?

api/ceramic.yaml
```diff
@@ -632,3 +679,33 @@ components:
...
+          description: Controller of the stream
+        dimensions:
+          type: object
+          description: Dimensions of the stream
+        metadata:
+          type: object
+          description: Metadata of the stream
+        data:
+          type: object
+          description: State of the stream
```
stbrody, line 711: "State" feels overloaded here.

"content" of the stream?

api/src/server.rs
```diff
@@ -817,6 +832,142 @@ where
...
+                col("state"),
+                col("index"),
+            ])
+            .map_err(|err| ErrorResponse::new(format!("failed to select: {err}")))?
+            .filter(col("stream_cid").eq(lit(stream_id.cid.to_bytes())))
+            .map_err(|err| ErrorResponse::new(format!("failed to filter: {err}")))?
+            .aggregate(
+                vec![col("stream_cid"), col("controller")],
+                vec![
+                    last_value(vec![col("state")])
```
stbrody, line 861: what is the "state" column?  Shouldn't this be "data" or "content"?

api/src/server.rs
```diff
@@ -817,6 +832,142 @@ where
...
+                        .order_by(vec![col("index").sort(true, true)])
+                        .build()
+                        .map_err(|err| {
+                            ErrorResponse::new(format!(
+                                "failed to define last_value state query: {err}"
+                            ))
+                        })?
+                        .alias("state"),
+                    last_value(vec![col("event_cid")])
+                        .order_by(vec![col("index").sort(true, true)])
```
stbrody, line 871: kinda sucks that you have to do this sort statement multiple times for each column.  I'm surprised you can't just do a sort of al the rows once, then select the columns from the selected row?

nathanielc: same, I was annoyed as well. Should not affect performance but its quite verbose.

api/src/server.rs
nathanielc: State is the name of the column that contains both the metadata and the data. We do no use the word content anywhere.

api/ceramic.yaml
nathanielc: dimensions are static for the stream, metadata is mutable over the lifetime of the stream. shouldIndex is the only metadata field currently

api/ceramic.yaml
nathanielc: I don't think so, this API wants to assume there is always a canonical singular tip to a stream. What would the ceramic-sdk do if it gets back multiple tips?

nathanielc: No sure, I like the name event_cid because that makes it clear you can use the cid to lookup an event directly. With tip it's not clear what the CID points to

api/ceramic.yaml
nathanielc: Its not the content, its both the metadata and the data (previously called content). Maybe we should call this state, metadata, content?

The state is currently defined as:
```
struct State{
   metadata: Object,
   data: Object,
}
```

api/ceramic.yaml
stbrody: how do we expect to handle a CRDT type stream that can diverge and remerge?  What do you do when you have two different updates at the same log height?  You need a way to return both tips and then let the application construct a new event that merges the states and includes both tips in the prev array

api/ceramic.yaml
stbrody: hmm, the problem is this whole structure is called `StreamState`.  Having a field called `state` inside a struct called `StreamState` feels confusing to me.  Now when we hear the word "state" we don't know if we mean the inner field or the outer structure.

> Its not the content, its both the metadata and the data
I think this isn't true, since there is already a different metadata field at the top level here.  So having `metadata` and `data` at the top level avoids the need to expose anything called "state" to the user at all.

I don't love that internally we call the pair of metadata and data "state", but at least that's hidden as an internal implementation detail.  If we can keep that concept out of the public API that might be good enough for now.

api/src/server.rs
stbrody: I wonder if we should rename what we call "state" to "data", and what we call "data" to "content"?

api/ceramic.yaml
nathanielc: Long term yes and its possible to get multiple tips via the Flight SQL API. This API is for the narrow use case of the SDK needing to do a read in order to do a write. When we add CRDT type streams the SDK will not need to do a read before a write. It can just do the write. I imagine there are other similar differences with yet unknown stream types. As such I want this API to remain specific to the immediate needs.

api/ceramic.yaml
nathanielc: I see the confusion, let me see if I can make the table structure mirror the flattened structure of this StreamState object more closely.

nathanielc: The trouble here is keeping this API generic to stream types. Metadata is a property of the MID stream type. That is causing the problems here. I say we leave metadata and data as opaque to this API

How does this look as a definition of the StreamState:

```
    StreamState:
      title: State of a Ceramic stream
      description: The state of a Ceramic stream as defined by the stream type aggregation and conflict resolution rules.
      type: object
      required:
        - id
        - event_cid
        - controller
        - dimensions
        - data
      properties:
        id:
          type: string
          description: Multibase encoding of the stream id
        event_cid:
          type: string
          description: CID of the event that produced this state
        controller:
          type: string
          description: Controller of the stream
        dimensions:
          type: object
          description: Dimensions of the stream, each value is multibase encoded.
        data:
          type: object
          description: Data of the stream
```


Then `data` for MID streams will be an object of 

```
{
   metadata: ...,
   content: ...,
}
```

Thoughts?

This way we do not overload names and uses of the existing clients will understand that `content` is the same as the original meaning of content.

api/src/server.rs
nathanielc: This is effectively what I propose above. Thanks for the suggestion.

api/ceramic.yaml
stbrody: yeah that sounds good!

nathanielc: @stbrody Ok adjusted to match the now consistent names and structure defined in the pipeline/DESIGN.md doc.

stbrody (approved):

PR #587, opened 2024-11-12
refactor: consistent pipeline table names and schemas

nathanielc:


dav1do (approved):
Looks good, a couple of small questions.

pipeline/DESIGN.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Features
+
+* Access to the identifying information for streams
+
+### Schema
+
+| Column      | Type              | Description              |
+| ------      | ----              | -----------              |
+| stream_cid  | bytes             | Cid of the stream        |
```
dav1do, line 60: Do all the tables in this doc store stream CID or is any stream ID? Or do we just combine stream type + cid on the fly for that?

pipeline/DESIGN.md
```diff
@@ -0,0 +1,190 @@
...
+| stream_type | u8                | Cid of the stream        |
```
dav1do, line 61: description (Cid of the stream) is wrong and it might be nice to provide a couple of examples e.g. 1 = init, 2 = data, 3 = time

pipeline/DESIGN.md
```diff
@@ -0,0 +1,190 @@
...
+| state       | string            | JSON encoding of the event state                                                                |
+
+### Transformation
+
+This table computes the aggregated state for each conclusion event.
+Additionally it validates the aggregated state matches the model schema of the stream.
+
+## stream_heads
+
+The stream_states table contains a row for head of each stream representing the canonical state of the stream heads.
```
dav1do, line 143: are we no longer using tips as the language here? heads is more consistent with git but js-ceramic/old docs seem to talk more about stream tips.

pipeline/DESIGN.md
```diff
@@ -0,0 +1,190 @@
...
+| stream_cid  | bytes             | Cid of the stream                                                                               |
+| stream_type | u8                | Stream type                                                                                     |
+| controller  | string            | Controller of the stream                                                                        |
+| dimensions  | map(string,bytes) | Set of key values dimension pairs of the stream                                                 |
+| event_cid   | bytes             | Cid of the event                                                                                |
+| event_type  | u8                | One of Data (0) or Time (1)                                                                     |
+| data        | bytes             | DAG-JSON encoding of the event payload                                                          |
+| previous    | list(bytes)       | List of CID previous to this event                                                              |
```
dav1do, line 93: is there an order to this list?

pipeline/DESIGN.md
```diff
@@ -0,0 +1,190 @@
...
+* Access to the multiple heads of streams for users building their on conflict resolution
+
+### Schema
+
+| Column      | Type              | Description                                                                                     |
+| ------      | ----              | -----------                                                                                     |
+| index       | u64               | Order of this event. Index is always greater than the index of any previous event in the stream |
+| stream_cid  | bytes             | Cid of the stream                                                                               |
+| stream_type | u8                | Stream type                                                                                     |
+| controller  | string            | Controller of the stream                                                                        |
```
dav1do, line 156: maybe we can add something about controller typically being a did:key or did:pkh somewhere in this doc?

pipeline/DESIGN.md
```diff
@@ -0,0 +1,190 @@
...
+| dimensions  | map(string,bytes) | Set of key values dimension pairs of the stream                                                 |
```
dav1do, line 157: would there be any value in adding more about what dimensions mean or linking to the forum post?

pipeline/DESIGN.md
```diff
@@ -0,0 +1,190 @@
...
+    stream_heads --> stream_states;
+```
+
+## raw_events
+
+The raw_events table contains a row for each event in a stream and contains the raw CAR data of the event.
+
+### Features
+
+* Access to the raw event data that can be used to validate signatures
```
dav1do, line 38: I think we landed on not storing any events with invalid signatures in the current code. Is this focused on the future where we store them (maybe with a ttl or something) and move them to the next phase of the pipeline?

pipeline/DESIGN.md
```diff
@@ -0,0 +1,190 @@
...
+## model_schemas
+
+The model_schemas table contains a row for each model and contains the complete resolved schema of the model.
+
+TBD how this table is populated and its schema.
+This table may be able to be combined with the streams table. Should we?
+
+## event_states
+
+The event_states table contains a row for each event in a stream and the state of the document at that point in the stream.
```
dav1do, line 116: I think adding more context to "state" here might be useful but I'm not totally sure what I want/expect. Maybe something about how the state determined by applying the  stream operation (currently JSON patch) to arrive at the materialized state.. I don't know, maybe not helpful. 

I'm trying to decide if this should be accessible to newer ceramic consumers and provide a lot of context or be more of a guide for us. Some of my other comments are also along that line and might not apply.

pipeline/DESIGN.md
```diff
@@ -0,0 +1,190 @@
...
+| index       | u64               | Order of this event. Index is always greater than the index of any previous event in the stream |
+| stream_cid  | bytes             | Cid of the stream                                                                               |
+| stream_type | u8                | Stream type                                                                                     |
+| controller  | string            | Controller of the stream                                                                        |
+| dimensions  | map(string,bytes) | Set of key values dimension pairs of the stream                                                 |
+| event_cid   | bytes             | Cid of the event                                                                                |
+| event_type  | u8                | One of Data (0) or Time (1)                                                                     |
+| state       | string            | JSON encoding of the event state                                                                |
```
dav1do, line 134: Is this always going to be JSON? 

I guess I'm wondering if this type should be bytes that are interpreted based on stream type? Either json or whatever we want it to be? We could always multibase encode arrow or avro or whatever into a string too so bytes vs string might not matter (and it's possible I'm just missing something xd)

pipeline/src/lib.rs
```diff
@@ -37,10 +37,10 @@ pub use conclusion::{
 };
 pub use config::{ConclusionFeedSource, Config};
 
-pub const CONCLUSION_FEED_TABLE: &str = "ceramic.v0.conclusion_feed";
-pub const DOC_STATE_TABLE: &str = "ceramic.v0.doc_state";
-pub const DOC_STATE_MEM_TABLE: &str = "ceramic._internal.doc_state_mem";
-pub const DOC_STATE_PERSISTENT_TABLE: &str = "ceramic._internal.doc_state_persistent";
+pub const CONCLUSION_EVENTS_TABLE: &str = "ceramic.v0.conclusion_events";
```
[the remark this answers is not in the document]
nathanielc: Correct, its still behind experimental flags

pipeline/src/lib.rs
```diff
@@ -56,17 +56,17 @@ pub async fn session_from_config<F: ConclusionFeed + 'static>(
     match config.conclusion_feed {
```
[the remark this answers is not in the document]
nathanielc: This was intentional as the feed now exclusively refers to the trait that can get a feed of events from sqlite.

pipeline/DESIGN.md
nathanielc: Its always stream cid and you combine with type on the fly to get the stream id. The CID is the unique identifier for the stream. The type is just extra metadata.

pipeline/DESIGN.md
nathanielc: Ok yeah I like tips, will change.

pipeline/DESIGN.md
nathanielc: Yes, I'll explain in description.

pipeline/DESIGN.md
nathanielc: Yeah I think we keep with that strategy of not storing invalid events. However this table gives consumers direct access to the raw data so that if they do not trust that ceramic-one validated signatures correctly they can do it themselves. Without this table its impossible for anyone to validate signatures themselves.

pipeline/DESIGN.md
nathanielc: I have gone back on forth on this, making it a string makes it much easier to work with as all the tooling will just display it. However to make it more generic it should be bytes. Let me see if the built in json functions work with bytes columns and then if so I'll update this to be bytes.

pipeline/DESIGN.md
nathanielc: The audience of this doc are developers of ceramic-one. For now I'll leave the details of these other bits light. However this does bring up some good content for a similar doc we will want on the docs sites.

pipeline/DESIGN.md
nathanielc: When with bytes. It simple to cast to a string when needed i.e. `select state::varchar->>'data' from event_states`

pipeline/DESIGN.md
```diff
@@ -0,0 +1,212 @@
...
+| Tile                    | 0x00 | A stream type representing a json document                 | https://cips.ceramic.network/CIPs/cip-8                                                             |
+| CAIP-10 Link            | 0x01 | Link blockchain accounts to DIDs                           | https://cips.ceramic.network/CIPs/cip-7                                                             |
+| Model                   | 0x02 | Defines a schema shared by group of documents in ComposeDB | https://github.com/ceramicnetwork/js-ceramic/tree/main/packages/stream-model                        |
+| Model Instance Document | 0x03 | Represents a json document in ComposeDB                    | https://github.com/ceramicnetwork/js-ceramic/tree/main/packages/stream-model-instance               |
+| UNLOADABLE              | 0x04 | A stream that is not meant to be loaded                    | https://github.com/ceramicnetwork/js-ceramic/blob/main/packages/stream-model/src/model.ts#L163-L165 |
+| EventId                 | 0x05 | An event id encoded as a cip-124 EventID                   | https://cips.ceramic.network/CIPs/cip-124                                                           |
```
stbrody, line 203: I forgot we did this.  Do you remember why we need this?

stbrody: > encoded as a cip-124 EventID

Except in this context isn't it encoded as a StreamID?

stbrody: We should also probably clarify in the description that this isn't a "real" StreamType.  Similar to UNLOADABLE

pipeline/DESIGN.md
```diff
@@ -0,0 +1,212 @@
...
+
+Source https://cips.ceramic.network/CIPs/cip-59#streamid-multicodec
+
+### Event Types
+
+| name | code | description                                      |
+| ---- | ---- | -----------                                      |
+| Data | 0x00 | An event containing data for the stream          |
```
stbrody, line 211: what about init?

pipeline/DESIGN.md
```diff
@@ -0,0 +1,212 @@
...
+| event_type  | u8                | Type of the event, see [event type values](#event-types)                                        |
+| data        | bytes             | DAG-JSON encoding of the event payload                                                          |
+| previous    | list(bytes)       | Ordered list of CID previous to this event. Meaning of the order is stream type dependent       |
+
+
+### Transformation
+
+Raw events are transformed into a flattened structure where stream dimensions and conclusions about data are added to the data.
+
+Conclusions include:
```
stbrody, line 100: also timestamp

pipeline/DESIGN.md
```diff
@@ -0,0 +1,212 @@
...
+
+* The event has a valid signature
+* The dimensions and controller of the event
+
+This table joins the raw_events, chain_proofs, and streams tables in order to make the conclusions about the raw events.
+
+## model_schemas
+
+The model_schemas table contains a row for each model and contains the complete resolved schema of the model.
```
stbrody, line 109: each model... known about by the node?  Presumably we don't plan to keep indexing all Models forever

pipeline/DESIGN.md
```diff
@@ -0,0 +1,212 @@
...
+| index       | u64               | Order of this event. Index is always greater than the index of any previous event in the stream |
+| stream_cid  | bytes             | Cid of the stream                                                                               |
+| stream_type | u8                | Type of the stream, see [stream type values](#stream-types)                                     |
+| controller  | string            | Controller of the stream                                                                        |
+| dimensions  | map(string,bytes) | Set of key values dimension pairs of the stream                                                 |
+| event_cid   | bytes             | Cid of the event                                                                                |
+| event_type  | u8                | Type of the event, see [event type values](#event-types)                                        |
+| state       | bytes             | JSON encoding of the event state                                                                |
```
stbrody, line 134: rename to "data"?

pipeline/DESIGN.md
```diff
@@ -0,0 +1,212 @@
...
+This table computes the aggregated state for each conclusion event.
+Additionally it validates the aggregated state matches the model schema of the stream.
+
+## stream_tips
+
+The stream_states table contains a row for head of each stream representing the canonical state of the stream tips.
+
+### Features
+
+* Access to the multiple tips of streams for users building their on conflict resolution
```
stbrody, line 147:
```suggestion
* Access to the multiple tips of streams for users building their own conflict resolution
```

pipeline/DESIGN.md
```diff
@@ -0,0 +1,212 @@
...
+
+### Schema
+
```
stbrody, line 150: should we add event height to the schema here?  Seems like it will be needed in the next pipeline phase to do conflict resolution

pipeline/DESIGN.md
stbrody, line 147: we probably want to clarify what "tip" means here.  Something like it's the most recent event in a single history of the stream

pipeline/DESIGN.md
```diff
@@ -0,0 +1,212 @@
...
+| stream_type | u8                | Type of the stream, see [stream type values](#stream-types)                                     |
+| controller  | string            | Controller of the stream                                                                        |
+| dimensions  | map(string,bytes) | Set of key values dimension pairs of the stream                                                 |
+| event_cid   | bytes             | Cid of the event                                                                                |
+| event_type  | u8                | Type of the event, see [event type values](#event-types)                                        |
+| state       | bytes             | JSON encoding of the event state                                                                |
+
+### Transformation
+
+Computes the singular head that is the canonical state of the stream.
```
stbrody, line 189: maybe add a TODO to link to something writing out the conflict resolution rules?

pipeline/DESIGN.md
```diff
@@ -0,0 +1,212 @@
...
+
+Data enters Ceramic via its API and is stored into the raw_events table.
+From there various transformations are applied producing various intermediate tables.
+Each table schema is considered public API and provides access to kind of query pattern against the data.
+
+```mermaid
+graph LR;
+    raw_events;
+    streams;
+    chain_proofs;
```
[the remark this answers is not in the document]
nathanielc: The intent of this table is to store the [ChainInclusionProof](https://github.com/ceramicnetwork/rust-ceramic/blob/main/validation/src/blockchain/eth_rpc/types.rs#L97) structs. This way the conclusion_events table can be built by joining the raw_events table with this table to make time based conclusions. Should we call it chain_inclusion_proofs? It is not the proof contained within the raw events but rather the timestamp and root cid that we know are included in a chain.

pipeline/DESIGN.md
```diff
@@ -0,0 +1,212 @@
...
+
+* Access to events allowing in order access to event within stream
+* Access to pre-aggregated data for users building their own aggregation system
+* Validation of the event signatures
+
+### Schema
+
+| Column      | Type              | Description                                                                                     |
+| ------      | ----              | -----------                                                                                     |
+| index       | u64               | Order of this event. Index is always greater than the index of any previous event in the stream |
```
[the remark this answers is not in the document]
nathanielc: Its the global node-specific in order delivery index. However that is an implementation detail. The only property of the index that matters is that all previous events in a stream have a lower index than the current event. This way if you need to query a stream for all events previous to an event its possible to do so without traversing the event graph directly.

pipeline/DESIGN.md
nathanielc: Not sure, just copied from the spec. I'll add a note.

pipeline/DESIGN.md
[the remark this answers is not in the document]
stbrody: yeah that struct is kind of named weird cause it's not actually storing the proof.  It's storing the derived conclusions that come from validating the proof.

How about... `time_conclusions`?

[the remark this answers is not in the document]
nathanielc: Works for me.

pipeline/DESIGN.md
nathanielc: Yes, its only for conclusion events and its derivatives. Raw events do not have an event type enum. Its not clear yet how much we should explode the raw event table to have fields about the event itself. For now it has only the cids and the car file.

PR #597, opened 2024-11-15
feat: backoff recon synchronization to peers with bad data

nathanielc:
With this change peers will backoff synchronization attempts with peers that fail synchronization. This will have the effect that the bootstrap nodes do not talk to nodes that are sending bad data very often.

However the backoff state is tied to the connection state and is reset if the connection resets. Allowing developers to restart their node and be in good standing with bootstrap nodes.

Additionally a small change was added to prefer using the connected address to a peer when reporting the address of connected peers.

### DIFF
--- a/recon/src/libp2p/tests.rs
+++ b/recon/src/libp2p/tests.rs
@@ -134,12 +134,17 @@ macro_rules! setup_test {
             Metrics::register(&mut Registry::default()),
         );
 
+        // Use shorter timings for testing.
+        let config = crate::libp2p::Config {
+            per_peer_sync_delay: std::time::Duration::from_millis(10),
+            per_peer_sync_backoff: 100.0,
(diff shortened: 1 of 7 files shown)


smrz2001 (approved):
🥳

recon/src/libp2p/tests.rs
```diff
@@ -243,7 +250,120 @@ async fn responder_model_error() {
...
+            // Third model sync is skipped because the backoff pushed it past the interests sync
+            Some(StreamSet::Interest),
+            Some(StreamSet::Interest),
+        ];
+        assert_eq!(
+            stream_sets,
+            vec![expected_stream_set_order.clone(), expected_stream_set_order]
+        );
+        assert_eq!(
+            p2_e4,
```
smrz2001, line 343: `p2_e8` would also be a failure, right? Don't have to check for it, just confirming my understanding of how the test works.

PR #612, opened 2024-11-20
chore: automatically select release level

nathanielc:
Doing all the releases has annoyed me that I have to check the release log before running the Create PR action to determine whether it should be a minor or patch release. Now its automatic. This logic comes from the rust-dag-jose repo originally.


smrz2001 (approved):
The strategy looks good to me. With scripts, the devil is always in the details LOL, so as long as it runs and works, it's good to go.

🚀

PR #613, opened 2024-11-20
fix: handle double write of same data

nathanielc:
Prior to this change a write of the same data (keyed by the event id) would cause an error the batching logic of the write path. It was assumed that double writes would not happen and so the channel to report the insert result was dropped, causing an error.

Now all channels are preserved and the result is sent to all.


smrz2001 (approved):
LGTM!

🚢

Issue #603, opened 2024-11-20
Create Signed Peer Structure


nathanielc:
## Description

We need a Rust structure that can be used to communicate the existence of a peer. This structure must (de)serialize as it will be used as the value of the Recon peers ring. 

## DOD

- [ ] The Rust structure exists that implements serde traits
- [ ] The structure contains the following information:
    - [ ] did:key of local peer
    - [ ] expiration time
    - [ ] list of multiaddrs on which the peer can be contacted
- [ ] This structure should have methods to sign it and produce a compact string encoded [JWS](https://www.rfc-editor.org/rfc/rfc7515#section-3.3) (ala data.data.data)
- [ ] This structure should have methods to be constructed from a string encoded JWS that verifies its signature.
- [ ]  Unit tests for the construction serialization and signing/verification of the struct

PR #615, opened 2024-11-21
feat: add peer entry and peer key structs

nathanielc:
These structures allow for communicating information about Peers and their addresses. The structs can be signed as JWS and verified.

Fixes: #603


core/src/peer.rs
```diff
@@ -0,0 +1,208 @@
...
+    /// Construct an entry about a peer with address that is no longer valid after expiration seconds after the
+    /// UNIX epoch.
+    pub fn new(local_id: NodeId, expiration: u32, addresses: Vec<Multiaddr>) -> Self {
+        let peer_id = local_id.peer_id();
+        Self {
+            id: local_id,
+            expiration,
+            addresses: addresses
+                .into_iter()
+                .map(|addr| ensure_multiaddr_has_p2p(addr, peer_id))
```
dav1do, line 27: should this be an error if the peer has no addresses?

core/src/peer.rs
```diff
@@ -0,0 +1,208 @@
...
+            "{:0>11}.{}",
+            entry.expiration,
+            entry.to_jws(node_key)?
+        )))
+    }
+    /// Decode and verify key as a [`PeerEntry`].
+    pub fn to_entry(&self) -> anyhow::Result<PeerEntry> {
+        let idx = self
+            .0
+            .find('.')
```
dav1do, line 107: could maybe use `split_once` here instead?

core/src/peer.rs
nathanielc: I don't think so. Its not useful to advertise your existence with no addresses, but I don't think its an error.

dav1do (approved):

PR #619, opened 2024-11-27
chore: log current highwater mark when processing events

dav1do:
This should tell us how far we've made it in addition to how many events we've had to process so far.


smrz2001 (approved):
:shipit:

PR #620, opened 2024-11-27
feat: use many tasks to order streams and discover undelivered events at startup

dav1do:
After an IPFS migration, we have to review all the data in the database to make sure we have the complete stream history before we can send events out of the API. We originally kept it simple and would read events and process them, then repeat in a single task. This was taking far too long on large datasets (e.g. 100s of GBs). Now we spawn multiple tasks to read the events from the database, and they send events over a channel to the ordering task (like we do during normal operation). This task was also modified to spawn multiple tasks to process events by stream and order them. Both changes appeared necessary during testing as one side would waiting on the other. This allows us to keep a solid rate of processing going and we've seen a substantial improvement in runtime (~60-100x faster).

On the discovery side: At startup, we spawn 16 tasks to read batches from the database. The number of events read each time was reduced to 250, as 1000 was taking seconds. The values are slightly arbitrary but this seemed like a "fast enough" choice during testing (the goal is simply to keep the channel full). The event data is partitioned using `(rowid % number_tasks) = task_number` so we don't have to do anything clever to split the data into batches up front. Each task starts from the beginning and pick up any events that have been missed. Once it finishes, the subsequent runs are fast, so we spawn the tasks regardless of whether they're needed.

On the ordering side a few changes were made. First, the channel size was reduced to 10000 (the previous value was far too large) and we try to empty it before doing any ordering since we have more tasks to process the set, and any events found may avoid database reads if they're for the same stream. Once events are grouped by stream, we split the streams into batches and spawn 1-16 tasks to process each batch. This processing has cpu bound work, but also requires database reads so multiple tasks have been beneficial. The tasks then send their ordered data back to the manager, which handles writing to the database. During testing, I made a change to remove a RO connection from the pool (and allow it to grow afterward) for each of these tasks. It didn't seem to make an obvious difference, but it may be useful to revisit.

### DIFF
--- a/event-svc/src/event/ordering_task.rs
+++ b/event-svc/src/event/ordering_task.rs
@@ -5,7 +5,9 @@ use anyhow::anyhow;
 use ceramic_event::unvalidated;
 use cid::Cid;
 use ipld_core::ipld::Ipld;
+use itertools::Itertools;
 use tokio::sync::mpsc::Sender;
+use tokio::task::JoinSet;
 use tracing::{debug, error, info, trace, warn};
 
 use crate::event::service::PENDING_EVENTS_CHANNEL_DEPTH;
@@ -19,6 +21,12 @@ type EventCid = Cid;
 type PrevCid = Cid;
 
 const LOG_EVERY_N_ENTRIES: usize = 10_000;
+/// The max number of tasks we'll spawn when ordering events.
+/// Each stask is given a chunk of streams and wil order them
+const MAX_STREAM_PROCESSING_TASKS: usize = 16;
+/// The minimim number of streams each task will process. If there are many streams, we'll spawn
+/// STREAM_NUM / MAX_STREAM_PROCESSING_TASKS.
+const MIN_NUM_STREAMS_PER_BATCH: usize = 25;
 /// The number of events we initially pull from the channel when doing startup undelivered batch processing.
 /// Being larger was measured to go faster when processing millions of events, however there's no need to
 /// allocate such a large array when normally processing events in the background as we tend to keep up.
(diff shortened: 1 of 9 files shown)


event-svc/src/event/ordering_task.rs
```diff
@@ -460,106 +539,244 @@ impl OrderingState {
...
+    /// Splits `to_process` into chunks and spawns at most `MAX_STREAM_PROCESSING_TASKS` to review and order
+    /// the events for each chunk of streams.
+    fn spawn_tasks_for_stream_batches(
+        event_access: Arc<EventAccess>,
+        to_process: HashMap<Cid, StreamEvents>,
+    ) -> JoinSet<Result<Vec<(Cid, StreamEvents)>>> {
+        let mut task_set = JoinSet::new();
+
+        let chunk_size =
+            (to_process.len() / MAX_STREAM_PROCESSING_TASKS).max(MIN_NUM_STREAMS_PER_BATCH);
```
dav1do, line 625: to avoid spawning many tasks for small sets, we make sure they're at least 25 streams each. this choice was arbitrary and could be 50, 100, etc. with 10000 events found at once, the batches are ~600 streams each.

event-svc/src/event/ordering_task.rs
```diff
@@ -17,6 +21,19 @@ type EventCid = Cid;
 type PrevCid = Cid;
 
 const LOG_EVERY_N_ENTRIES: usize = 10_000;
+/// The max number of tasks we'll spawn when ordering events.
+/// Each stask is given a chunk of streams and wil order them
+const MAX_STREAM_PROCESSING_TASKS: usize = 16;
+/// The minimim number of streams each task will process. If there are many streams, we'll spawn
+/// STREAM_NUM / MAX_STREAM_PROCESSING_TASKS.
+const MIN_NUM_STREAMS_PER_BATCH: usize = 25;
```
nathanielc, line 29: What is `STREAM_NUM`?

nathanielc (approved):

event-svc/src/event/ordering_task.rs
dav1do: oops, that comment makes no sense (and it looks like a const). it was supposed to explain how the batch for each task is determined. I'll fix it.

PR #622, opened 2024-12-03
fix: sqlite optimize during connect could timeout and allow specifying db options

dav1do:
Calling optimize when setting up the connection pool could timeout on big databases. Apparently `acquire_timeout` applies to all statements executed at startup. This also removes the `acquire_timeout` override and uses the default of 30 seconds, continues processing the ordering task loop when there's a "transient" error and, and allows specifying a few sqlite pragmas from the command line which may improve performance (`mmap_size`, `cache_size`, `temp_store`, `analysis_limit`) and a max number of read only connections for the pool (default is 8).

Analysis limit is a bit odd, as it has a default of 1000 and needs to be negative to be disabled. I don't think you can set a clap default to `Some(val)` and have the cli set it to none, but maybe I missed something. If that value is set, we run database optimize at startup and daily in the background.

Importantly, sqlx was reverted to 0.7. Something in the 0.8 upgrade (most likely the libsqlite3-sys version bump) is causing the `pragma optimize` statements to hang forever on the gitcoin database (though apparently not on our test infrastructure). I haven't gotten to the bottom of it, but this is a workaround for now. If this PR is contentious, I can pull that fix and the more trivial changes to logging out.

### DIFF
--- a/one/src/daemon.rs
+++ b/one/src/daemon.rs
@@ -182,7 +182,7 @@ pub struct DaemonOpts {
     )]
     flight_sql_bind_address: Option<String>,
 
-    /// Remote anchor service URL
+    /// Remote anchor service URL. Requires using the experimental-features flag
     #[arg(
         long,
         env = "CERAMIC_ONE_REMOTE_ANCHOR_SERVICE_URL",
@@ -193,6 +193,7 @@ pub struct DaemonOpts {
     /// Ceramic One anchor interval in seconds
     ///
     /// The interval between building a tree for all unanchored events and sending to a CAS
+    /// Requires using the experimental-features flag
     #[arg(
         long,
         default_value_t = 3600,
@@ -202,6 +203,7 @@ pub struct DaemonOpts {
     anchor_interval: u64,
 
     /// Ceramic One anchor batch size
+    /// Requires using the experimental-features flag
     #[arg(
         long,
         default_value_t = 1_000_000,
@@ -214,6 +216,7 @@ pub struct DaemonOpts {
     /// Ceramic One anchor polling interval in seconds
     ///
     /// The interval between requests to cas to determine if the chain transaction is completed.
+    /// Requires using the experimental-features flag
     #[arg(
         long,
         default_value_t = 300,
@@ -224,6 +227,7 @@ pub struct DaemonOpts {
     anchor_poll_interval: u64,
 
     /// Ceramic One anchor polling retry count
+    /// Requires using the experimental-features flag
     #[arg(
         long,
         default_value_t = 12,
--- a/one/src/daemon.rs
+++ b/one/src/daemon.rs
@@ -1,8 +1,8 @@
 use std::{path::PathBuf, time::Duration};
 
 use crate::{
-    default_directory, handle_signals, http, http_metrics, metrics, network::Ipfs, DBOpts, Info,
-    LogOpts, Network,
+    default_directory, handle_signals, http, http_metrics, metrics, network::Ipfs, DBOpts,
+    DBOptsExperimental, Info, LogOpts, Network,
 };
 use anyhow::{anyhow, bail, Context, Result};
 use ceramic_anchor_remote::RemoteCas;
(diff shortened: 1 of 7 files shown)


one/src/daemon.rs
```diff
@@ -323,11 +323,33 @@ fn spawn_database_optimizer(
     mut shutdown: tokio::sync::broadcast::Receiver<()>,
 ) -> tokio::task::JoinHandle<()> {
     tokio::spawn(async move {
+        // Recommended practice is that applications with long-lived database connections should run "PRAGMA optimize=0x10002"
```
smrz2001, line 326: The code and loops reads much better now!

smrz2001 (approved):
🚀

one/src/daemon.rs
```diff
@@ -323,11 +323,33 @@ fn spawn_database_optimizer(
...
+        // when the database connection first opens, then run "PRAGMA optimize" again at periodic intervals - perhaps once per day
+        // https://www.sqlite.org/pragma.html#pragma_optimize
+        match sqlite_pool.run_statement("PRAGMA optimize=0x10002").await {
```
nathanielc, line 329: How is this statement different from `sqlite_pool.optimize()`? Do we need to run this statement once per day as well? The comment implies we do but the code does not.

one/src/daemon.rs
```diff
@@ -323,11 +323,33 @@ fn spawn_database_optimizer(
...
+            }
+            Err(e) => {
+                warn!(error=?e, "failed to run initial database optimize statement");
+            }
+        }
+        let mut duration = std::time::Duration::from_secs(60 * 60 * 24); // once daily
         loop {
-            let mut duration = std::time::Duration::from_secs(60 * 60 * 24); // once daily
+            let mut interval = tokio::time::interval(duration);
+            interval.tick().await; // first tick is immediate
```
nathanielc, line 340: Why do we manually run optimize above and then throw away the first tick? Also why do we recreate the interval within the loop. I would expect to create the interval once and then let it fire immediately to do the first optimization run. But maybe there is a difference between the optimization runs I do not understand.

one/src/lib.rs
```diff
@@ -172,6 +193,25 @@ struct DBOpts {
...
+    /// Set to 0 to disable. None is the default
+    #[arg(long, env = "CERAMIC_ONE_DB_MMAP_SIZE")]
+    db_mmap_size: Option<u64>,
+    /// The maximum number of read only connections in the pool
+    #[arg(long, default_value = "8", env = "CERAMIC_ONE_DB_MAX_CONNECTIONS")]
+    db_max_connections: u32,
+    /// The sqlite temp_store value to use
+    /// 0 = default, 1 = file, 2 = memory
+    #[arg(long, env = "CERAMIC_ONE_DB_TEMP_STORE")]
+    db_temp_store: Option<SqliteTempStore>,
```
nathanielc, line 214: Do we need to expose these to the end operator? Do we expect these to change based on DB size or some other usage pattern?

one/src/daemon.rs
dav1do: yeah, I did it a bit weird didn't I? I changed it so we don't do a first run in the task anymore. It was hanging forever on sqlx 0.8 and preventing the ordering task from getting a writer connection. Now it does one pass in the daemon inline and spawns this task if `analysis_limit` is configured (by default yes). It resets the interval in case we changed how long it should be due to an error, but a sleep or alternative might be better.

one/src/lib.rs
dav1do: Honestly, I'm not sure. I _hope_ we don't expect/need anyone to set them, and we can simply rely on the defaults we've configured, however, they're all documented as things to try for sqlite performance tuning that are very dependent on hardware and database size. This way we can at least make modifications without recompiling the application. Would they make more sense under experimental (tuning) features?

one/src/daemon.rs
dav1do: I added some more comments about it and refactored when/how it's called. It's supposed to be called that way once at startup (`Run ANALYZE on tables that might benefit. On by default`) so it doesn't seem strictly necessary to include but I included it jic.

one/src/lib.rs
nathanielc: >Would they make more sense under experimental (tuning) features?

Yes, lets do this. That way we can iterate on finding good defaults. And hopefully we find good defaults that works for the vast majority of deployments. And only if we cannot find a set of defaults that covers a wide range of deployments do we promote these out of experimental.

one/src/daemon.rs
nathanielc: Agreed, lets just use a sleep since we never read more than two ticks from the interval and we have a dynamic period we want to wait.

nathanielc (approved):
LGTM once the new cli params are experimental.

PR #623, opened 2024-12-03
feat: add support for local file object store

nathanielc:
Testing and developing ceramic-one is much easier with fewer dependencies. This change makes it so that an s3 API compatible object store is not required. Rather objects can be stored on local disk.


one/src/daemon.rs
```diff
@@ -562,54 +572,80 @@ pub async fn run(opts: DaemonOpts) -> Result<()> {
...
-                None
-            };
+                } else {
+                    object_store_url.path().into()
+                };
+                // Create object_store dir if it does not exist.
+                if !tokio::fs::try_exists(&path).await? {
+                    tokio::fs::create_dir(&path).await?;
+                }
+                debug!(path = %path.display(), "object store path");
```
dav1do, line 607: This could maybe be info but debug is probably fine

PR #624, opened 2024-12-04
refactor: remove offset/limit where too generic

nathanielc:
This change removes the offset/limit pagination pattern where it implies a more flexible API than was needed or used. Having the flexible API meant that casts from usize to smaller types for limits and offsets could break pagination assumptions. Now with this change more specific APIs are exposed where its safe to make assumptions about the size of limit offset values in queries.

Additionally in the API where true pagination was the desired API u32 is used instead of usize to keep page sizes small.

Finally, the range_with_values method on the Recon trait was removed as it was unused


dav1do (approved):
Nice, looks like an improvement. Thanks for doing this.

Issue #608, opened 2024-11-20
Add endpoint to Ceramic API for currently connected peers


nathanielc:
## Description

To operate a node well we need to be able to inspect which peers are connected to the local node.

## DOD

- [ ] An API endpoint part of the ceramic API exists that lists connected peers and metadata about the peers
- [ ] did:key of remote peer
- [ ] Last connected time
- [ ] Bytes sent/recv to peer since start (this may be optionally exposed via metrics)
- [ ] Unit tests for the endpoint

PR #618, opened 2024-11-26
feat: adds transitive sharing of peer information

nathanielc:
With this change all the pieces are in place such that peers can discover peers from connected peers and connect to those peers.

Fixes: #605
Fixes: #606

PR #626, opened 2024-12-06
feat: adds /peers endpoint

nathanielc:
With this change a new API endpoint is added for accessing connected peers. The API is similar to the /api/v0/swarm/peers endpoint however it allows for multiple addresses and does not have to remain Kubo RPC compatible.

The endpoint allows posting an address and therefore connecting to a new peer.

Fixes: #608

### DIFF
--- a/api/ceramic.yaml
+++ b/api/ceramic.yaml
@@ -519,6 +519,62 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorResponse"
+  "/peers":
+    options:
+      summary: cors
+      parameters:
+        - name: addresses
+          in: query
+          description: Multiaddress of peer to connect to, at least one address must contain the peer id.
+          schema:
+            type: array
+            items:
+              type: string
+          required: true
+      responses:
+        "200":
+          description: cors
+    get:
+      summary: Get list of connected peers
+      responses:
+        "200":
+          description: success
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Peers"
+        "500":
+          description: Internal server error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorResponse"
+    post:
+      summary: Connect to a peer
+      parameters:
+        - name: addresses
+          in: query
+          description: Multiaddress of peer to connect to, at least one address must contain the peer id.
+          schema:
+            type: array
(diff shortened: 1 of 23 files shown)


api/src/server.rs
```diff
@@ -457,7 +488,7 @@ where
     }
 
     async fn process_events(events: &mut Vec<EventInsert>, event_store: &Arc<M>, node_id: NodeId) {
-        tracing::debug!(count = events.len(), "process_events");
+        trace!(count = events.len(), "process_events");
```
dav1do, line 491: 👍

api/ceramic.yaml
```diff
@@ -519,6 +519,62 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorResponse"
+  "/peers":
+    options:
+      summary: cors
```
dav1do, line 524: It'd be nice to add this on all routes by defining it once but I couldn't figure out a way last time I looked. Nothing to do, just putting it in your head in case you know of one or see it sometime 😀

one/src/network.rs
```diff
@@ -110,12 +112,54 @@ impl<S> Ipfs<S> {
...
+            .map(|(peer_id, addresses)| {
+                Ok(ceramic_api::PeerInfo {
+                    id: NodeId::try_from_peer_id(&peer_id)?,
+                    addresses,
+                })
+            })
+            .collect()
+    }
+    async fn peer_connect(&self, addrs: &[ceramic_api::Multiaddr]) -> Result<()> {
+        // Find peer id in one of the addresses
```
dav1do, line 143: is it possible to make a weird request were you specify addresses with multiple peer IDs? it seems like we'd end up replacing the ID in following addresses or am I misunderstanding?

dav1do (approved):
Looks good, one question about what happens with a weird/bad request with multiple peer IDs

one/src/network.rs
nathanielc: hmm good point, we should check all peers ids are the same. As written it would likely just result in that address failing to connect but better to catch the error sooner.

nathanielc:
Accidentally merged this into #618 I'll back it out.

nathanielc:
New PR here https://github.com/ceramicnetwork/rust-ceramic/pull/628

PR #628, opened 2024-12-06
feat: adds /peers endpoint

nathanielc:
With this change a new API endpoint is added for accessing connected peers. The API is similar to the /api/v0/swarm/peers endpoint however it allows for multiple addresses and does not have to remain Kubo RPC compatible.

The endpoint allows posting an address and therefore connecting to a new peer.

Fixes: #608

### DIFF
--- a/one/src/network.rs
+++ b/one/src/network.rs
@@ -2,12 +2,14 @@
 
 use std::sync::Arc;
 
-use anyhow::Result;
-use ceramic_core::{EventId, Interest, NodeKey, PeerKey};
+use anyhow::{anyhow, Result};
+use async_trait::async_trait;
+use ceramic_core::{EventId, Interest, NodeId, NodeKey, PeerKey};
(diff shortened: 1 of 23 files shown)


one/src/network.rs
```diff
@@ -110,12 +112,54 @@ impl<S> Ipfs<S> {
...
+            .into_iter()
+            .map(|(peer_id, addresses)| {
+                Ok(ceramic_api::PeerInfo {
+                    id: NodeId::try_from_peer_id(&peer_id)?,
+                    addresses,
+                })
+            })
+            .collect()
+    }
+    async fn peer_connect(&self, addrs: &[ceramic_api::Multiaddr]) -> Result<()> {
```
dav1do, line 142: Linking [the comment](https://github.com/ceramicnetwork/rust-ceramic/pull/626/files#r1873671829) from the other PR for visibility. Other than this concern, LGTM.

dav1do (approved):

PR #629, opened 2024-12-06
fix: shutdown ordering tasks with sigint

dav1do:
The ordering tasks wouldn't shutdown gracefully and could take forever. Worst case you killed the process and left a lock on the sqlite file and had to go find another pid to kill it. Now they exit with sigint.

### DIFF
--- a/one/src/daemon.rs
+++ b/one/src/daemon.rs
@@ -411,6 +411,8 @@ pub async fn run(opts: DaemonOpts) -> Result<()> {
     let (shutdown_signal_tx, mut shutdown_signal) = broadcast::channel::<()>(1);
     let signals = Signals::new([SIGHUP, SIGTERM, SIGINT, SIGQUIT])?;
     let handle = signals.handle();
+    debug!("starting signal handler task");
+    let signals_handle = tokio::spawn(handle_signals(signals, shutdown_signal_tx));
 
     // Construct sqlite_pool
     let sqlite_pool = opts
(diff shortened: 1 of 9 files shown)


p2p/src/node.rs
```diff
@@ -1416,7 +1416,9 @@ mod tests {
             let sql_pool = SqlitePool::connect_in_memory().await.unwrap();
 
             let metrics = Metrics::register(&mut prometheus_client::registry::Registry::default());
-            let store = Arc::new(EventService::try_new(sql_pool, true, true, vec![]).await?);
+            let store = Arc::new(
+                EventService::try_new(sql_pool, UndeliveredEventReview::Skip, true, vec![]).await?,
```
dav1do, line 1420: I changed this behavior to skip ordering here but I don't think it is necessary for these tests.

one/src/daemon.rs
```diff
@@ -434,8 +436,19 @@ pub async fn run(opts: DaemonOpts) -> Result<()> {
     let peer_svc = Arc::new(PeerService::new(sqlite_pool.clone()));
     let interest_svc = Arc::new(InterestService::new(sqlite_pool.clone()));
     let event_validation = opts.event_validation.unwrap_or(true);
+    let mut ss = shutdown_signal.resubscribe();
     let event_svc = Arc::new(
```
dav1do, line 440: this whole PR could probably be redone as simply this (which I didn't think of for some reason)... piping it down could allow for a chance to exit gracefully and clean up, but it's more or less implemented like this currently just lower

```rust
    let mut ss = shutdown_signal.resubscribe();
    let event_svc = tokio::select! {
        svc = EventService::try_new(sqlite_pool.clone(), true, event_validation, rpc_providers) => {
            Arc::new(svc?)
        }
        _ = ss.recv() => {
            bail!("received shutdown signal before startup tasks finished")
        }
    };
    ```

PR #634, opened 2024-12-19
chore: use explicit os versions

nathanielc:
This week github updated the `ubuntu-latest` tag to be ubuntu-24.04. This meant that version 0.47.0 of ceramic-one no longer runs on ubuntu-22. This is unexpected and not in our control. With this change we explicitly name the os versions we target (this matters because of glibc versions). This means we will need to add new versions and deprecate old versions as they are released.

### DIFF
--- a/.github/workflows/publish-prerelease.yml
+++ b/.github/workflows/publish-prerelease.yml
@@ -72,7 +72,7 @@ jobs:
           gem install fpm
       - name: Run package script
         run: |
-          ./ci-scripts/package.sh -a ${{ matrix.arch }} -e ${{ matrix.ext }}
+          ./ci-scripts/package.sh -a ${{ matrix.arch }} -e ${{ matrix.ext }} -v ${{ matrix.os }}
       - name: Archive Binary
         uses: actions/upload-artifact@v3
         with:
--- a/.github/workflows/publish-prerelease.yml
+++ b/.github/workflows/publish-prerelease.yml
@@ -20,23 +20,26 @@ jobs:
         # For these target platforms
         include:
           - target: x86_64-unknown-linux-gnu
-            os: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-prerelease.yml
```diff
@@ -73,14 +76,14 @@ jobs:
...
+          name: ceramic-one_${{ matrix.target }}-${{ matrix.os }}
+          path: ceramic-one_${{ matrix.target }}-${{ matrix.os }}.bin.tar.gz
       - name: Archive Debian Package
         uses: actions/upload-artifact@v3
         if: ${{ matrix.ext == 'deb' }}
         with:
-          name: ceramic-one_${{ matrix.target }}
-          path: ceramic-one_${{ matrix.target }}.tar.gz
+          name: ceramic-one_${{ matrix.target }}-${{ matrix.os }}
+          path: ceramic-one_${{ matrix.target }}-${{ matrix.os }}.tar.gz
```
smrz2001, line 86: We'll also need to use the new OS suffix in `ci-scripts/package.sh` when preparing the archive, and in `.github/workflows/update-homebrew.yml` when retrieving the archive SHA.

smrz2001: Actually, wait, let me check that again.

smrz2001: Yes, I think we'll need to update the other two files since the specific OS version is now also part of the archive name.

smrz2001: As well as [this](https://github.com/ceramicnetwork/js-ceramic/blob/develop/ci-scripts/install-rust-ceramic.sh#L20) `js-ceramic` file.

smrz2001 (approved):
🚀

PR #636, opened 2024-12-20
chore: add tests for eip55 cacaos

dav1do:
Added tests using two events from the recent gitcoin signature failures. One is a full carfile parsed as an event, the second uses the envelope/cacao strings and parses them to demonstrate how one could swap in new envelope/cacao strings and make sure it passes (or fails).

Refactored a bit to make it easier to pass in data from a dag-json envelope downloaded from IPFS. I'd like to use the `Signature` type instead of passing protected/signature bytes directly, but I didn't get the deserialization right while testing. There might be an easy way I'm just missing, but for now this is just an internal wrapper.

### DIFF
--- a/validation/src/verifier/event_verifier.rs
+++ b/validation/src/verifier/event_verifier.rs
@@ -1,12 +1,12 @@
 use anyhow::{anyhow, bail, Context as _, Result};
 use ceramic_core::Jwk;
-use ceramic_event::unvalidated::signed;
+use ceramic_event::unvalidated::signed::{self, cacao::Capability};
 use ipld_core::ipld::Ipld;
-use ssi::did_resolve::ResolutionInputMetadata;
+use ssi::{did_resolve::ResolutionInputMetadata, jws::Header};
 
 use super::{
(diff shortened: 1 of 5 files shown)


validation/src/verifier/event_verifier.rs
```diff
@@ -287,4 +320,174 @@ mod test {
...
+        let event = event.as_object().unwrap();
+        // bafyreiago5d674lillsjbdzvmsbt3okwargjgekcevsu7wxuzwfcxz47xa from the protected data
+        let cacao = r#"{
+            "h": {
+              "t": "eip4361"
+            },
+            "p": {
+              "aud": "did:key:zDnaeU3z6wJKYcxdeQdNB1b6bgnUZBtNy6u8W7AT3rVPdwj2f",
+              "domain": "app.passport.xyz",
+              "exp": "2024-12-24T20:27:13.330Z",
```
dav1do, line 417: TODO: update these tests to use a time check before the exp date so these don't start failing next week

smrz2001 (approved):
:shipit:

PR #641, opened 2025-01-14
chore: lockfile v4 (MSRV now 1.78)

nathanielc:
Cargo lockfile v4 is the default in rust 1.83 and has been supported since 1.78 (which is now our MSRV). Our metadata crate needed a dependency bump to work with this. 

The older version of built does not understand new Cargo.lock files.


dav1do (approved):
Haha I did this too: https://github.com/ceramicnetwork/rust-ceramic/pull/639

I can rebase that one for the bench changes (not that we really use it) unless you want to close this one. This does bump MSRV but we don't have a backward compatibility requirement so that seems fine.

nathanielc:
@dav1do Oops, should have checked. Yeah lets merge this one and then we can get the bench fixes after.

PR #643, opened 2025-01-14
refactor: add simpler shutdown handling

nathanielc:
This change adds a shutdown crate that simplifies shutdown handling. The API removes the need to create as many async move blocks and clone broadcast channels.


dav1do (approved):
nice :shipit:

PR #644, opened 2025-01-14
feat: adds actor framework

nathanielc:
This change adds a lightweight actor framework. The main features that are important to our use case are strongly typed messages and preserving of tracing spans.

### DIFF
--- a/actor-macros/src/lib.rs
+++ b/actor-macros/src/lib.rs
@@ -1,0 +1,137 @@
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::{parse_macro_input, Attribute, DeriveInput, GenericParam, Lit};
+
+#[proc_macro_derive(Actor, attributes(actor))]
+pub fn actor(item: TokenStream) -> TokenStream {
+    let item = parse_macro_input!(item as DeriveInput);
+    // Extract struct name
+    let struct_name = item.ident;
+
+    let Config {
+        trait_name,
+        envelope_name,
+        handle_name,
+    } = Config::from_attributes(&struct_name, &item.attrs);
+
+    let generics = item.generics;
+    let generic_types: Vec<_> = generics
+        .params
+        .iter()
+        .filter_map(|param| match param {
+            GenericParam::Type(type_param) => Some(&type_param.ident),
+            _ => None,
+        })
+        .collect();
+    let phantom_fields = generic_types.iter().map(|ty| {
+        let name = syn::Ident::new(&format!("__{}", ty).to_lowercase(), ty.span());
+        quote! {
+            #name: std::marker::PhantomData<#ty>
+        }
+    });
+    let phantom_values: Vec<_> = generic_types
+        .iter()
+        .map(|ty| {
+            let name = syn::Ident::new(&format!("__{}", ty).to_lowercase(), ty.span());
+            quote! {
+                #name: Default::default()
+            }
+        })
+        .collect();
+
+    // Generate the implementation
+    let expanded = quote! {
+        impl #generics ceramic_actor::Actor for #struct_name < #(#generic_types,)*> {
+            type Envelope = #envelope_name;
+        }
+        impl #generics #trait_name for #struct_name < #(#generic_types,)*> { }
+
+        impl #generics #struct_name < #(#generic_types,)*> {
+            /// Start the actor returning a handle that can be easily cloned and shared.
+            /// The actor stops once all handles are dropped.
+            pub fn spawn(size: usize, actor: impl #trait_name + ::std::marker::Send + 'static, shutdown: impl ::std::future::Future<Output=()> + ::std::marker::Send + 'static) -> (#handle_name < #(#generic_types,)*>, tokio::task::JoinHandle<()>) {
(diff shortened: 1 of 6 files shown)


actor/src/macros.rs
```diff
@@ -0,0 +1,122 @@
...
+        $enum_name:ident,
+        $trait_name:ident,
+        $(
+            $variant_name:ident => $message_type:ty,
+        )*
+    ) => {
+        /// Wrapping envelope for all messages of the actor.
+        pub enum $enum_name {
+            $(
+                #[doc = stringify!($variant_name)]
```
dav1do, line 20: rust analyzer seems to have a hard time displaying these `#[doc]` macro comments. I hoped something like this would make a difference but no luck

```rust
#[doc = concat!(
                    stringify!($variant_name),
                    " contains messages of type [`",
                    stringify!($message_type),
                    "`]."
                )]
```

actor/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+    type Result: 'static + Send;
+
+    /// Returns the messages' type name string
+    fn type_name() -> &'static str {
+        std::any::type_name::<Self>()
+    }
+}
+
+/// Handle a type of [`Message`] for an [`Actor`].
+#[async_trait]
```
dav1do, line 37: I hoped we might be able to remove `async_trait` from this crate now that we have a higher MSRV but it's a bit annoying so probably not worth it right now.

actor-macros/src/lib.rs
```diff
@@ -0,0 +1,137 @@
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::{parse_macro_input, Attribute, DeriveInput, GenericParam, Lit};
+
+#[proc_macro_derive(Actor, attributes(actor))]
+pub fn actor(item: TokenStream) -> TokenStream {
```
dav1do, line 6: It might be nice to add some comments on this explaining the default names/attributes/example usage, but the example does a good job so I'm not too worried about it now.

actor-macros/src/lib.rs
```diff
@@ -0,0 +1,137 @@
...
+    let expanded = quote! {
+        impl #generics ceramic_actor::Actor for #struct_name < #(#generic_types,)*> {
+            type Envelope = #envelope_name;
+        }
+        impl #generics #trait_name for #struct_name < #(#generic_types,)*> { }
+
+        impl #generics #struct_name < #(#generic_types,)*> {
+            /// Start the actor returning a handle that can be easily cloned and shared.
+            /// The actor stops once all handles are dropped.
+            pub fn spawn(size: usize, actor: impl #trait_name + ::std::marker::Send + 'static, shutdown: impl ::std::future::Future<Output=()> + ::std::marker::Send + 'static) -> (#handle_name < #(#generic_types,)*>, tokio::task::JoinHandle<()>) {
```
dav1do, line 52: do we care to return the `JoinHandle` here? It seems like we'll pretty much just drop it and rely on the actor Handle to control the task. Not really a big deal, maybe it's more important in the future usage.

actor/src/macros.rs
```diff
@@ -0,0 +1,122 @@
...
+        impl std::convert::From<($message_type, tokio::sync::oneshot::Sender<<$message_type as $crate::Message>::Result>)> for $enum_name {
+            fn from(value: ($message_type, tokio::sync::oneshot::Sender<<$message_type as $crate::Message>::Result>)) -> Self {
+                Self::$variant_name(value.0, value.1)
+            }
+        }
+        impl std::convert::TryFrom<$enum_name> for $message_type {
+            type Error = ();
+            /// Extracts the message from the envelope.
+            ///
+            /// # Panics
```
dav1do, line 111: This isn't true now, is it?

dav1do (approved):
A few small things but overall looks good. The proc macro won't work with lifetimes but all the actors need to be `'static` so that makes sense.

actor/src/macros.rs
nathanielc: yeah I saw this too, I found this https://users.rust-lang.org/t/trying-to-generate-doc-comments-with-macros/81783, but it doesn't seem to work. And it doesn't complain either.

actor/src/lib.rs
nathanielc: Yeah, I get a warning that async in public traits is not recommended yet. So I used async_trait.

actor-macros/src/lib.rs
nathanielc: Good call

actor-macros/src/lib.rs
nathanielc: The examples do not use it, but the pipeline code does as well as test code.

actor/src/macros.rs
nathanielc: good catch, used to be just From hence the comment.

actor/src/macros.rs
nathanielc: I haven't been able to figure this out. The generated docs are not that important. Maybe this a regression in recent versions of rust? Either way I am going to leave it in place for now.

PR #645, opened 2025-01-14
feat: stream flight sql queries

nathanielc:
With this change it is now possible to query _stream variant tables that produce an unbounded stream of new data as it arrives in the tables.

This change was achieved by refactoring the pipeline crate to use an actor model. This makes it possible for events to be passed between actors efficiently and with low latency.


dav1do (commented):
This is a lot to review but I don't have anything major right (but I haven't totally wrapped my head around all of it 😅).. was going to leave it open a day or two in case others have a chance to look. 

I see how the actors are able to subscribe to each other but I'm missing how it does (or should?) happen for consumers. Is there a way to get a stream via flight sql already or does that need to be added?

pipeline/src/concluder/mod.rs
```diff
@@ -0,0 +1,307 @@
...
+                            arrow::array::Array::is_valid(&index_col, 0).then(|| index_col.value(0))
+                        })
+                    })
+                })
+            }
+        };
+        ctx.session()
+            .register_table(
+                CONCLUSION_EVENTS_STREAM_TABLE,
+                Arc::new(StreamTable::new(handle.clone())),
```
nathanielc, line 105: This is how consumers can get a stream of data. The `StreamTable` implements the datafusion traits to be a table, then it calls into the actors (via the handle) to subscribe to that actors data. 

There are a few other touch points I'll call out below.

pipeline/src/since/stream.rs
```diff
@@ -0,0 +1,213 @@
...
+        &self,
+        _partition: usize,
+        _context: Arc<datafusion::execution::TaskContext>,
+    ) -> datafusion::error::Result<datafusion::execution::SendableRecordBatchStream> {
+        let source = self.source.clone();
+        let projection = self.projection.clone();
+        let offset = self.offset as u64;
+        let limit = self.limit;
+        let stream = try_stream! {
+            let mut stream = source.subscribe_since(projection, Some(offset), limit).await?;
```
nathanielc, line 188: Here we are inside an `execute` method that gets called during a datafusion query. The implementation calls `subscribe_since` on its source (which is an actor handle).

pipeline/src/concluder/mod.rs
```diff
@@ -0,0 +1,307 @@
...
+    fn schema(&self) -> SchemaRef {
+        crate::schemas::conclusion_events()
+    }
+    async fn subscribe_since(
+        &self,
+        projection: Option<Vec<usize>>,
+        offset: Option<u64>,
+        limit: Option<usize>,
+    ) -> anyhow::Result<SendableRecordBatchStream> {
+        Ok(self
```
nathanielc, line 299: Here is where the Concluder implements the `subscribe_since` method on from the source. All it does is send the `SubscribeSinceMsg` to the actor. 

The Aggregator implements the same trait in the same way and registers an `event_states_stream` table. This pattern is repeatable so that any new actors can easily expose a `_stream` suffixed table if needed.

pipeline/src/concluder/mod.rs
```diff
@@ -0,0 +1,307 @@
...
+                    .table(CONCLUSION_EVENTS_TABLE)
+                    .await?
+                    .aggregate(
+                        vec![],
+                        vec![datafusion::functions_aggregate::min_max::max(col("index"))
+                            .alias("max_index")],
+                    )?
+                    .collect()
+                    .await?;
+                batches.get(0).and_then(|batch| {
```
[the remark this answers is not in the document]
nathanielc: Thanks, yeah clippy complained about another one of these earlier, not sure why it didn't complain about this one. Maybe has to do with it being test code? Either way its fixed.

flight/tests/server.rs
```diff
@@ -149,60 +157,235 @@ fn events(start_index: u64) -> Vec<ConclusionEvent> {
...
+                "index",
+                cid_string(stream_cid) as stream_cid,
+                stream_type,
+                controller,
+                dimensions,
+                cid_string(event_cid) as event_cid,
+                event_type,
+                data::varchar as data,
+                array_cid_string(previous) as previous
+            FROM conclusion_events_stream LIMIT 2"#
```
nathanielc, line 281: Here is a test that shows you can query the _stream table to get the data. Of note in this test is there is no logic to wait till the test harness has written the data, as that _is_ the feature of the stream table. It is a blocking query until the limit is reached. The non stream table queries have explicit logic to wait until the test harness knows the data is written before querying.

api/src/server.rs
```diff
@@ -876,130 +878,47 @@ where
...
-                            ErrorResponse::new(format!(
-                                "failed to define last_value event_cid dimensions: {err}"
-                            ))
-                        })?
-                        .alias("dimensions"),
-                ],
-            )
-            .map_err(|err| ErrorResponse::new(format!("failed to define window: {err}")))?
-            .collect()
+        let state = aggregator
```
nathanielc, line 884: And here is how our own code can consume the actors. This is not subscribing to the actor, just making a stream state query.

dav1do: Yeah, this make sense (and is nicer than before), which means we can write a query against the flight sql endpoint that will wait until the data exists. I guess I'm wondering about SSE/WS updates. Could we provide a new route that spawns a task for the connection and uses `SubscribeSinceMsg` to receive updates and then push them out or am I thinking about this wrong?

nathanielc: The plan is to use FlightSQL native streaming support instead of SSE or WS. I have a WIP PR I hope to get up today/tomorrow that adds support for these queries to the SDK. 

We could add SSE/WS support if there is enough demand for it, but I expect/hope that FlightSQL is sufficient.

nathanielc: I pushed up what I have here https://github.com/ceramicnetwork/ceramic-sdk/pull/48

dav1do: nice, that makes sense

actor-macros/src/lib.rs
```diff
@@ -0,0 +1,178 @@
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::{parse_macro_input, Attribute, DeriveInput, GenericParam, Lit};
+/// Derive the `Actor` trait for the struct.
```
dav1do, line 4: 👍

actor/src/lib.rs
```diff
@@ -0,0 +1,222 @@
+//! Actor provides a lightweight actor framework based on tokio.
+//!
+//! Features:
+//!  - In memory message passing
+//!  - Strongly typed
+//!  - tracing spans are preserved from caller to actor
+//!
+//! # Example
+//! ```
+#![doc = include_str!("../examples/game/main.rs")]
```
dav1do, line 10: 👍

dav1do (approved):
I don't have anything more right now. It looks like it needs a rebase/merge but I'll approve and leave it up to you to decide if you want more input. The pattern makes sense. We may need to adjust a bit as we add more functionality but that's reasonable.

PR #646, opened 2025-01-21
feat: add pipeline metrics

### LINKED ISSUE #582: Add metrics to pipeline steps
DOD
- [ ] Metrics for counts of data written into each pipeline table
- [ ] Metrics for the size of each batch written into each pipeline table

nathanielc:
Adds metrics for the pipeline system. Support was added to the actor framework to directly support metrics.  

Fixes #582


actor/src/macros.rs
```diff
@@ -47,15 +51,23 @@ macro_rules! actor_envelope {
...
         }
-        #[doc = std::stringify!($trait_name)]
+        #[doc = std::stringify!($actor_trait)]
         #[doc = " is an [`crate::Actor`] and [`crate::Handler`] for each message type in the actor envelope "]
         #[doc = stringify!($enum_name)]
         #[doc = "."]
-        pub trait $trait_name : $crate::Actor<Envelope = $enum_name> $( + $crate::Handler<$message_type> )* { }
+        pub trait $actor_trait : $crate::Actor<Envelope = $enum_name> $( + $crate::Handler<$message_type> )* + ::std::marker::Send + 'static { }
+
+        #[doc = std::stringify!($actor_trait)]
```
dav1do, line 60: this should be`$recorder_trait`

pipeline/src/concluder/mod.rs
```diff
@@ -187,12 +196,15 @@ async fn poll_new_events(
     handle: ConcluderHandle,
     ctx: SessionContextRef,
     mut last_processed_index: Option<u64>,
+    metrics: Metrics,
     mut shutdown: ShutdownSignal,
 ) -> anyhow::Result<()> {
     let mut interval = interval(Duration::from_millis(1_000));
+    interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
```
dav1do, line 203: 👍

dav1do (approved):
Nice. The generic bounds on the Actor trait are a bit rough, but the requirements/impl is captured by the macros so it shouldn't be necessary to think about/look at them very often. :shipit:

PR #647, opened 2025-01-24
fix: fix off by one, projection and not found bugs

nathanielc:
There was an off by one bug in translating highwater marks to offsets.

There was a bug in projecting stream tables as they reported the wrong schema.

Finally there was a bug in the stream state query that would panic if the stream didn't exist and datafusion returned an empty batch.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,2 @@
-[build]
+[target.'cfg(all())']
 rustflags = ["--cfg", "tokio_unstable"]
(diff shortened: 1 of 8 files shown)


.cargo/config.toml
```diff
@@ -1,2 +1,2 @@
-[build]
```
nathanielc, line 1: This is effectively a noop change but it makes using a custom linker easier.  TL;DR rustflags are only cumulative within the same config section (i.e. target vs build).

And FWIW using mold has been very fast. Still wish the whole project would build faster but I'll take any improvements I can get.

dav1do (approved):

PR #648, opened 2025-01-27
fix: conclusion poll loop no longer stops

nathanielc:
The bug was a misunderstanding of how select! branches work. We were stuck forever in a select branch because we did not handle the Ok case. This has been fixed and we have tests.

### DIFF
--- a/pipeline/src/concluder/mod.rs
+++ b/pipeline/src/concluder/mod.rs
@@ -29,7 +29,7 @@ use tokio::{
     task::JoinHandle,
     time::{interval, MissedTickBehavior},
 };
-use tracing::{debug, error, warn};
+use tracing::{debug, error, info, warn};
 
 use crate::{
     metrics::Metrics,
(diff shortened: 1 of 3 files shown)


pipeline/src/concluder/mod.rs
```diff
@@ -246,8 +265,10 @@ async fn poll_new_events(
                             _ = &mut shutdown => {
                                 return Ok(());
                             },
-                            Err(err) = handle.notify(NewEventsMsg { events: batch }) => {
-                                warn!(?err,"failed to notify concluder about new events");
+                            r = handle.notify(NewEventsMsg { events: batch }) => {
```
dav1do, line 268: Ah, ouch. I didn't notice the missing Ok branch. I was thinking it'd panic when you missed something but it's perfectly happy to ignore values as long as not all branches are disabled. 

```
The select! macro panics if all branches are disabled and there is no provided else branch. A branch is disabled when the provided if precondition returns false or when the pattern does not match the result of <async expression>.
```

I had to think about this because the ok still happened and would have gone into the channel.. but now I see that the branch is disabled so the select doesn't return and the loop turned into send one successful message and wait for shutdown. `Select!` seems straightforward but it can be pretty subtle. 

From the docs, this first sentence makes it seem like it'd have worked as is but if you keep going it includes the nuance (especially in the full lifecycle section). 

```
Waits on multiple concurrent branches, returning when the first branch completes, cancelling the remaining branches.
```

dav1do (approved):

PR #651, opened 2025-01-28
feat: adds stream_id_string udf

nathanielc:
This UDF is helpful in decoding the `model` dimension as it is a stream id bytes.

For example this query:

```sql
```


nathanielc:
Wrong changes, new PR up in a sec.

PR #652, opened 2025-01-29
fix: use invoke_batch on UDFs as invoke is deprecated

nathanielc:
Datafusion recently deprecated the invoke method on ScalarUDFs. This change updates to be following best practices going forward.


dav1do (approved):
:shipit:

PR #653, opened 2025-01-29
feat: adds stream_id_string udf

nathanielc:
This UDF is helpful in decoding the model dimension as it is a stream id bytes.

For example this query:

```sql
```

### DIFF
--- a/pipeline/src/stream_id_string.rs
+++ b/pipeline/src/stream_id_string.rs
@@ -1,0 +1,387 @@
+//! Provides a scalar udf implementation that converts StreamId bytes to a utf8 string.
+
+use std::{any::Any, sync::Arc};
+
+use arrow::{
+    array::{AsArray as _, GenericByteArray},
+    datatypes::{GenericBinaryType, GenericStringType},
+};
+use ceramic_core::StreamId;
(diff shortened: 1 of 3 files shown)


pipeline/src/stream_id_string.rs
```diff
@@ -0,0 +1,378 @@
...
+    fn name(&self) -> &str {
+        "array_stream_id_string"
+    }
+    fn signature(&self) -> &Signature {
+        &self.signature
+    }
+    fn return_type(&self, _args: &[DataType]) -> datafusion::common::Result<DataType> {
+        Ok(DataType::new_list(DataType::Utf8, true))
+    }
+    fn invoke(&self, args: &[ColumnarValue]) -> datafusion::common::Result<ColumnarValue> {
```
dav1do, line 182: probably want to update this to use `invoke_batch` too?

dav1do (approved):

pipeline/src/stream_id_string.rs
nathanielc: ah yep, thanks. So many of these floating around I missed this one.I'll get it updated

PR #654, opened 2025-01-29
feat: add dimension_extract udf

nathanielc:
Simplifies a query like

```sql
```
to

```sql
SELECT stream_id_string(dimension_extract(dimensions, 'model')) FROM event_states WHERE ...
```


dav1do (approved):
Nice, looks like a clippy lint to fix but otherwise LGTM.

PR #657, opened 2025-01-30
refactor: more efficient list udf impls

nathanielc:
This is a more efficient list UDF implementation when the list structure does not change. We simply map over the values array directly and reuse the list offsets and nulls.

This is not strictly necessary. However setting a standard to always operate on a single Arrow array at a time is kinda the whole point of Arrow. Therefore I thinks is valuable to make this change so we set a good pattern to follow in the future.

### DIFF
--- a/pipeline/src/cid_string.rs
+++ b/pipeline/src/cid_string.rs
@@ -2,12 +2,11 @@
 
 use std::{any::Any, sync::Arc};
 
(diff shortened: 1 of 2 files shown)


kreoworld (approved):
:shipit:

smrz2001 (approved):
🚀

PR #658, opened 2025-02-03
feat: partition event states by event cid

nathanielc:
BREAKING CHANGE:

This change is breaking to the structure of the event_states table in object store. With this change any existing data in the event_states object store is invalidated. On the next startup the data will be recomputed from the conclusion_events table. It is recommended once a successful upgrade is complete to delete the old parquet files for the event_states table from object store, as that data will no longer be read in its current form. The old parquet files will be in the `ceramic/v0/event_states/` directory. The new parquet files are in `ceramic/rev1/event_states/` directory.

This change achieve two main improvements:

1. By partitioning the event_states table by the event_cid the join to aggregate a new event with its previous event is much more efficient.
2. In some cases previously the join would fail as it was a very large NestedLoopJoin. A nested loop join would need to compute a cross product of left and right sides. Given the batch size is 10K rows that is a output product of 100M rows. This large structure would either consume huge amounts of memory or simply fail if there was not enough memory. With this change we now join on the partition columns and can use a more efficient join implementation which avoids the failure.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2478,7 +2478,6 @@ name = "ceramic-one"
 version = "0.48.0"
 dependencies = [
  "anyhow",
- "arrow",
  "arrow-cast",
  "arrow-flight",
  "async-stream",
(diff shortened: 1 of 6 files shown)


kreoworld (approved):
🚢

smrz2001 (approved):
🚀

PR #659, opened 2025-02-03
feat: adds output format options to query command

nathanielc:
With this change it is now possible to output the results of queries in machine friendly ways.


smrz2001 (approved):
🚀

PR #661, opened 2025-02-10
chore: use artifacts v4 as v3 is deprecated

nathanielc:
We did not use v3 in any of the ways that break with v4 https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md


smrz2001 (approved):
🚀

PR #662, opened 2025-02-12
feat: add/update UDFs for the ceramic pipeline

nathanielc:
These UDFs do simple transformations on data. These are all used as part of schema validation. They are added separately to keep the set of changes smaller.


smrz2001 (approved):
LGTM! :shipit:

PR #670, opened 2025-03-10
chore: do not deploy to durable envs

nathanielc:
We do not deploy to durable envs, however we continue to publish the docker images with their tags. The old unused publish-image.yml was replace with the cd-to-infra.yml removing the cd bits.


smrz2001 (approved):
🚀  LGTM

As always with GitHub workflows, there might be issues when we run them. If so, can fix in follow-up PRs.

PR #673, opened 2025-03-12
fix: make sdk use correct id in data event payloads

nathanielc:
Prior to this change the sdk used the previous cid as the id in the header of data events.

Additionally the update test were flaky so they were ported to use the sdk (which is how the bug was found).


tests/suite/src/__tests__/correctness/fast/model-sync.test.ts
```diff
@@ -33,108 +33,107 @@ const testModel: ModelDefinition = {
...
+interface Clients {
+  url: string,
+  ceramic: CeramicClient,
+  flight: FlightSqlClient,
+  model: ModelClient,
+  modelInstance: ModelInstanceClient,
 }
 
 
-describe('model integration test for list model and MID', () => {
```
nathanielc, line 56: this test had overlap with the update test, but the update test was dynamic for the number of ceramic nodes. Now this is as well.

smrz2001 (approved):
🥳

Issue #680, opened 2025-03-17
BUG: 0.53.0 extreme memory usage on aggregator initialisation


m0ar:
(as discussed in the discord, adding here for visibility)

**Describe the bug**
0.53.0 runs the event aggregator by default, and it makes the service extremely memory hungry during the first initialisation. We could barely start it without k8s OOM-killing either it, or other pods on the node to make space for it. 

This is a problem mainly because it limits the ability for the community to run nodes on more frugal hardware, if the node fails to start due to being memory capped as the event set grows.

The screenshot was the highest I could visually document, I think it was able to pass after peaking at about 16GB of memory use. 

**Expected behavior**
Ideally, the aggregator uses chunking/streaming/disk/temp tables to avoid having to load up so much data directly in memory.

**Screenshots**
![Image](https://github.com/user-attachments/assets/baab241c-32da-4590-8a61-20aeba5f5b63)

**Ceramic versions** 
0.53.0

**Machine, OS, browser information**
Amazon Linux, EC2 t3a.2xlarge, official rust-ceramic container.

PR #682, opened 2025-03-18
fix: remove distinct step as its prohibitively expensive

nathanielc:
The distinct step when fetching pending events was to correct for the case when we failed to delete old pending events. However it was prohibitively expensive consuming many GBs of memory for roughly 200K pending event rows.

There are few reasons why the distinct was too aggregisve and no the best solution to the duplicate problem.

1. While duplicate pending events can arise seeing the same model cid twice is not possible. This means that in worst case scenario pending events get processed once but fail to be deleted, in which case they remain in the pending events table forever. But those duplicates will not result in event state rows
2. With retries (now added) the likely hood of duplicates pending events is very small and not worth paying the de-duplication cost on every read of the table.

With this change we retry any failures to delete old pending events and explicitly log any errors. We no longer de-duplicate pending events on read. In a worst case scenario an operator can delete the event_states and pending_event_states tables and rebuild all the data to remove the pending events.

Fixes: #680

### DIFF
--- a/pipeline/src/aggregator/mod.rs
+++ b/pipeline/src/aggregator/mod.rs
@@ -369,8 +369,7 @@ impl Aggregator {
         //  3. Delete previously existing files.
         //
         //  If we fail in between any of these steps we will end up with duplicate rows, but not
-        //  missing rows. The logic above can handles the duplicate rows explicitly.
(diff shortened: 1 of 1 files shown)


smrz2001 (approved):
🚀

smrz2001 (approved):
🚀

PR #684, opened 2025-03-19
chore: use different gnosis rpc url

nathanielc:
Of the ones found on https://docs.gnosischain.com/tools/RPC%20Providers/ this is the only one that doesn't error when looking up transactions by hash. All the other would occasionally report that the transaction was not found causing sync to fail.


smrz2001 (approved):
🚀

PR #685, opened 2025-03-19
feat: add some validation logic to migration from-ipfs

nathanielc:
Since ceramic-one nodes refuse to sync with nodes that share bad data its important that the migration process ensure no bad data is migrated.

This change adds the ability to validate signatures and chain ids skipping events that do not pass validation.

Additionally you can also filter events to only events for a specific model.

Finally the pipeline aggregator is more forgiving of bad data, recording validation errors for individual events instead of failing the whole aggregator pipeline.


smrz2001 (approved):
:shipit:

PR #687, opened 2025-03-19
feat: conflict resolution scaffolding

nathanielc:
The meat of this PR can be found in its comments. There are several TODO comments. At a high level there are two halfs to completing conflict resolution:

1. Annotate conclusion events with their _before_ times.
2. Implement a conflict resolver UDF that uses the set of event tips to determine the winning tip.

The conflict resolver will need the _before_ time data as well as the _previous_ fields in addition to the data already present on the event_states table. Getting _previous_ field will be straightforward as the conclusion_events already have that data but its not preserved into the event_states table.

Annotation the conclusion events with their _before_ times has two challenges. First while we validate events we promptly forget the block time stamp of the event time proof. We need to preserve that information. Second its not clear how to propagate the _before_ conclusion from time events to data events if that is even needed. I would recommend starting on part two to understand exactly what data the resolver UDF needs and then work backwards to the concluder to determine how to get that information.


pipeline/src/concluder/mod.rs
```diff
@@ -169,7 +169,6 @@ actor_envelope! {
     ConcluderRecorder,
     NewEvents => NewEventsMsg,
     SubscribeSince => SubscribeSinceMsg,
-    EventsSince => EventsSinceMsg,
```
nathanielc, line 172: Note this was dead code, so I removed it to keep things simpler

pipeline/src/concluder/event.rs
```diff
@@ -110,7 +110,30 @@ pub struct ConclusionTime {
...
+    // stream.
+    //
+    // Challenges: The anchor proof information is gather when validating the event but then
+    // forgotten. It likely needs to be persisteted into a new sqlite table and then the conclusion
+    // feed call needs to join against that table to produce these fields.
+    //
+    // How do we populate that table for existing data?
+    // Likely need some kind of explicit migration command that can be run to populate the table.
+    /// Proof of time event anchoring
+    pub time_proof: Option<TimeProof>,
```
stbrody, line 127: does this need to be optional?  Since this is in a `ConclusionTime` struct, seems like there should always be `TimeProof`

pipeline/src/schemas.rs
```diff
@@ -190,3 +194,148 @@ pub fn pending_event_states() -> SchemaRef {
...
+                        false,
+                    ))])
+                    .collect::<Vec<_>>(),
+            ))
+            .finish(),
+        )
+    }))
+}
+
+/// The `stream_states` table contains the canonical tip for each stream
```
stbrody, line 269: does this mean after conflict resolution?

pipeline/src/resolver/mod.rs
```diff
@@ -0,0 +1,615 @@
...
+#[derive(Debug)]
+pub struct StreamStateMsg {
+    /// Id of the stream
+    pub id: StreamId,
+}
+impl Message for StreamStateMsg {
+    type Result = anyhow::Result<Option<StreamState>>;
+}
+
+/// State of a single stream
```
stbrody, line 447: I'm surprised we don't already have a type for this?  Don't we already have a stream_states table that must have a type for this?

pipeline/src/resolver/mod.rs
```diff
@@ -0,0 +1,615 @@
+//! Conflict resolution functions for Ceramic Model Instance Document streams.
```
stbrody, line 1: this file scares me 😅

smrz2001: Me too haha

pipeline/src/resolver/mod.rs
smrz2001: The pipeline naming is generally a bit confusing, and also, there are two instances of this struct, one in the Aggregator (`aggregator.rs`) and another in the Resolver (`resolver.rs`).

In the Aggregator (which applies commit patches), it's used with the `event_states` table, while in the Resolver, it's used with the `stream_states` table.

I could rename it to `EventState` in the Aggregator. It isn't _really_ the _stream_ state at that stage of the pipeline because the true stream state can only be determined after conflict resolution in the Resolver.

WDYT?

event-svc/src/event/service.rs
```diff
@@ -518,6 +548,40 @@ impl EventService {
             Ok(())
         }
     }
+
+    /// This is a helper function for migrations to get the chain proof for a given event from the database,
```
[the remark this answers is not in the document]
smrz2001: Yeah, this function isn't used in migrations at all, and is only used in the pipeline. I'll just simplify the comment.

event-svc/src/event/service.rs
```diff
@@ -574,6 +638,12 @@ pub enum ValidationError {
         key: EventId,
         reason: String,
     },
+    /// 'Soft error' -> should not kill recon conversation but should not be persisted
+    /// A time event could not be validated because no RPC provider was available
```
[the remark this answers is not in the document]
smrz2001: I'm considering removing this error (and actually have for now but let's discuss).

Do we want to let the Recon conversation continue if a Time Event couldn't be validated?
- If it's a transient issue (like RPC rate limiting), then it'll eventually go away.
- If it's a Time Event for which the node _should_ have a RPC endpoint configured but doesn't, then we want the failure to be loud so that the operator fixes their configuration.
- If it's a bad Time Event, then the Recon conversation _should_ be terminated.

WDYT?

event-svc/src/event/service.rs
```diff
@@ -518,6 +548,40 @@ impl EventService {
...
+    ) -> std::result::Result<ChainProof, crate::eth_rpc::Error> {
+        let tx_hash = event.proof().tx_hash();
+        let tx_hash = tx_hash_try_from_cid(tx_hash).unwrap().to_string();
+        let proof = self
+            .event_access
+            .get_chain_proof(event.proof().chain_id(), &tx_hash)
+            .await
+            .map_err(|e| crate::eth_rpc::Error::Application(e.into()))?;
+
+        if let Some(proof) = proof {
```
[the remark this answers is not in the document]
smrz2001: You're right.

With my other changes, we can simplify this function because if a Time Event wasn't available at this point in the pipeline, then something is wrong. Time Events should have been validated prior to this stage.

pipeline/src/resolver/mod.rs
stbrody: definitely think they should have different names if they have different meanings.  Maybe this one is `ResolvedStreamState` or something like that?

event-svc/src/event/validator/event.rs
```diff
@@ -133,46 +136,43 @@ impl EventValidator {
...
-            });
-        };
-
         let mut validated = ValidatedEvents::new_with_expected_valid(parsed_events.len());
-        // partition the events by type of validation needed and delegate to validators
+
+        // Partition the events by type of validation needed and delegate to validators
         let grouped = GroupedEvents::from(parsed_events);
 
-        let (validated_signed, validated_time) = try_join!(
```
stbrody, line 169: not worth changing now, but maybe add a TODO to make data and time event validation concurrent again.  Since time event validation is mostly network i/o bound and data event validation as mostly CPU bound, they should parallelize pretty effectively.

event-svc/src/event/validator/time.rs
```diff
@@ -25,6 +17,7 @@ pub struct TimeEventValidator {
     /// we could support multiple providers for each chain (to get around rate limits)
     /// but we'll just force people to run a light client if they really need the throughput
     chain_providers: HashMap<caip2::ChainId, ChainInclusionProvider>,
+    // add a sql connection / block table access
```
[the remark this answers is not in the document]
stbrody: ping

event-svc/src/store/sql/entities/chain_proof.rs
```diff
@@ -0,0 +1,33 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, sqlx::FromRow)]
+/// A chain inclusion proof stored in the database
+/// Unique by chain_id and transaction_hash
+pub struct ChainProof {
+    /// The chain ID of the proof
+    pub chain_id: String,
+    /// The transaction hash of the proof
+    pub transaction_hash: String,
+    /// The transaction input of the proof
```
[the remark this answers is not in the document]
stbrody: ping

pipeline/src/schemas.rs
```diff
@@ -185,8 +193,155 @@ pub fn pending_event_states() -> SchemaRef {
...
```
stbrody, line 275:
```suggestion
/// The `stream_states` table contains the canonical tip for each stream (after conflict resolution has selected from all divergent histories)
```

stbrody (approved):
overall lgtm.  Left a few small comments to clean up, but they're minor.  Feel free to merge once addressed

event-svc/src/event/service.rs
```diff
@@ -518,6 +540,24 @@ impl EventService {
             Ok(())
         }
     }
+
+    /// This is a helper function to get the chain proof for a given event from the database, or to validate and store
```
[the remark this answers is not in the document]
smrz2001: Ah right, I changed the function. Updating the comment.

event-svc/src/event/validator/time.rs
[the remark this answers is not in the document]
smrz2001: Another "hidden comment" lol.

I added a TODO. I'll revisit in the next PR. I'll have to think some more about whether this is needed.

PR #689, opened 2025-03-20
chore: add CI workflows to publish sdk

nathanielc:


.github/workflows/publish-sdk.yml
```diff
@@ -0,0 +1,149 @@
...
+        run: pnpm artifacts
+      - name: List packages
+        working-directory: ${{ env.FLIGHT_SQL_PATH }}
+        run: ls -R ./npm
+        shell: bash
+      - name: Publish
+        working-directory: sdk
+        run: |
+          COMMIT_MESSAGE=$(git log -1 --pretty=%B)
+          if [[ $(echo "$COMMIT_MESSAGE" | grep 'chore(sdk): version v') ]]; then
```
smrz2001, line 136: Should we move this step earlier on in the workflow so that other steps don't run if this isn't a release commit?

smrz2001 (approved):
LGTM 🚀

If there are issues later, I'll take care of them.

PR #690, opened 2025-03-20
fix: make apply patch not fail the pipeline

nathanielc:
With this change events with bad patches or data will just end up with null states instead of preventing the entire pipeline from processing more data.


smrz2001 (approved):
:shipit:

PR #695, opened 2025-03-21
Improved docs and logging for `migrations from-ipfs`

m0ar:
Heya :wave: 

 I did a pass over the docs and logging to make it a bit easier to understand what's going on, after being confused more times than I care to admit with failing block imports (mostly related to incorrect paths in the `input_file_list` causing silent skips of the entire block set).

Took a swing at explaining some of the new validation options as well, LMK if something is factually incorrect!

Additionally, I got bitten by the fact that the `--ethereum_rpc_urls` flag/var seems to accept multiple endpoints, but in reality [ignores everything after the first valid one](https://github.com/ceramicnetwork/rust-ceramic/blob/main/event-svc/src/event/validator/time.rs#L46-L55). That's now made clear in the docs too :cupcake:

### DIFF
--- a/one/src/daemon.rs
+++ b/one/src/daemon.rs
@@ -240,6 +240,8 @@ pub struct DaemonOpts {
     anchor_poll_retry_count: u64,
 
     /// Ethereum RPC URLs used for time events validation. Required when connecting to mainnet and uses fallback URLs if not specified for other networks.
+    /// 
+    /// Note: only the first valid RPC URL will be used by the time event validator
     #[arg(
         long,
         use_value_delimiter = true,
--- a/one/src/migrations.rs
+++ b/one/src/migrations.rs
@@ -13,7 +13,7 @@ use clap::{Args, Subcommand};
 use futures::{stream::BoxStream, StreamExt};
 use multihash_codetable::{Code, Multihash, MultihashDigest};
 use tokio::io::AsyncBufReadExt;
-use tracing::{debug, info};
+use tracing::{debug, info, warn};
 
 use crate::{default_directory, DBOpts, Info, LogOpts};
 
@@ -66,13 +66,14 @@ pub struct FromIpfsOpts {
     #[command(flatten)]
     log_opts: LogOpts,
 
-    /// Path of file containing list of newline-delimited file paths to migrate.
+    /// Path of file containing list of newline-delimited paths to blocks to migrate. The paths must either be absolute,
+    /// or relative to the current directory where the migration command is run.
     ///
     /// See below for example usage when running a migration for a live IPFS node. Multiple migration runs using lists
     /// of files that have changed between runs is useful for incremental migrations. This method can also be used for
     /// a final migration after shutting down the IPFS node so that all inflight blocks are migrated to the new C1 node.
     ///
-    /// # Get list of files in sorted order
+    /// # Get list of absolute file paths to blocks, in sorted order
     ///
     /// find ~/.ipfs/blocks -type f | sort > first_run_files.txt
     ///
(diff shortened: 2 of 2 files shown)


one/src/migrations.rs
```diff
@@ -322,17 +331,26 @@ impl FSBlockStore {
...
     if !block_path.is_file() {
+        if block_path.is_absolute() {
+            warn!(
+                path = %block_path.display(),
+                "block file not found, skipping");
+        } else {
+            warn!(
+                path = %block_path.display(),
+                "block file not found at relative path, skipping");
+        }
```
smrz2001, line 342: I believe the check at line 33 is for skipping directories, so we shouldn't be getting here for any files.

one/src/daemon.rs
```diff
@@ -240,6 +240,8 @@ pub struct DaemonOpts {
     anchor_poll_retry_count: u64,
 
     /// Ethereum RPC URLs used for time events validation. Required when connecting to mainnet and uses fallback URLs if not specified for other networks.
+    /// 
+    /// Note: only the first valid RPC URL will be used by the time event validator
```
smrz2001, line 244: I believe the code uses the first RPC URL _to return a successful result_ but continues iterating through the rest until it finds one that succeeds, or all of them fail.

one/src/migrations.rs
m0ar: @smrz2001 it does trigger when I "accidentally" pass relative paths that can't be found, but it might _also_ trigger on empty dirs I guess :thinking: 

image

one/src/daemon.rs
m0ar: Yes, but I think this is only run when the `TimeEventValidator` is _constructed_, not every time it tries to validate an event, and it returns a hashmap which can only have one entry per key (which is the chain ID).

So what's happening here is that when starting the daemon, it checks if the first RPC URL can respond with it's own chain ID (see `HttpEthRpc` below). If so, it's set in a hashmap where the chain ID is the key, hence only one provider per chain ID is set in `chain_providers`.

image

image


Hence, when it's time to validate it just checks if there is a (as in one) provider for that chain ID, and uses it for the check:
image

m0ar: Oh also I did try this out when switching to the blastapi RPC Nathaniel recommended. If that RPC URL is first, the node can sync, but if I flip it it fails just like if I only had the Anrk one set.

image

one/src/migrations.rs
m0ar: Any ideas on how to reconcile those checks? The issue I want to cover is it silently skipping every single block if the paths aren't correct (`!block_path.is_file()`). 

It's used in 2 spots:
- `blocks_from_list`, this looks fine
- `blocks_from_dir`, used during sharded imports. Since we still return `Ok`, I think it'd just emit a warn for not being able to find a block at a directory path (which arguably is correct), if that's not avoided by the dir check:
    image


Anyway, happy for suggestions on how to improve the fix!

dav1do: nit: since the only difference between these branches is the message data, you could capture that and only have a single warn included.. e.g. one of these

```rust
warn!(path = %block_path.display(),
    relative_path = %!block_path.is_absolute(),
    "block file not found, skipping");

warn!(path = %block_path.display(),
    "block file not found at {} path, skipping", if block_path.is_absolute() {"absolute"} else {"relative"} );
 
let msg = if block_path.is_absolute() {} else {};
 warn!(path = %block_path.display(), "{}", msg);
```

smrz2001:
> @smrz2001 what's the status of this PR? Can this be closed?

Let me discuss this with @m0ar.

one/src/daemon.rs
smrz2001, line 244:
```suggestion
    /// Note: only the first valid RPC URL for a particular chain will be used by the time event validator
```

m0ar:
@stbrody @smrz2001 Hey guys, sorry for snoozing on this. I've applied the suggestions :)

smrz2001 (approved):
🚀

PR #713, opened 2025-04-11
fix: modify set account relation validation

dav1do:
We now check required fields are used to set `unique` header, and require `content` exists on data events. We don't handle set fields that are objects, arrays.. we could, the [js code](https://github.com/ceramicnetwork/js-ceramic/blob/7bd4288c42036aa0ed411d37fcac9ca077169735/packages/stream-model-instance-handler/src/model-instance-document-handler.ts#L431) appears to use `[object Object]` for objects so it'd be broken and we would need to support that.


pipeline/src/aggregator/mod.rs
```diff
@@ -3124,7 +3124,7 @@ mod tests {
...
+            | 2                 | baeabeihrpnqmppjqdohl2stvfevu7jn57zjqzqb4dflbl4ettogmt56xoy | 3           | did:key:alice | {controller: 6469643a6b65793a616c696365, model: ce01020100122031c0f042da99464be96f4b0d29afc5e9406219cb2885152105b476f1dd3ccc38, unique: 616c696365} | baeabeials2i6o2ppkj55kfbh7r2fzc73r2esohqfivekpag553lyc7f6bi | 0          | 0            | {"content":null,"metadata":{}}                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      | [Missing content for SET account relation]                |
```
dav1do: I changed this behavior.. this should be inline with js-ceramic behavior that rejects data events for SET account relation without content.

dav1do: also added a new test for a mismatch, so the diff looks wonky but it adds a new event 2 and moves everything down

smrz2001 (approved):
:rocket:

PR #718, opened 2025-05-14
fix: prevent panic on duplicate tracing initialization

smrz2001:


sdk/packages/flight-sql-client/src/lib.rs
```diff
@@ -29,14 +29,14 @@ use crate::flight_client::{execute_flight, setup_client, ClientOptions};
...
         .with_env_filter(
             EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("info")),
         )
-        .finish();
-
-    // Set the global default subscriber
-    tracing::subscriber::set_global_default(subscriber).expect("Failed to set tracing subscriber");
+        .try_init()
+    {
+        warn!("Tracing already initialized: {}", result);
```
stbrody, line 38: do we know that the only possible error is already initialized?  Is there a specific error code we can check to confirm and only swallow already initialized but Panic if there's a different type of failure?

sdk/packages/flight-sql-client/src/lib.rs
```diff
@@ -29,14 +29,14 @@ use crate::flight_client::{execute_flight, setup_client, ClientOptions};
...
     // Set up a subscriber that logs to stdout
-    let subscriber = tracing_subscriber::FmtSubscriber::builder()
+    if let Err(result) = tracing_subscriber::FmtSubscriber::builder()
         .with_env_filter(
             EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("info")),
         )
-        .finish();
-
-    // Set the global default subscriber
-    tracing::subscriber::set_global_default(subscriber).expect("Failed to set tracing subscriber");
```
stbrody, line 39: this line seems to go away entirely?  Is it just not needed?  What was it doing before?

smrz2001: Yeah, the [`try_init()`](https://docs.rs/tracing-subscriber/latest/src/tracing_subscriber/fmt/mod.rs.html#1156) does the same thing in a way that's safer to use in libraries.
```
/// Install a global tracing subscriber that listens for events and
/// filters based on the value of the [`RUST_LOG` environment variable],
/// if one is not already set.
```

smrz2001: `tracing::subscriber::set_global_default()` is appropriate to use in applications, not libraries (like our SDK).

smrz2001: https://docs.rs/tracing/latest/tracing/index.html#in-executables
```
Warning: In general, libraries should not call set_global_default()! Doing so will cause conflicts when executables that depend on the library try to set the default later.
```

sdk/packages/flight-sql-client/src/lib.rs
smrz2001: Good point. Updated. Though there are other ways, a string match appears to be the simplest way to check.

stbrody (approved):
lgtm

PR #724, opened 2025-06-10
fix(sdk): use correct controller during stream creation

### LINKED ISSUE #725: BUG: C1 SDK uses session DID Key as new stream controller instead of DID PKH

smrz2001:
cc @m0ar 

Closes #725

### DIFF
--- a/tests/suite/src/__tests__/correctness/fast/model-mid-did-session.test.ts
+++ b/tests/suite/src/__tests__/correctness/fast/model-mid-did-session.test.ts
@@ -71,14 +71,15 @@ export const authorizedSessionDid = async (
     return session.did
 }
 
-describe('create stream using session did', () => {
+describe('create/update stream using session did', () => {
     let flightClient: FlightSqlClient
     let client: CeramicClient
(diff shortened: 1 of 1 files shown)


sdk/packages/model-instance-client/src/client.ts
```diff
@@ -28,8 +28,8 @@ import type { DocumentState, UnknownContent } from './types.js'
 export type CreateSingletonParams = {
   /** The model's stream ID */
   model: StreamID
-  /** The controller of the stream (DID string or literal string) */
-  controller: DIDString | string
```
stbrody, line 32: this is losing the ability for the user to set the controller DID string explicitly, but maybe that's fine?  Almost everyone is going to be using did-session anyway so maybe that doesn't really matter as a feature

smrz2001: Yeah, I think that should be ok.

I _could_ have left the DID strings in but that would have meant leaving it to callers to check whether the DID had a parent PKH and then to pass that in, which felt risky.

PR #730, opened 2025-07-30
fix(sdk): set correct unique value for mids with set relation

smrz2001:
Please also review DX for improvements.


stephhuynh18 (approved):
From what i can see it looks fine but also i'm def not the best person

PR #732, opened 2025-08-14
daemon: allow configuring initial interests

### LINKED ISSUE #707: Add `--initial-interests` configuration variable
It would be great to be able to ship a ceramic-one configuration in a compose or env file which includes which models it should initially care about, and not have to manually set interests with the http-client or a request to `/interests`.

I imagine this functionality: when the daemon starts, pass any models in `--initial-interests` to whatever handles interests, and "upsert" those into potentially existing interests.

(Note: I do plan to PR this myself, but any pointers to where/how would be appreciated 👌 )

m0ar:
This PR implements a new CLI flag `--initial-interests` (and envvar `CERAMIC_ONE_INITIAL_INTERESTS`) to automatically register node interests when starting the daemon, eliminating the need for manual API calls after startup.

# Changes

- New CLI argument `--initial-interests`, accepting comma-separated stream/model IDs
- New startup module `one/src/startup.rs` to orchestrate daemon configuration tasks tangential to the startup process
- Integration:
  - Uses the existing `InterestService`
  - Uses the same pattern as the `/ceramic/interests` API handler
  - Processes interests early in daemon startup, after services init but before P2P setup

# Safety & backward compatibility

- Optional flag with empty default
- Stream ID format and multibase decoding validation
- Descriptive errors for invalid inputs + handling of failures
- Logging for registration status and debugging
- Idempotent: Handles already-registered interests without errors

# Usage

## Flag
```bash
ceramic-one daemon --initial-interests model1,model2
```

## Envvar
```bash
export CERAMIC_ONE_INITIAL_INTERESTS="model1,model2"
ceramic-one daemon
```

## Tests

Test suite added to `one/src/startup.rs`, can be run with the C1 crate:
```bash
> cargo test --locked --release -p ceramic-one
[ build logs ]

running 8 tests
test http::test::expected_prom_cardinality_replacement ... ok
test http::test::should_not_replace_extra_parts ... ok
test http::test::should_not_replace_wrong_method ... ok
test startup::tests::test_process_initial_interests_invalid_stream_id ... ok
test startup::tests::test_process_initial_interests_empty ... ok
test startup::tests::test_process_initial_interests_multibase_decoding ... ok
test startup::tests::test_process_initial_interests_mixed_valid_invalid ... ok
test startup::tests::test_process_initial_interests_whitespace_handling ... ok

test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
```

# Other notes
It seems to work, but I can't say I fully understand what's going on with the interest builder EventID fencepost ranges. Probably a good spot to focus a review :eyes: 

Closes #707

### DIFF
--- a/one/src/daemon.rs
+++ b/one/src/daemon.rs
@@ -1,7 +1,7 @@
 use std::{path::PathBuf, time::Duration};
 
 use crate::{
-    default_directory, handle_signals, http, http_metrics, metrics, network::Ipfs, DBOpts,
+    default_directory, handle_signals, http, http_metrics, metrics, network::Ipfs, startup, DBOpts,
     DBOptsExperimental, Info, LogOpts, Network,
 };
 use anyhow::{anyhow, bail, Result};
@@ -274,6 +274,19 @@ pub struct DaemonOpts {
         default_value = "file://./pipeline"
     )]
     object_store_url: url::Url,
+
+    /// Comma-separated list of stream IDs or model IDs to register as interests on startup.
+    /// The node will track streams implementing these models, and sync their data.
+    ///
+    /// For this to work with user-defined models, the daemon also needs to track the
+    /// metamodel: kh4q0ozorrgaq2mezktnrmdwleo1d
+    #[arg(
+        long,
+        use_value_delimiter = true,
+        value_delimiter = ',',
+        env = "CERAMIC_ONE_EXTRA_INTERESTS"
+    )]
+    extra_interests: Vec<String>,
 }
 
 fn spawn_database_optimizer(
@@ -446,6 +459,10 @@ pub async fn run(opts: DaemonOpts) -> Result<()> {
 
     let node_key = NodeKey::try_from_dir(opts.p2p_key_dir).await?;
     let node_id = node_key.id();
+
+    // Process initial interests if provided
+    startup::process_extra_interests(&opts.extra_interests, &interest_svc, &network, &node_id)
+        .await?;
 
     // Register metrics for all components
     let recon_metrics = MetricsHandle::register(recon::Metrics::register);
--- a/one/src/startup.rs
+++ b/one/src/startup.rs
@@ -1,0 +1,191 @@
+//! Startup utilities for the Ceramic One daemon.
+
+use anyhow::{anyhow, Result};
+use ceramic_api::InterestService as InterestServiceTrait;
+use ceramic_core::{EventId, Interest, NodeId, StreamId};
+use ceramic_interest_svc::InterestService;
+use std::str::FromStr;
+use std::sync::Arc;
+use tracing::{debug, info, warn};
+
+/// Process initial interests by registering them with the interest service
+pub async fn process_extra_interests(
+    extra_interests: &[String],
+    interest_svc: &Arc<InterestService>,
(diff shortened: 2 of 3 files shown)


one/src/daemon.rs
```diff
@@ -274,6 +274,16 @@ pub struct DaemonOpts {
...
+
+    /// Comma-separated list of stream IDs or model IDs to register as initial interests on startup.
+    /// The node will track these streams and sync their data.
+    #[arg(
+        long,
+        use_value_delimiter = true,
+        value_delimiter = ',',
+        env = "CERAMIC_ONE_INITIAL_INTERESTS"
+    )]
+    initial_interests: Vec<String>,
```
dav1do, line 286: nit: trying to decide what this name should be.. it's not exactly "initial" because they don't ever go away, but _NEW_ or _MODEL_ steams a bit silly

one/src/startup.rs
```diff
@@ -0,0 +1,186 @@
...
+
+        // Test daemon-specific multibase decoding logic
+        let stream_ids = vec![
+            "k2t6wz4ylx0qr6v7dvbczbxqy7pqjb0879qx930c1e27gacg3r8sllonqt4xx9".to_string(), // Valid multibase-encoded stream ID
+        ];
+
+        let result =
+            process_initial_interests(&stream_ids, &interest_svc, &network, &node_id).await;
+        // Should succeed with proper multibase decoding
+        assert!(result.is_ok());
```
dav1do, line 184: nit: maybe add an assertion this was written to the db?

```rust        
let vals = interest_svc.full_range().await.unwrap();
assert!(vals.count(), 1);
```

one/src/startup.rs
```diff
@@ -0,0 +1,186 @@
...
+        // Validate that the model stream ID is parseable
+        let _stream_id = StreamId::from_str(stream_id_str)
+            .map_err(|e| anyhow!("Invalid model ID '{}': {}", stream_id_str, e))?;
+
+        // Create an interest for the "model" separator key covering the full range for this stream
+        // This follows the same pattern as the API endpoint /ceramic/interests/model/{stream_id}
+        let stream_id_bytes = multibase::decode(stream_id_str)
+            .map_err(|e| anyhow!("Failed to decode stream ID '{}': {}", stream_id_str, e))?
+            .1;
+        let start = EventId::builder()
```
dav1do, line 39: I was considering if it'd be worth putting this in interest service or core so we could share most of this with the http server. But I think this API is fine to be exposed, I just couldn't remember how to build interest ranges and wanted a higher level API that's like `subscribe_to_model` 😅

one/src/daemon.rs
dav1do: also, I wonder if we want to subscribe to the `METAMODEL_STREAM_ID` automatically/easily? unless we do and I missed it, but that's the "I'd like to see all models on the network".

dav1do (commented):
Looks good. I have a few small things/questions but nothing blocking.

one/src/daemon.rs
m0ar: `EXTRA_INTERESTS` kinda matches how the daemon treats peers, so I'll go with that :ok_hand:

PR #733, opened 2025-08-15
bug: flush aggregator cache on shutdown

### LINKED ISSUE #728: BUG: sporadic pipeline data loss
**Describe the bug**
Fairly randomly, possibly in combination with node restarts, we're losing events from the `event_states` stage of the pipeline.

These are the 915 unique stream CID's from `event_states` after we migrate our kubo blockstore: [gist](https://gist.githubusercontent.com/m0ar/b923f7f3c4068f8d1be4f5d21cdda804/raw/374d21baafa04284c75a36d3a9d03c5114e1b3d1/distinct_stream_cid_initial.txt). After migration, the pipeline state is stable over multiple restarts.

Then, for unknown reasons, it sometimes unexpectedly drops to 596 known streams: [gist](https://gist.githubusercontent.com/m0ar/b923f7f3c4068f8d1be4f5d21cdda804/raw/374d21baafa04284c75a36d3a9d03c5114e1b3d1/distinct_stream_cid_after_disappeared.txt)

When this happens, the node never manages to reconcile the pipeline and is stuck with the aggregated state loss. Querying the `conclusion_events` table, I find 915 unique stream CID's which indicates the problem is localized to the aggregator pipeline. 

This has now happened 3 times, without us doing anything wonky that could potentially trigger it.

**To Reproduce**
@smrz2001 has a tarball of the sqlite DB and pipeline after the loss happened.

**Ceramic versions** 
0.55.1

m0ar:
# Note: WIP

Closes #728

### DIFF
--- a/actor/src/lib.rs
+++ b/actor/src/lib.rs
@@ -61,6 +61,16 @@ pub trait Actor {
     }
 }
 
+/// Optional trait for actors that need to perform cleanup operations during shutdown.
+/// Only actors that need cleanup should implement this trait.
+#[async_trait]
+pub trait Shutdown {
+    /// Called when the actor is shutting down to allow for cleanup operations.
+    /// This method is called after all pending messages have been processed
+    /// but before the actor loop exits.
+    async fn shutdown(&mut self);
+}
+
 /// Wrapper of any T with its tracing span context.
 #[derive(Debug)]
 pub struct Traced<T> {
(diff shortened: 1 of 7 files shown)


pipeline/src/aggregator/mod.rs
```diff
@@ -802,38 +802,47 @@ impl Aggregator {
...
+        self.ctx
+            .table(EVENT_STATES_MEM_TABLE)
+            .await?
+            .count()
+            .await
+            .context("count mem table")
+    }
+
+    async fn flush_cache(&self) -> Result<usize> {
+        let cnt = self.count_cache().await?;
```
dav1do, line 823: we double count in the base case (check count, if over our threshold, flush where we count again for logging purposes).. it's probably not a bit deal, but pointing it out

actor/src/lib.rs
```diff
@@ -61,6 +61,16 @@ pub trait Actor {
...
 }
 
+/// Optional trait for actors that need to perform cleanup operations during shutdown.
+/// Only actors that need cleanup should implement this trait.
+#[async_trait]
+pub trait Shutdown {
+    /// Called when the actor is shutting down to allow for cleanup operations.
+    /// This method is called after all pending messages have been processed
+    /// but before the actor loop exits.
+    async fn shutdown(&mut self);
```
dav1do, line 71: As I thought about why we used have a cache table at all (to make sure the parquet files aren't tiny), I realized this could lead to a similar result where we write smaller batches of events. For now, I think that's fine. Ideally, we're able to sort out making the aggregator idempotent on resume.

dav1do (commented):
Looks good. @smrz2001 said he can take a look at the actions.

actor/src/lib.rs
m0ar: The shutdown write will always be smaller than the 10k cache limit of course, but a full batch write yields parquet files only between 25-62 events long. I wouldn't think it makes much of a difference tbh if there are some that are even shorter if it means the aggregator being reliable on restart :grin: 

I plotted histograms of the distributions of an automatic 10k flush, and a shutdown flush of about 4k events:
image

m0ar:
Closing in favour of #735 (non-fork)

PR #737, opened 2025-08-30
fix: aggregator restart missing events

dav1do:
Fix issue where the aggregator could restart and miss events, causing validation errors on valid streams.

This includes two changes:
- First, we make the aggregator `process_conclusion_events_batch` and `join_event_states` idempotent and exclude duplicate incoming events by preferring the on disk events when they are duplicated in the batch. This means we _could_ go backward/see events again and not cause errors.
- Second, fix the issue where the memory cache seemed to be "going backwards", in that we'd flush events to disk and then we'd have events with a lower `conclusion_event_order` in memory, so on restart we'd skip them, as we started from our previous on disk max. As we'd have missing events (usually init), this would cause aggregator errors and streams would be full of validation errors. This was due to the fact that we write events from the batch to disk in phases (Models, MIDs). Although our batch was from [X, Y], the two internal batches could have interleaving order, so if the first batch caused a cache flush, we'd leave the remaining events in memory. Now we only flush the cache once per conclusion events batch, so they are all written to disk together. This means our cache table can grow even more over it's allotted size, but it's short lived and worth it.

With this, we shouldn't need the shutdown task to flush but I don't see any harm in keeping it so it's still there.

### DIFF
--- a/tests/hermetic/src/cli/tester.rs
+++ b/tests/hermetic/src/cli/tester.rs
@@ -791,7 +791,7 @@ fn job_init() -> Container {
         command: Some(vec![
             "/bin/sh".to_owned(),
             "-c".to_owned(),
-            "apk add --no-cache jq curl && /network/process-peers.sh".to_owned(),
+            "apk add --no-cache bash jq curl && /network/process-peers.sh".to_owned(),
         ]),
         env: Some(vec![
             EnvVar {


pipeline/src/aggregator/mod.rs
```diff
@@ -802,10 +833,14 @@ impl Aggregator {
...
-        // If we have enough data cached in memory write it out to persistent store
-        if count >= self.max_cached_rows {
+        if allow_cache_flush {
             self.flush_cache().await?;
+            let count = self.count_cache().await?;
+            tracing::debug!(%count, will_flush = %count >= self.max_cached_rows, "counts for mem table");
+            // If we have enough data cached in memory write it out to persistent store
+            if count >= self.max_cached_rows {
+                self.flush_cache().await?;
+            }
```
m0ar, line 843: Would be good to document why this check is important. I don't really understand it outside of testing use cases, in particular the block on l301 :thinking:

pipeline/src/aggregator/mod.rs
```diff
@@ -3503,4 +3563,303 @@ mod tests {
...
+        events
+            .iter_mut()
+            .enumerate()
+            .for_each(|(i, event)| match event {
+                ConclusionEvent::Data(data) => data.order = i as u64,
+                ConclusionEvent::Time(time) => time.order = i as u64,
+            });
+
+        // splice the 3 events together making sure each vec isn't reordered but not all in a row
+        let events = &events;
```
m0ar, line 3815: Not sure if this comment refers to the right thing

pipeline/src/aggregator/mod.rs
```diff
@@ -292,7 +298,7 @@ impl Aggregator {
         let models = self.patch_models(models).await.context("patching models")?;
         let models = self.validate_models(models).context("validating models")?;
         let mut models = self
-            .store_event_states(models)
+            .store_event_states(models, false)
```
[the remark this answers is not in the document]
dav1do: so this is the crux of the fix of what I tried to explain in the second bullet point (verbosely and imprecisely). basically, we put the entire batch of conclusion events into memory (models here and MIDs below) and only flush once so that we can't end up with the in memory order being behind the on disk order.

PR #756, opened 2026-01-08
fix: self-anchoring proof storage and validation fixes

m0ar:
# WIP

## Re-implement RPC proof discovery
#755 prematurely removed some fallback logic required for nodes to self-heal after failing to persist self-anchoring inclusion proofs. The relevant changes in `discover_chain_proof` are reverted here.

## Fallback validation to account for incorrect codec slice in anchor root
Single event anchor batch roots didn't initially get a `dag-pb` tree node wrapper, instead the `dag-jose` hash was written to the chain. However, it assumed it should slice off the first 4 bytes from the cid to strip it down to a hash, but `dag-jose` has a longer codec sequence. This resulted in the root hash keeping the `0x20` (digest length 32) in the front, causing the last digest byte to be dropped.

The fix is falling back to a partial 31-byte comparison for gnosis (clay) anchors, for anchors in a specific time window, and only if the prefix is `0x20`.

## Self-anchoring proof persistence
The self-anchoring process didn't store the inclusion proofs in the DB, causing pipeline errors for self-anchored events. This is now implemented so the node immediately can validate the events that it created and anchored.

Approach:

1. anchor-service/src/transaction_manager.rs: Added ChainInclusionData struct and chain_inclusion field to RootTimeEvent
2. anchor-service/src/lib.rs: Exported ChainInclusionData
3. anchor-service/src/anchor.rs: Added chain_inclusion field to TimeEventBatch and TimeEventInsertable, updated conversion methods
4. anchor-service/src/anchor_batch.rs: Flows chain_inclusion from RootTimeEvent to TimeEventBatch
5. anchor-evm/src/evm_transaction_manager.rs:
6. Added AnchorResult struct (moved outside impl block)
7. Modified submit_and_wait to return block hash, timestamp, and tx_input
8. Creates ChainInclusionData in anchor_root with all needed fields
9. Fixed get_block_by_number API call (changed BlockTransactionsKind::Hashes to false)
10. anchor-service/src/cas_mock.rs and anchor-remote/src/cas_remote.rs: Added chain_inclusion: None to RootTimeEvent
11. event-svc/src/store/sql/entities/chain_proof.rs: Added From<ChainInclusionData> impl for ChainProof
12. event-svc/src/event/store.rs: Updated ceramic_anchor_service::Store::insert_many to extract and persist chain inclusion proofs
13. Test files: Updated all expectation files to include chain_inclusion: None


stbrody:
@m0ar is this PR still relevant or is it superseded by the other ones you've opened?

m0ar:
> @m0ar is this PR still relevant or is it superseded by the other ones you've opened?

This has been tidied up and split into 3 others PRs 👌 Can close.

PR #757, opened 2026-01-19
fix: self-anchoring never stored inclusion proofs, causing concluder failures

m0ar:
The self-anchoring process didn't store the inclusion proofs in the DB, causing pipeline errors for self-anchored events. This is now implemented so the node immediately can validate the events that it created and anchored.

Approach:

1. `anchor-service/src/transaction_manager.rs`: Added `ChainInclusionData` struct and `chain_inclusion` field to `RootTimeEvent`
2. `anchor-service/src/lib.rs`: Exported `ChainInclusionData`
3. `anchor-service/src/anchor.rs`: Added `chain_inclusion` field to `TimeEventBatch` and `TimeEventInsertable`, updated conversion methods
4. `anchor-service/src/anchor_batch.rs`: Flows `chain_inclusion` from `RootTimeEvent` to `TimeEventBatch`
5. `anchor-evm/src/evm_transaction_manager.rs`:
    - Added `AnchorResult` struct (moved outside impl block)
    - Modified `submit_and_wait` to return block `hash`, `timestamp`, and `tx_input`
    - Creates `ChainInclusionData` in `anchor_root` with all needed fields
10. `anchor-service/src/cas_mock.rs` and `anchor-remote/src/cas_remote.rs`: Added `chain_inclusion: None` to `RootTimeEvent`
11. `event-svc/src/store/sql/entities/chain_proof.rs`: Added `From<ChainInclusionData>` impl for `ChainProof`
12. `event-svc/src/event/store.rs`: Updated `ceramic_anchor_service::Store::insert_many` to extract and persist chain inclusion proofs
13. Test files: Updated all expectation files to include `chain_inclusion: None`


event-svc/src/event/store.rs
```diff
@@ -260,6 +260,28 @@ impl ceramic_anchor_service::Store for EventService {
         items: Vec<ceramic_anchor_service::TimeEventInsertable>,
         informant: NodeId,
     ) -> Result<()> {
+        // Extract chain inclusion data from the first item that has it.
+        // All items in a batch share the same chain inclusion data since they're from the same anchor tx.
```
stbrody, line 264: hmm, that's potentially a lot of duplicate data in memory.  Maybe fine for now but probably something to keep an eye on if memory usage becomes an issue down the line

stbrody (approved):

event-svc/src/event/store.rs
m0ar: Yeah, from the CAS perspective which acts like a bottleneck it'd be pretty bad, but with recurring self-anchoring it'd be unexpected with batches that large. With that in mind, we can also tune it with the anchor batch size and interval.

m0ar:
Closing in favor of continued work in our fork @ https://github.com/desci-labs/rust-ceramic

PR #759, opened 2026-01-19
Fix: fallback validation of shifted proof hashes on clay

m0ar:
Single event anchor batch roots didn't initially get a `dag-pb` tree node wrapper, instead the `dag-jose` hash was written to the chain. However, it assumed it should slice off the first 4 bytes from the cid to strip it down to a hash, but `dag-jose` has a longer codec sequence. This resulted in the root hash keeping the `0x20` (digest length 32) in the front, causing the last digest byte to be dropped.

The fix is falling back to a partial 31-byte comparison for gnosis (clay) anchors, for anchors in a specific time window, and only if the prefix is `0x20`.

Note: based on `fix/revert-proof-autodiscovery-removal` / 54a2122d505a4860f4699870d0ab4d424ca2007f

### DIFF
--- a/event-svc/src/event/validator/time.rs
+++ b/event-svc/src/event/validator/time.rs
@@ -90,7 +90,35 @@ impl TimeEventValidator {
         // Compare the root hash in the TimeEvent's AnchorProof to the root hash that was actually
         // included in the transaction onchain. We compare hashes (not full CIDs) because the
         // blockchain only stores the hash - the codec is not preserved on-chain.
-        if chain_proof.root_cid.hash() != event.proof().root().hash() {
+        let chain_digest = chain_proof.root_cid.hash().digest();
+        let event_proof_root = event.proof().root();
+        let proof_digest = event_proof_root.hash().digest();
+
+        if chain_digest != proof_digest {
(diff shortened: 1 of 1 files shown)


event-svc/src/event/validator/time.rs
```diff
@@ -90,7 +90,35 @@ impl TimeEventValidator {
         // Compare the root hash in the TimeEvent's AnchorProof to the root hash that was actually
         // included in the transaction onchain. We compare hashes (not full CIDs) because the
         // blockchain only stores the hash - the codec is not preserved on-chain.
-        if chain_proof.root_cid.hash() != event.proof().root().hash() {
+        let chain_digest = chain_proof.root_cid.hash().digest();
+        let event_proof_root = event.proof().root();
+        let proof_digest = event_proof_root.hash().digest();
+
+        if chain_digest != proof_digest {
```
stbrody, line 97: nitpick: would be a bit cleaner if this logic was encapsulated in a helper function

stbrody (approved):

m0ar:
Closing in favor of continued work in our fork @ https://github.com/desci-labs/rust-ceramic

